PR #1, opened 2019-05-20
Add macroexpansion style expander

nilern:
Applies
```
:foo ;=> (get registry :foo)
[:foo :bar] ;=> (apply (get registry :foo) [:bar])
```
exhaustively at each node in a prewalk of the schema.

If things are not found in registry, leaves them unexpanded. Maybe that should be an error instead but it depends on the semantics of the expanded schemas which is left open here.


mangolas:
This looks fine for me, so if you fix conflicts let's merge and carry on forward.

PR #5, opened 2019-05-29
Design draft for having discussion in PR format

mangolas:
Let's try get common ideas and opinions further with this PR. 

So no hurry to merge this ... Tommi :)

### DIFF
--- a/src/malli/design_draft.clj
+++ b/src/malli/design_draft.clj
@@ -1,0 +1,96 @@
+(ns malli.design-draft
+  (:import (clojure.lang IPersistentMap)))
+
+
+;;; malli schemas
+
+;; Core schemas are not namespace qualified keywords to make using them user
+;; friendly.
+
+;; Schemas are written using Hiccup style vectors, where order is significant.
+
+;; Every schema definition can have additional options-map after schema name.
+;; Options-map can have arbitrary amount of data, with some keys having special
+;; meaning, like below, which describes a closed map.
+
+[:map {:closed true, :description "My closed map"}]
+
+;; Schema definitions live in registry. Malli provides a core registry for
+;; all the essential schemas. Registry can be extended on the runtime and it's
+;; possible to have as many registries one needs.
+
+;; The actual registry definitions and format is wip, but could be something
+;; like below.
+
+;; Compiles given definition to Schema
+(declare compile)
+(declare Schema)
+
+(def -map (compile {:id :map
+                    :type IPersistentMap
+                    :title "Map"
+                    :json-schema/type "object"
+                    :valid-fn (fn [data])}))
+
+(def -and (compile {:id :and
+                    ;; Does this make any sense?
+                    :json-schema/op "allOf"}))
+
+;; or create Schema directly
+(def -map (reify Schema '...))
+
+(def core-registry
+  {:map -map
+   :and -and})
+
+;; We could expose core schemas as vars as well for having development time
+;; documentation etc.
+(def ^{:doc "A map"} map :map)
+
+
+;;; TBD List core schemas
+;;; TBD List main functions operating on schemas
+;;; TBD Examples of validation and validation error messages
+;;; TBD Examples of coercion
+;;; TBD Examples of schema format transformation eg. to Json schema
+
+;;; Larger schema example
+
+;; Creating an inlined model inspired from schema.org Person
+(def example1
+  [:schema
+   [:map {:id :my/Thing
+          :title "A Thing"
+          :description
+          :schema.org/type "http://schema.org/ImageObject"}
+    :identifier [:or :uuid :url [:and :string :not-empty]]]
+   [:and {:id :my/ImageObject
+          :schema.org/type "http://schema.org/ImageObject"
+          :json-schema/id "http://my.org/ImageObject"}
+    [:map {:closed true
+           ;; Use keys from another map schemas?
+           :merge-keys [:my/Thing]
+           ;; Optional keys as map option?
+           :opt-keys [:caption :thumbnail]}
+           ;; All keys optional shortcut?
+           ;; `:opt-keys :all`}
+     :caption :string
+     :thumbnail [:coll-of :my/ImageObject]
+     :contentUrl :url
+     :uploadDate :date]
+    [:map-of
+     [[:and :keyword :camel-case] :any]]]
+   [:enum {:id :my/gender
+           :schema.org/type "http://schema.org/GenderType"}
+    [:male :female]]
+   [:map {:id :my/Person
+          :merge-keys [:my/Thing]
+          :json-schema/id "http://my.org/Person"
+          :description "A schema.org Person"
+          :schema.org/type "Person"}
+    ;; Optional key as key operator
+    [:opt :email] :string
+    ;; Optional key as key opt-map
+    :gender {:opt true} [:or :my/gender [:and :string :not-empty]]
+    :parent [:coll-of :my/Person]
+    :height {:opt true} :number]])
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 ## Rationale
 - Schemas as immutable data
-  - can be transported over the wire
-  - saved to the database
-  - combined freely
+  - Can be transported over the wire
+  - Saved to the database
+  - Combined freely
 - Does not depend on global registry
 - Programmable, on the runtime as well
 
@@ -28,3 +28,7 @@ Could have:
 - JOI https://github.com/hapijs/joi
 - JSON Schema https://json-schema.org/understanding-json-schema/index.html
 
+## WIP Design and examples
+
+Discussion about format and all open topics at 
+[src/malli/design_draft.clj](src/malli/design_draft.clj).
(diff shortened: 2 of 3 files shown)


src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+
+;; Schema definitions live in registry. Malli provides a core registry for
+;; all the essential schemas. Registry can be extended on the runtime and it's
+;; possible to have as many registries one needs.
+
+;; The actual registry definitions and format is wip, but could be something
+;; like below.
+
+;; Compiles given definition to Schema
+(declare compile)
```
nilern: I don't see the need for something like this. Instead I would just have `expand` (or `hydrate`?) to turn the hiccup-like data schema into a tree of `Schema` protocol implementors and then those can be compiled to checkers, coercions or generators. I think it is clearer to just create the schema implementations in registry with vanilla `deftype` or `reify`.

 I guess it would be possible to have `Schema` implement checking, coercion and generation directly but any other operations on schemas (e.g. emit SQL table schemas, produce error messages for end users...) would have to be added outside `Schema` anyway. And supporting several operations directly would probably inhibit some optimizations that a separate coercion compiler could do.

ikitommi: :+1:

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+(declare Schema)
+
+(def -map (compile {:id :map
```
ikitommi: agree with @nilern that direct impls for core things (like there is in the core now) is clearer. The core should be small and separate concerns, like emitting json-schemas, should be in separate namespaces. At bare minimum, the core only has models and validation. Coercion and generation and definitely model transformations, should be optional.

for end-uses, support of extension via map data is a good idea. e.g. define a single `my-uuid` and attach all the possible things to it via namespaced keys in the schema data.

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+                    :type IPersistentMap
```
ikitommi: we should target clojurescript, this is a Clojure type.

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+          :title "A Thing"
+          :description
+          :schema.org/type "http://schema.org/ImageObject"}
+    :identifier [:or :uuid :url [:and :string :not-empty]]]
+   [:and {:id :my/ImageObject
+          :schema.org/type "http://schema.org/ImageObject"
+          :json-schema/id "http://my.org/ImageObject"}
+    [:map {:closed true
+           ;; Use keys from another map schemas?
+           :merge-keys [:my/Thing]
```
ikitommi: I'm not fond of the flat sequence of keys and values, harder to program with those, kind of like the `& rest` params which is not used anymore with clojure function args to represent options. Instead, I would use a vector of key+value tuples (vectors) for maps.

so, instead:
```clj
[:map :x :int :y :string]
```
this:
```clj
[:map [:x :int] [:y :string]]
```
In clojure, the map is a sequence of entries, which are tuples. I think we should use that too: you can count the keys with `(count childs)`, not `(/ (count childs) 2)` etc.

if the `expand` starts from the top, it can delegate the `expand` to the child schemas, and `:map` knows how to expand it's children.

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+           ;; Optional keys as map option?
+           :opt-keys [:caption :thumbnail]}
```
ikitommi: are there too many ways to say the same?

I would use the opt/req model from Schema/data-specs:

```clj
[:map
  [[:opt :x] :int]
  [[:req :y] :string]
  [:z [:maybe :int]]]
```
if there is also extra info about these in the map meta-data, which one wins?

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+     :thumbnail [:coll-of :my/ImageObject]
+     :contentUrl :url
+     :uploadDate :date]
+    [:map-of
+     [[:and :keyword :camel-case] :any]]]
+   [:enum {:id :my/gender
+           :schema.org/type "http://schema.org/GenderType"}
+    [:male :female]]
+   [:map {:id :my/Person
+          :merge-keys [:my/Thing]
```
ikitommi: same, here, the merge could be both a function and a schema:

a function:

```clj
(m/merge
  [:map [:x :int]]
  [:map [[:opt :x] :string]])
```

a schema:

```clj
[:merge
  [:map [:x :int]]
  [:map [[:opt :x] :string]]]
```

is there merge in json-schema?

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+          :json-schema/id "http://my.org/Person"
+          :description "A schema.org Person"
+          :schema.org/type "Person"}
+    ;; Optional key as key operator
+    [:opt :email] :string
```
ikitommi: :+1: Ambiguity with the map option...

src/malli/design_draft.clj
```diff
@@ -0,0 +1,96 @@
...
+    ;; Optional key as key opt-map
+    :gender {:opt true} [:or :my/gender [:and :string :not-empty]]
```
ikitommi: here, there are 3 things related to one entry. hard to read it is.

ikitommi:
commented. in brief:

* I would suggest that we don't add any logic to the schema meta-data now, I think we can build most of the things without those, and there is ambiguity problem in the examples
* I would like to see the map entries as tuples, either direct childs or one sequence of childs
* One `m/expand` to expand the models to types/reified Schemas

... this could be the initial version to discuss more.

cheers.

src/malli/design_draft.clj
mangolas: Ok, sounds good, `expand` turns data-only schema to tree of `Schema` protocol instances. 

I don't have an opinion yet of should the checking aka validation, coercion and generation be part of the Schema implementation or not. But I think it definitely worth trying out to keep them separate from the Schemas to make multiple usages of Schemas as broad as possible. Lets see how it unfolds.

src/malli/design_draft.clj
mangolas: Ok, sounds good, I'll try to reflect that on the update to draft.

src/malli/design_draft.clj
mangolas: Good point

ikitommi: here are spec-tools types: https://github.com/metosin/spec-tools/blob/master/src/spec_tools/parse.cljc#L66-L86. Should we just use the schema name as a type if it is not defined? e.f. `:map` is a `:map` unless it has a spesidic `:type` overriding it? e.g. having a custom (non-serializable) function/reified schema, one could add a type hint for it. Spec-tools has this, I think it's a one of the good parts.

Also, I guess we want to support all platform-types, so one can say that something is an `Interger` and something is a `Long`. Needed for example in java-interop (schema for a function that calls a `MyJavaLogic(Short s)`.

ikitommi: the function Schemas... if we are going to support reified Schemas in schema values, is there a reason not to support functions and coerce them in to Schemas in the `expand` (or whatever) phase? a optional registry from function -> schema would be great?

```clj
;; all data, ok
[:and :int [:> 18]]

:: reified, ok, doesn't de/serialize unless `read` and `write` functions of Schema protocol are implemented, like EDN readers.
[:and (reify Schema (validator [_ value] (number? value))) [:> 18]]

;; explicit coercion of int? -> :int 
[:and int? [:> 18]]

;; fails, no conversion from int? => Schema
(p/schema
  [:and int? [:> 18]])

;; mega-success!
(p/schema
  [:and int? [:> 18]]
  {:registry (merge p/default-registry {int? :int})})
```

src/malli/design_draft.clj
mangolas: Ok, let's convert to stricter version, even if the we could allow sugar of kv-tuples without wrapping, that would complicate matters.

src/malli/design_draft.clj
mangolas: I would say key level opt/req is sugar for map level information. So map wins.

src/malli/design_draft.clj
mangolas: No, changing it to ref.

src/malli/design_draft.clj
mangolas: Sugar, map-option

src/malli/design_draft.clj
mangolas: Yes, let's take that opt away, no need for it

src/malli/design_draft.clj
mangolas: Surely we can should have functions, but as json-schema has `ref` I think for the declarative schema usage it makes more sense to try to approach merge cases with `ref`.

Thus
```
[:map {:id :A}]
[:map {:id :B}]
[:map {:id Merged :ref [:A :B]}
```

ikitommi:
Discussion was great and most of the things are cherry picked into the master. Let's do a final review before releasing initial version.

PR #6, opened 2019-06-26
Tommi's spike branch (WIP)

ikitommi:
Initial implementation of some schemas for learning purposes. Open for comments. Will add tests & rename the ns if this would be a good way to proceed.


src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
+(ns malli.malli)
```
mangolas: This looks really good! I think it just needs some more spikes in regard to producing (or even round tripping, which would be nice) json-schema.

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+
+(set! *warn-on-reflection* true)
+
+;;
+;; protocols
+;;
+
+(defprotocol IntoSchema
+  (-id [this])
```
mangolas: Is `id` good name? Could it be `name` or even `keyword` (which is used in the json-schema)/

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+  (-aliases [this])
```
mangolas: When is this needed on protocol level? I'm not sure I follow.

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+  (-into-schema [this props childs opts]))
+
+(defprotocol Schema
+  (-validator [this opts])
+  (-props [this])
```
mangolas: [nit] Perhaps using full word `options` or `properties`?

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+   (throw (ex-info (str type) {:type type, :data data}))))
+
+(defn create-form [id props childs]
+  (cond
+    (and (seq props) (seq childs)) (into [id props] childs)
+    (seq props) [id props]
+    (seq childs) (into [id] childs)
+    :else id))
+
+(defn- -var-schema [v]
```
mangolas: Is this expecting that `v` is predicate function?

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+     :keys (->> entries (mapv first))
+     :entries entries}))
+
+(defn- -map-schema []
+  ^{:type ::into-schema}
+  (reify IntoSchema
+    (-id [_] :map)
+    (-aliases [_])
+    (-into-schema [_ props childs opts]
+      (when-not (seq childs)
```
mangolas: I think empty `:map` should be allowed.

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+  ([?schema value]
+   (validate ?schema value nil))
+  ([?schema value opts]
+   ((validator ?schema opts) value)))
+
+;;
+;; registry
+;;
+
+(def default-registry
```
mangolas: This looks really nice!

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+   [[:req :z] string?]]
+
+  ;; tuples
+  [:map
+   [:x boolean?]
+   [:opt :y int?]
+   [:req :z string?]]
+
+  ;; attrs
+  [:map
```
mangolas: I would vote for this form

src/malli/malli.cljc
```diff
@@ -0,0 +1,302 @@
...
+
+(m/schema [:map
+           [:x boolean?]
+           [[:opt :y] int?]
+           [:z string?]])
+
+(comment
+  ;; schema-style
+  [:map
+   {:closed true}
```
mangolas: Is `:map` open or closed by default?

mangolas: I guess open, so non-default could be expressed as `{:open false}` as well.

src/malli/malli.cljc
mangolas: If key is by default required, is `{:opt true}' only meaningful setting?

src/malli/malli.cljc
ikitommi: it's `name` now.

src/malli/malli.cljc
ikitommi: `properties`.

src/malli/malli.cljc
ikitommi: removed.

src/malli/malli.cljc
ikitommi: :+1:

src/malli/malli.cljc
ikitommi: If `props` => `properties`, I don't the property should be `optional` Instead of `opt`. Let's revisit the closed/open & opt/req later when things look more solid.

src/malli/malli.cljc
ikitommi: not sure, but currently is.

PR #8, opened 2019-06-30
Set up Circle integration and ClojureScript tests

miikka:


ikitommi:
oh, so good

miikka:
Was that a review?

PR #9, opened 2019-07-03
Added :schema Schema and properties helper

ikitommi:


nilern (changes requested):
I find `:schema` to be a confusing name. I would expect it to be a schema that matches any schema definition (circular, but useful). Here it seems to attach props to any schema (like `with-meta`). Maybe something like `:with-props` would be more descriptive?

ikitommi:
removed `:schema`, not needed.

PR #28, opened 2019-08-06
Motivation

ikitommi:


README.md
```diff
@@ -224,7 +224,23 @@ Wrapping schemas into `m/schema` makes them first class entities. Here `User` is
 ; => true
 ```
 
-## Links
+## Motivation
+
+We are building dynamic multi-tenant systems, where data-models should be first-class: they should drive the runtime value transformations, forms and processes. We should be able to edit the models at runtime, persistent and load back them from database and over the wire, for both Clojure and ClojureScript. Think of [JSON Schema](https://json-schema.org/), but for Clojure/Script.
```
miikka:
```suggestion
We are building dynamic multi-tenant systems where data-models should be first-class: they should drive the runtime value transformations, forms and processes. We should be able to edit the models at runtime, persist them and load them back from database and over the wire, for both Clojure and ClojureScript. Think of [JSON Schema](https://json-schema.org/), but for Clojure/Script.
```

Did not add it to the suggestion, but maybe it's more like *JSON Schema for EDN* than for Clojure/Script.

README.md
```diff
@@ -224,7 +224,23 @@ Wrapping schemas into `m/schema` makes them first class entities. Here `User` is
...
+
+Isn't the problem solved (many times) already?
```
miikka:
```suggestion
Hasn't the problem been solved (many times) already?
```

README.md
```diff
@@ -224,7 +224,23 @@ Wrapping schemas into `m/schema` makes them first class entities. Here `User` is
...
+
+Well, there is [Schema](https://github.com/plumatic/schema), which is awesome, proven and collaborative open source project and we absolutely love it. We still used in most of our projects. Sad part: serializing & de-serializing schemas is non-trivial and there is no back-tracking on branching.
```
miikka:
```suggestion
There is [Schema](https://github.com/plumatic/schema), which is awesome, proven and collaborative open source project, and we absolutely love it. We still use it in most of our projects. Sad part: serializing & de-serializing schemas is non-trivial and there is no back-tracking on branching.
```

README.md
```diff
@@ -224,7 +224,23 @@ Wrapping schemas into `m/schema` makes them first class entities. Here `User` is
...
+
+[Spec](https://clojure.org/guides/spec) is de facto data spesification library for Clojure. It has many great ideas, but it is based on macros, has a global registry and doesn't support runtime transformations. [Spec-tools](https://github.com/metosin/spec-tools) was created to "fix" some of the things, but after [three years](https://github.com/metosin/spec-tools/commit/18aeb78db7886c985b2881fd87fde6039128b3fb) of developing it, it's still kinda hack and not fun to maintain.
```
miikka:
```suggestion
[Spec](https://clojure.org/guides/spec) is the de facto data specification library for Clojure. It has many great ideas, but it is based on macros, it has a global registry and it doesn't support runtime transformations. [Spec-tools](https://github.com/metosin/spec-tools) was created to "fix" some of the things, but after [three years](https://github.com/metosin/spec-tools/commit/18aeb78db7886c985b2881fd87fde6039128b3fb) of developing it, it's still kinda hack and not fun to maintain.
```

README.md
```diff
@@ -224,7 +224,23 @@ Wrapping schemas into `m/schema` makes them first class entities. Here `User` is
...
+
+So, we decided to try to spin out our own library, which would do all the things we feel is important for dynamic system development. It's based on the best parts of existing libraries and several project-spesific tools we have done over the years.
```
miikka:
```suggestion
So, we decided to try to spin out our own library, which would do all the things we feel is important for dynamic system development. It's based on the best parts of the existing libraries and several project-specific tools we have done over the years.
```

miikka (commented):
Typo fixes

Issue #35, opened 2019-08-13
Question: Way to add custom predicates


jeans11:
Hi!
I'm trying your lib and I'm wondering how add custom predicates.

Is this the way to add custom predicates?

```clojure
(defn size-between
  [s a b]
  (< a (count s) b))

(defn small-string?
  [s]
  (size-between s 1 10))

(deftest lib-test
  (let [registry
        (merge
          m/comparator-registry
          m/base-registry
          m/predicate-registry
          {small-string? (m/fn-schema :small-string? small-string?)})
        schema [:and string? small-string?]]
    (is (false? (m/validate schema "aaaaaaaaaaaaaaaaaaaaaaaaaa" {:registry registry})))))
```

ikitommi:
That is one way to do it, having all the schemas in a registry allows one to serialize & deserialize the schemas using it. If you don't need that, you can also just create and inline `Schema` instances. Currently, there is no helper for that (`fn-schema` returns a `IntoSchema` instance), but there could be a `:fn` wrapper for that:

```clj
(m/validate [:fn small-string?] "aaa")
; => true
```

for one-way serialization, there could be a property `:form` to tell how it's written:

```clj
(m/form [:and string? [:fn {:form 'small-string?} small-string?]])
; => [:and string? small-string?]
```

Ideas and comments most welcome.

ikitommi:
Also, as the `:registry` option is a function of `name->IntoSchema`, one could provide a custom implementation which in case of miss, checks if the input was a function and creates a anonymous function schemas out of it - all predicates would just work:

```clj
(def my-registry-auto-wrapping-functions-as-schemas ...)

(m/validate small-string? "aaa" {:registry my-registry-auto-wrapping-functions-as-schemas)
; => true
```
could add an example out of this into the README. Not likely to be a built-in feature.

ikitommi:
One more thing... the `validate`, `explain`, `transform` etc. are for convenience: they first compile the schemas and functions and then invoke them once (and lose the compilation result). To get also stellar perf, you should use `validator`, `explainer`,  `transformer` functions that just do the compilation step and return an optimized function for the actual job. All compiled functions are all pure.

Cheers.

jeans11:
Thanks @ikitommi

ikitommi:
Will keep this open until a helper is done.

ikitommi:
There is also a https://github.com/borkdude/sci, which could be used, if you know what you are doing:

```clj
[...]

PR #40, opened 2019-08-16
Add :fn helper

jeans11:
#35 

```clojure
(m/validate [:fn #(< 1 (count %) 10)] "aaa")
;; true

(m/validate [:and string? [:fn #(< 1 (count %) 10)]] "aaa")
;; true
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -275,6 +275,29 @@
           (-properties [_] properties)
           (-form [_] form))))))
 
+(defn- -function-schema []
+  ^{:type ::into-schema}
+  (reify IntoSchema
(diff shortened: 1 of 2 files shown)


ikitommi:
Looks just what the issue though of, thanks! Still, this would break the contract of having all the defaults specs always de+serializable. Need to think about how to handle this.

Added a guide into README how to do EVIL MUTATION on the `malli.core/default-registry` if you want to use this with minimum effort meanwhile.

jeans11:
I can close this PR in favor of https://github.com/metosin/malli/pull/44?

ikitommi:
Ok, let's close this. Thanks anyway!

PR #41, opened 2019-08-17
Add :map-of schema

jeans11:
#38 

```clojure
(m/validate [:map-of string? int?] {"age" 18})
;; true

(m/validate [:map-of keyword? int?] {"age" 18})
;; false

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -286,8 +286,9 @@
             key-valid? (-validator key-schema)
             value-valid? (-validator value-schema)
             validate (fn [m]
-                       (reduce
-                         #(or (and (key-valid? (first %2)) (value-valid? (second %2))) (reduced false))
+                       (reduce-kv
+                         (fn [_ key value]
+                           (or (and (key-valid? key) (value-valid? value)) (reduced false)))
                          true m))]
         ^{:type ::schema}
         (reify Schema
@@ -303,16 +304,15 @@
                              :schema this
                              :value  m
                              :type   ::invalid-type})
-                  (reduce
-                    (fn [acc [i [key value]]]
+                  (reduce-kv
+                    (fn [acc key value]
                       (let [in (conj in key)
-                            distance-path (+ i distance)
-                            key-explainer' (key-explainer (into path [distance-path 0]))
-                            value-explainer' (value-explainer (into path [distance-path 1]))]
+                            key-explainer' (key-explainer (conj path distance))
+                            value-explainer' (value-explainer (conj path (inc distance)))]
                         (->> acc
                              (key-explainer' key in)
                              (value-explainer' value in))))
-                    acc (map-indexed vector m))))))
+                    acc m)))))
           (-properties [_] properties)
           (-form [_] (create-form :map-of properties (mapv -form schemas))))))))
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/malli/core.cljc [hunk 1]
```diff
@@ -275,6 +275,47 @@
...
+  (reify IntoSchema
+    (-name [_] :map-of)
+    (-into-schema [_ properties childs opts]
+      (when-not (and (seq childs) (= 2 (count childs)))
+        (fail! ::invalid-map-of))
+      (let [[key-schema value-schema :as schemas] (mapv #(schema % opts) childs)
+            key-valid? (-validator key-schema)
+            value-valid? (-validator value-schema)
+            validate (fn [m]
+                       (reduce
```
ikitommi: `reduce-kv` should be used

ikitommi: also guard that the value is a `map?`

src/malli/core.cljc
```diff
@@ -275,6 +275,47 @@
...
+                  (conj acc {:path   path
+                             :in     in
+                             :schema this
+                             :value  m
+                             :type   ::invalid-type})
+                  (reduce
+                    (fn [acc [i [key value]]]
+                      (let [in (conj in key)
+                            distance-path (+ i distance)
+                            key-explainer' (key-explainer (into path [distance-path 0]))
```
ikitommi: I don't think this is right, the `:path` should point to the schema in a form, so `(conj [path distance])` and `(conj [path (inc distance)])`?

src/malli/core.cljc
```diff
@@ -275,6 +275,47 @@
...
+                  key-explainer (partial -explainer key-schema)
+                  value-explainer (partial -explainer value-schema)]
+              (fn explain [m in acc]
+                (if-not (map? m)
+                  (conj acc {:path   path
+                             :in     in
+                             :schema this
+                             :value  m
+                             :type   ::invalid-type})
+                  (reduce
```
ikitommi: `reduce-kv` here too

ikitommi:
Thanks for the PR! Was doing the same but will revert and use this instead. Few comments attached. I'll add a transformer after it's merged.

Also, both Prismatic & JSON Schema support maps with both names keys and predicate keys. This might be merged later with `:map`.

src/malli/core.cljc
jeans11: Are you talking about this? https://github.com/metosin/malli/pull/41/files#diff-80b352d39081ba397dc83fc6fafe7a47R294

ikitommi: Oh, the `map?` was there already, my bad

src/malli/core.cljc
```diff
@@ -275,6 +275,47 @@
...
+                (if-not (map? m)
+                  (conj acc {:path   path
+                             :in     in
+                             :schema this
+                             :value  m
+                             :type   ::invalid-type})
+                  (reduce-kv
+                    (fn [acc key value]
+                      (let [in (conj in key)
+                            key-explainer' (key-explainer (conj path distance))
```
ikitommi: These actually only use data which is available already at line 296->, and outside of the returned function. Lines 298 & 299 can create the concrete explainers already.

ikitommi:
Looking good, one small change request

ikitommi:
Thanks!!

ikitommi:
Merged the Schema transformations into master and added support for `-transformer` to `:map-of` too.

ikitommi:
related #43

PR #44, opened 2019-08-19
Implement :fn schema (WIP)

ikitommi:
Current status:

```clj
(m/validate [:fn '#(and (int? %) (< 10 % 18))] 12)
; => true
```

* also renamed `:problems` into `:errors` 

TODO:
* [x] list safe symbols so all scripts terminate
* [ ] think of the packaging and defaults (ok to depend on `sci` from core)?
* [x] support raw function forms, if https://github.com/borkdude/sci/issues/33 is implemented, great for reading from edn-sources
* [ ] use `eval` on JVM if the form is safe (better perf)

Tip from @borkdude to see if we are on Graal:

```clj
(def native? (= "executable" (System/getProperty "org.graalvm.nativeimage.kind")))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,32 @@ Detailed errors with `m/explain`:
 ;          {:path [3 1 4 1 2], :in [:address :lonlat 1], :schema double?, :value nil})}
 ```
 
+Using inline (serializable) functions:
+
+```clj
+(m/explain
+  [:and
+   [:map
+    [:x int?]
+    [:y int?]]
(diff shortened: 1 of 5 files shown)


jeans11:
Is there a `:cljs`?

ikitommi:
Could be `:script` or `:fn` or something else, this works on both clj & cljs

ikitommi:
ship it, it's all experimental anyways.

PR #50, opened 2019-08-28
correctly validate false map values

loomis:
The current code does not correctly validate map values when they evaluate to a false (`false` or `nil`) value.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -174,8 +174,7 @@
                                (fn [[key {:keys [optional]} value]]
                                  (let [valid? (-validator value)
                                        default (boolean optional)]
-                                   (fn [m] (let [v (key m ::missing-key)]
-                                             (if (not= v ::missing-key) (valid? v) default)))))
+                                   (fn [m] (if-let [map-entry (find m key)] (valid? (val map-entry)) default))))
                                entries)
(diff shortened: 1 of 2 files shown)


src/malli/core.cljc
```diff
@@ -174,7 +174,8 @@
                                (fn [[key {:keys [optional]} value]]
                                  (let [valid? (-validator value)
                                        default (boolean optional)]
-                                   (fn [m] (if-let [v (key m)] (valid? v) default))))
+                                   (fn [m] (let [v (key m ::missing-key)]
+                                             (if (not= v ::missing-key) (valid? v) default)))))
```
ikitommi: Good catch! Equality for Keywords is slow on Clojure, I recall `identical?` is faster. Still, here we could use `find` to get the whole map entry, it's `nil` or `Map$Entry` and you can get the value effectively with `val`.

ikitommi: If you could change that, would be happy merge this in.

loomis: I've used `find` as that keeps the logic simpler and closer to the original.

loomis: Let me know if you'd like further changes.

ikitommi:
Looking good, thanks!

Issue #43, opened 2019-08-19
Allow additional entries in :map schemas


ikitommi:
Prismatic Schema allows one to mix defined keys and extra keys in a same map. We could get rid of `:map-of` in favour of supporting predicate keys in `:map`:

```clj
(m/schema
  [:map
   [:id int?]
   [:name string?]
   [int? int?]])
```

ikitommi:
related to https://github.com/metosin/malli/issues/31#issuecomment-557893019

rschmukler:
My leaning on this would be to keep them separate. A few questions that come to mind as "undefined" behaviors if we do merge them:

1) What happens with something like the following? Is it an `:or`?

clj
```
[:map 
  [int? int?]
  [string? string?]]
```

2) What about an open map schema that specifies key predicates? Is it valid if it has non-conforming key-value pairs, so long as both the key and value don't conform?

3) What happens if you call `merge` on two map-of-like schemas?

To me, although the underlying data structure is similar, the behavior seems quite different - similar to `:tuple` and `:vector`. 

Also, a user who wants to add arbitrary predicates to a map schema could do so by writing a `:fn` check. It seems infrequent enough of a need (a map schema, which also conforms to a `map-of` schema for its "open" values) that I'm not sure if it's worth introducing potentially confusing behavior.

ikitommi:
1. Like JSON & Plumatic Schema, only one extra entry would be allowed per map
2. Good point, these should be either combined into single key or add rules how to handle cases where both exist. related comments https://github.com/metosin/malli/issues/31#issuecomment-557892061 & https://github.com/metosin/malli/issues/31#issuecomment-557904828. Non trivial anyway :(
3. Another good point. current merge (or union) doesn't handle these. Maybe:

```clj
(mu/union [:map-of int? string?] [:map-of string? int?])
; => [:map-of [:or int? string?] [:or string? int?]]
```

Having a formal syntax for this would allow malli to be more JSON Schema compliant, generating ` additionalProperties` from `:fn` might not be fun.

ikitommi:
Collecting alternatives from various sources:

### 1) don't support

* use `:fn` to handle extra keys

### 2) special key

```clj
[:map 
  [int? int?]
  [:malli.core/extra [string? string?]]]
```

### 3) special property

```clj
[:map {:extra [string? string?]}
  [int? int?]]
```

ikitommi:
More bad ideas, bloat `:closed` to support extras:

```clj
[...]

PR #52, opened 2019-08-29
Merge map-of into map

jeans11:
#43 

```clojure
(m/validate [:map keyword? int?] {:x 1 :y 2})
;; true

(m/validate
  [:map
    [:x string?]
    [[:opt :y] int?]
    [keyword? [:vector int?]]]
  {:x "x"
   :z [1 2 3]})
;; true

(m/validate [:map [:x int?] [keyword? string?]] {:x 1 :z 2})
;; false
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -141,6 +141,26 @@
     [(first xs) (rest xs)]
     [nil xs]))
 
+(defn- -valid-schema? [schema?]
+  (try
+    (schema schema?)
+    true
+    (catch Exception _ false)))
+
+(defn- -conform-childs [childs]
+  (if (and (= 2 (count childs)) (every? -valid-schema? childs))
+    [childs]
+    childs))
+
+(defn- -is-pred-key?
+  [entry]
+  (-> entry meta :pred-key?))
+
+(defn- -only-map-of?
+  [entries]
+  (and (= 1 (count entries))
+       (-is-pred-key? (first entries))))
+
 (defn- -expand-key [[k ?p ?v] opts f]
   (let [[p v] (if (map? ?p) [?p ?v] [nil ?p])
         v' (f (schema v opts))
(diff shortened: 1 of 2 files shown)


test/malli/core_test.cljc
```diff
@@ -204,6 +204,18 @@
...
+          schema4 (m/schema
+                    [:map
+                     [keyword? int?]
+                     [:x string?]
+                     [:y int?]])
+          schema5 (m/schema
+                    [:map
+                     [:x string?]
+                     [[:opt :y] int?]
+                     [keyword? int?]])
```
ikitommi: there should be a test that for multiple predicate keys too. Schema explicitly only allows one extra key per map, I think we might not need that restriction.

src/malli/core.cljc
```diff
@@ -141,40 +141,73 @@
...
+(defn- -conform-childs [childs]
+  (if (and (= 2 (count childs)) (every? -valid-schema? childs))
+    [childs]
+    childs))
+
+(defn- -is-pred-key?
+  [entry]
+  (-> entry meta :pred-key?))
+
+(defn- -only-map-of?
```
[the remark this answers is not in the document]
jeans11: Good idea!

test/malli/core_test.cljc
jeans11: What do you mean?

ikitommi: My bad, was thinking [Plumatic schema](https://github.com/plumatic/schema), which has the limitation - just one predicate key per map

jeans11: Allow something like that?
```clojure
(m/schema
  [:map
    [keyword? int?]
    [keyword? string?]])
```

ikitommi: Let's not. That can already be defined as:

```clj
[:map
  [keyword? [:or int? string?]]
```
... and this would cause trouble when later doing tools to merge maps and removing `keyword?` key might or might not remove all keys. Same actually applies to normal keys, this should be illegal:

```clj
[:map [:x int?] [:x string?]]
```

ikitommi:
Thanks! I'll merge this in and will check the perf & think about corner cases.

ikitommi:
perf regressed for `[:map [:x int?] [:y boolean?]]` from 17ns to 110ns (6x). I need this for the generator branch, so will do some perf tuning on top of this.

ikitommi:
There is actually ambiguity rule here: 

Given a schema:

```clj
[:map
  [:a int?]
  [:b string?]
  [::id string?]
```

One could have binded the `::id` into a defined value via a registry, so it's a valid schema. The code in this PR doesn't resolve `::id` as a schema as the code doesn't pass options into the `-valid-schema?` function. That's a bug. If that is fixed, the `::id` would be resolved into a schema and it's value is used instead `::id` as a key => there is no way to represent key `::id` in the map, which is not acceptable. Also, few commits ago, I made the `IntoSchema` work with `malli.core/schema` too, so:

```clj
(m/schema :tuple)
; :tuple
```  

=>  `:tuple` would be illegal key in any map.

Will remove the changes and put the commits aside until resolved. Need to think this over.

ikitommi:
Master is reset back before this, and these changed are stored in branch https://github.com/metosin/malli/tree/map-of-with-map

PR #62, opened 2019-09-08
Implement m/merge

### LINKED ISSUE #45: Programming with Schemas, part1
Programming schemas should be first-class. [schema-tools](https://github.com/metosin/schema-tools) is a good example of a great library in this space. We could start with `m/merge` which does potetially lossy merge like the `clojure.core/merge` does.

Default (overwriting keys):

```clj
(m/merge
  [:map
   [:x int?]
   [:y int?]]
  [:map
   [:x int?]
   [:y {:optional true} string?]]
  {:strategy :overwrite})
; [:map
;  [:x int?]
;  [:y {:optional true} string?]]
```

Compose leaves:

```clj
(m/merge
  [:map
   [:x int?]
   [:y int?]]
  [:map
   [:x int?]
   [:y {:optional true} string?]]
  {:strategy :merge})
; [:map
;  [:x int?]
;  [:y [:or int? string?]]]
``` 

Needs some discussion how to do this right.

Related: #37

ikitommi:
... one of the original reasons for `malli` was to be able deep-merge map-schemas easily. Not perfect (doesn't understand how to merge things like `:and` & `:or`), but a start. Code is kinda bloaty.

Will fix: #45 

```clj
(m/merge
  [:map 
   [:parameters
    [:map
     [:query-params 
      [:map [:x int?]]]]]]
  [:map 
   [:parameters
    [:map
     [:query-params 
      [:map [:y int?]]]
     [:body-params
      [:map [:z int?]]]]]])
;[:map 
; [:parameters
;  [:map
;   [:query-params 
;    [:map [:x string?] [:y int?]]]
;   [:body-params
;    [:map [:z int?]]]]]]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,31 @@ Transformers are composable:
 ;           :lonlat [61.4858322 23.7854658]}}
 ```
 
+## Merging Schemas
+
+Schemas can be deep-merged with `m/merge`:
+
+```clj
+(m/merge
+  Address
+  [:map
+   [:description string?]
+   [:address
+    [:map
+     [:country string?]]]])
+;[:map
+; [:id string?]
+; [:tags [:set keyword?]]
+; [:address 
+;  [:map 
+;   [:street string?] 
+;   [:city string?] 
+;   [:zip int?] 
+;   [:lonlat [:tuple double? double?]] 
+;   [:country string?]]]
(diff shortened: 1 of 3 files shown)


ikitommi:
Will go to the big internal review..

PR #67, opened 2019-09-11
Welcome :re schemas!

ikitommi:
```clj
(def Email 
  [:re #"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,63}$"])

(m/validate Email "tommi@kikka.com")
; => true

(-> Email (m/serialize) (m/deserialize) (m/validate "tommi@kikka.com"))
; => true

(mg/generate Email {:seed 42, :size 10})
; => "CaR@MavCk70OHiX.yZ"
```

Question: Should the `Pattern` implement `IntoSchema` so it could be used without the`:re` schema tag? Would be always on...

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,10 @@ Scehmas can be used to generate values:
 (mg/generate pos-int? {:seed 10, :size 100})
 ;; => 55740
 
+;; regexs work too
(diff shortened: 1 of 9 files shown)


ikitommi:
ClojureScript doesn't support regexs: https://github.com/gfredericks/test.chuck/issues/46 :(

PR #68, opened 2019-09-12
Regex schemas work unwrapped

ikitommi:
Making Regexs implement the `IntoSchema` so we can use them without `:re` tag. Is this a good idea? One can't turn off this support, it would be always on.

```clj
(def Email #"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,63}$")

(m/validate Email "tommi@kikka.com")
; => true

(-> Email (m/serialize) (m/deserialize) (m/validate "tommi@kikka.com"))
; => true

(mg/generate Email {:seed 42, :size 10})
; => "CaR@MavCk70OHiX.yZ"
```

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -27,8 +27,8 @@
                              "-Dclojure.compiler.direct-linking=true"]}}
  :deps {org.clojure/clojure {:mvn/version "1.10.1"}
         borkdude/sci {:git/url "https://github.com/borkdude/sci"
-                      :sha "939cfde3143c84e2134caaff1b5640510e6096f6"}
+                      :sha "e73331e74322b6c6a823ed487a724caec3f28a9e"}
         borkdude/edamame {:git/url "https://github.com/borkdude/edamame"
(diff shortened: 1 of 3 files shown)


ikitommi:
A pluggable way to have this on by default, but to be able to disable it would be to make the `s/schema` first look the registry with a `m/name`, then with the `class` of the `?schema` argument. Adding a separate `m/class-registy` (to be merged in  to `m/default-registry`) with entry of `[java.util.regex.Pattern (m/-re)]` would make regexs work oob by default and it would be easy to turn off support.

explicit configuration with sane defaults for the win :)

nilern (approved):
This is what we want. Just need to decide whether this is on by default, similarly to e.g. `int?` as a schema. Shortcuts like this are pleasant but unavoidably inconsistent. Regexp is more (de)serializable than functions but every language has slightly different regexps, so YMMV.

Issue #58, opened 2019-09-06
Add a custom target for transforming Map keys


ikitommi:
e.g. virtual `:malli.core/key` schema key, that can be attached with a transformer. This allows the keys to be transformed from `kebab-case` to `snake_case` etc.

```clj
(require '[malli.core :as m])
(require '[malli.transform :as mt])

(m/transform
  [:map
   [:a-key int?]
   [:another-key string?]]
  {:a_key 1, :another_key "2"}
  (mt/key-transformer kebab-case))
; => {:a-key 1, :another-key "2"}``

jeans11:
Do you wants use this lib? https://github.com/clj-commons/camel-snake-kebab

ikitommi:
We don't want extra dependencies to malli, this would be just an helper to plug in map-key transformations. Clients of malli could use the `camel-snake-kebab`, been using that myself.

jeans11:
Hi!
What do you mean with `:malli.core/key`?

ikitommi:
Transformers are currently of type `name -> schema -> f` where `f` is the actual 1arity transformation fn. Now as transformers can be merged, it's easy to override a transformation fn for a given schema. There is no `name` for map keys, but we should have that -> if present, all map keys are transformed with it.

PR #70, opened 2019-09-13
Keys transformer

jeans11:
#58

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -231,51 +231,73 @@
                       (explainer x in acc))
                     acc explainers)))))
           (-transformer [this transformer]
-            (let [transformers (some->> entries
-                                        (mapcat (fn [[k _ s]]
-                                                  (let [key-transformer (-transformer (map-key k) transformer)
-                                                        value-transformer (-transformer s transformer)
-                                                        noop-transformer (fn [x] x)]
-                                                    (cond
-                                                      (and key-transformer value-transformer)
-                                                      [k [key-transformer value-transformer]]
-
-                                                      (and (not key-transformer) value-transformer)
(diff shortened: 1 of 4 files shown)


ikitommi:
Could you run the performance test before and after the change? in case there are no key-transformers defined, we should not get any performance penalty.

https://github.com/metosin/malli/blob/master/perf/malli/perf_test.cljc#L266

jeans11:
Ok so performance issue found. I investigate.

jeans11:
Performance should be ok now

src/malli/core.cljc
```diff
@@ -233,10 +234,30 @@
...
+                                                        noop-transformer (fn [x] x)]
+                                                    (cond
+                                                      (and key-transformer value-transformer)
+                                                      [k [key-transformer value-transformer]]
+
+                                                      (and (not key-transformer) value-transformer)
+                                                      [k [noop-transformer value-transformer]]
+
+                                                      (and key-transformer (not value-transformer))
+                                                      [k [key-transformer noop-transformer]]))))
```
ikitommi: I think all keys should be transformed, not just the ones defined in the schema. e.g. the map might be open and the value can have extra keys, which should be transformed too.

Given that, I think the key-transformation could be an extra step, just like the `map-transformer`. With this, the performance for not using that would be identical as before. If both the `map-transformer` & `key-transformer` are present, there could be a max of 3 steps. I think it's ok, can't think of a way doing it faster.

jeans11:
Anything I can do to help this along?

ikitommi:
looks great, sorry for taking so long. Thanks!

ikitommi:
What do you think about https://github.com/metosin/malli/issues/91?

Issue #71, opened 2019-09-13
[BUG] Transform map add optional keys


jeans11:
```clojure
(m/transform
  [:map [:x int?] [[:opt :y] string?]]
  {:x "1"}
  transform/string-transformer)
;; {:x 1 :y nil}
```

Should check that key exist?

PR #72, opened 2019-09-13
Not add optional key

jeans11:
#71


ikitommi:
Good catch, thanks!

ikitommi:
Changed the sequence destructuring into calling `val`, which uses direct java method access and is bit faster, see https://github.com/metosin/malli/commit/26480ae1ebc18e4deb448241c6b74f883ae28089

PR #77, opened 2019-10-03
Fix returning acc in enum explainer

kmwang:
`enum` spec might cause collection explainers can't produce errors report properly. To reproduce the problem:
```
(do (require '[malli.core :as m])
    (m/explain [:map [:x [:fn '(constantly false)]] ; `explain` should complain about this.
                     [:a [:enum "A" "B"]]]
               {:a "A" :x "123"}))  ; but here returns `nil` instead
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -469,7 +469,8 @@
                 (conj acc {:path path
                            :in in
                            :schema this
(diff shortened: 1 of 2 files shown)


ikitommi:
Awesome, thanks!

PR #83, opened 2019-10-07
Drop support for schema-like syntax, fixes #81

### LINKED ISSUE #81: Drop support of Prismatic Schema like keys
Instead of:

```clj
(m/validate
  [:map
   [:x boolean?]
   [[:opt :y] int?]
   [[:req :z] string?]]
  {:x true, :z "kikka"})
; => true
```
Just use the default syntax:

```clj
(m/validate
  [:map
   [:x boolean?]
   [:y {:optional true} int?]
   [:z string?]]
  {:x true, :z "kikka"})
; => true
```

ikitommi:
See #81

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,20 +43,6 @@ Definining and validating Schemas:
 ; => true
 ```
 
-Optional [Schema-like](https://github.com/plumatic/schema) syntax for `:map` keys:
-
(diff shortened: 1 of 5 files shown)


ikitommi:
byebye

PR #90, opened 2019-10-21
Fix build and update readme

jeans11:


ikitommi:
Thanks!

ikitommi:
Modified the name to be qualified so that it doesn't clash with user defined schemas, see https://github.com/metosin/malli/commit/9af89a96bb4da23b5d0606b76f452def8ed54f84

Issue #91, opened 2019-10-22
keys-transformer doesn't convert keys not defined in schemas


ikitommi:
Currently:

```clj
(m/transform
  [:map [:x int?]]
  {:x 1, :y 2}
  (transform/key-transformer
    #(-> % name (str "_key") keyword)))
; => {:y 2, :x_key 1}
```

I would expect all keys to be transformed.

ikitommi:
FIxed by #93

PR #93, opened 2019-10-22
key-transformer should transformed all keys

jeans11:
#91 
So let `mt/key-transformer` to transform all keys in the map. 

If we want to transform only keys in the schema, compose `mt/strip-extra-keys-transformer` and `mt/key-transformer`. Ok with that @ikitommi?


ikitommi:
Thanks!

PR #97, opened 2019-10-25
Programming with schema

jeans11:
Add functions:
- `select-keys`
- `assoc`
- `dissoc`
- `update-in`

Few questions:

Should we add `update` too?

I don't know how to handle `properties`:

`(m/assoc [:map [:x int?]] :y {:optional true} string?)`
or
`(m/assoc [:map [:x int?]] [:y {:optional true} string?])`

@ikitommi what do you think about it?

Perf of functions:

`m/select-keys`
Screenshot 2019-10-25 at 11 52 14

`m/dissoc`
Screenshot 2019-10-25 at 11 53 13

`m/assoc`
Screenshot 2019-10-25 at 11 55 28

`m/update-in`
Screenshot 2019-10-25 at 11 56 10

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -715,6 +715,56 @@
     (-form [_] ::map-key)
     (-transformer [this transformer] (-value-transformer transformer this))))
 
+(defn- -equals? [a b]
+  #?(:clj  (.equals a b)
+     :cljs (= a b)))
+
+(defn- -assert-map-schema [s]
+  (when-not (-equals? (-name s) :map)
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks! Need to think about this.

ikitommi:
Still need to think how this should work. We have the `visitor` that knows how to walk over anything.

ikitommi:
I think this should be re-thinked as we now have `malli.core/MapSchema`. Should also work with Sequential Schemas. I think we need to get back to the drawing board. Sorry, but will close this.

Issue #98, opened 2019-10-27
All schemas should target transformations to themselves


ikitommi:
Currently, the following Schemas do not have `this` as a transformation target, e.g. adding a transformation hint via properties will not work with these:

* [x] `:and`, `:or`
* [x] `:map-of`
* [x] `:vector`, `:list`, `:set`
* [x] `:tuple`
* [x] `:enum`
* [x] `:re`
* [x] `:fn`
* [x] `:maybe`

This doesn't work currently (`:and` doesn't have a transformation target):

```clj
(m/encode
   [:and {:decode/string '(constantly str/upper)} string?]
   "kikka"
    mt/string-transformer)
; => "kikka"
```

But this does (`string?` has):

```clj
(m/encode
   [string? {:decode/string '(constantly str/upper)}]
   "kikka"
    mt/string-transformer)
; => "KIKKA"
```

As transformation is a primary use case, we should unroll the functions for performance. @jeans11 did this already for `:map`.

boxxxie:
when is `m/encode`? i think it may have moved to a different namespace from core. i can't run the examples

PR #99, opened 2019-10-27
Separate encode & decode + schema-hinted transformations

ikitommi:
* `m/transform` => `m/encode` & `m/decode`
* Support for schema-driven transformations
* #98 should be fixed with this of right after.

```clj
(deftest schema-hinted-tranformation
  (let [schema [string? {:title "lower-upper-string"
                         :decode/string (constantly str/upper-case)
                         :encode/string (constantly str/lower-case)}]
        value "KiKkA"]
    (testing "defined transformations"
      (is (= "KIKKA" (m/decode schema value mt/string-transformer)))
      (is (= "kikka" (m/encode schema value mt/string-transformer)))
      (is (= "kikka" (as-> value $
                           (m/decode schema $ mt/string-transformer)
                           (m/encode schema $ mt/string-transformer)))))
    (testing "undefined transformations"
      (is (= value (m/decode schema value mt/json-transformer)))
      (is (= value (m/encode schema value mt/json-transformer))))))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,12 +234,12 @@ Errors can be targetted using `:error/path` property:
 
 ## Value Transformation
 
-Schema-driven value transformations with `m/transform`:
+Schema-driven value transformations with `m/decode` and `m/encode`:
 
 ```clj
 (require '[malli.transform :as mt])
 
-(m/transform
+(m/decode
   Address
   {:id "Lillan",
    :tags ["coffee" "artesan" "garden"],
@@ -259,7 +259,7 @@ Schema-driven value transformations with `m/transform`:
 Transform map keys with `mt/key-transformer`:
 
 ```clj
-(m/transform
+(m/decode
   Address
   {:id "Lillan",
    :tags ["coffee" "artesan" "garden"],
@@ -284,7 +284,7 @@ Transformers are composable:
     mt/strip-extra-keys-transformer
     mt/json-transformer)
 
-(m/transform
+(m/decode
   Address
   {:id "Lillan",
    :EVIL "LYN"
@@ -591,7 +591,7 @@ Coercion:
 
 ;; 140ns
 (let [schema [:map [:id int?] [:name string?]]
-      transform (m/transformer schema transform/string-transformer)]
+      transform (m/decoder schema transform/string-transformer)]
   (cc/quick-bench
     (transform {:id "1", :name "kikka"})))
 ```
--- a/README.md
+++ b/README.md
@@ -333,16 +333,16 @@ Schemas can be deep-merged with `m/merge`:
 Writing and Reading schemas as [EDN](https://github.com/edn-format/edn), no `eval` needed.
 
 ```clj
-(require '[malli.edn :as me])
+(require '[malli.edn :as edn])
 
 (-> [:and
      [:map
       [:x int?]
       [:y int?]]
      [:fn '(fn [{:keys [x y]}] (> x y))]]
-    (me/write-string)
+    (edn/write-string)
     (doto prn) ; => "[:and [:map [:x int?] [:y int?]] [:fn (fn [{:keys [x y]}] (> x y))]]"
-    (me/read-string)
+    (edn/read-string)
     (doto (-> (m/validate {:x 0, :y 1}) prn)) ; => false
     (doto (-> (m/validate {:x 2, :y 1}) prn))) ; => true
 ;[:and 
--- a/deps.edn
+++ b/deps.edn
@@ -28,7 +28,8 @@
                              "-Dclojure.compiler.direct-linking=true"]}}
  :deps {org.clojure/clojure {:mvn/version "1.10.1"}
         borkdude/sci {:git/url "https://github.com/borkdude/sci"
-                      :sha "a1f6c2204bbfa0c191dbb99efef99709dc454fbe"}
-        borkdude/edamame {:mvn/version "0.0.6"}
+                      :sha "a66dd0dce8edab65627b86f6388100740e36bd66"}
+        borkdude/edamame {:git/url "https://github.com/borkdude/edamame"
+                          :sha "34755e7a875fae2ab4d64e6be89b6270046a91a3"}
         org.clojure/test.check {:mvn/version "0.10.0"}
         com.gfredericks/test.chuck {:mvn/version "0.2.10"}}}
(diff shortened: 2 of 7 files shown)


boxxxie:
This looks pretty great.
you should keep the m/transform function, though. this is going to break a lot of people's code, and i think it's possible to avoid that. you can add a deprecated doc/meta to the m/transform function.

ikitommi:
Status of malli is *Pre-alpha, in design and prototyping phase.*, so we can break things to get a clear initial design for a first alpha. So, not going to keep deprecated functions. If this change would happen after a release, deprecation would be a good option.

RokLenarcic:
I am concerned by this particular line:

https://github.com/metosin/malli/pull/99/files#diff-f4026c83e07e6f542344d5945d310607R25

This seems to try to get encoder/decoder from schema keys like `:decode/string` and if that is missing then try to get an encoder/decoder from the options map, keys `:encoders` and `:decoders`. 

1. Shouldn't the order in `or` be reversed? As it stands, if a schema has an encoder/decoder defined by keyword like `:decode/string`, I cannot override it by passing a different encoder/decoder in the options map. I believe it would be better to try to find one in the options map first, which would allow a developer to change encode/decode contextually by passing different implementations in the options map.

2. The value from a key like `:decode/string` is passed into `m/eval`, but the value from options map `encoders` and `decoders` key is not passed through `m/eval`. Is this right? Seems to be inconsistent, please check.

ikitommi:
Thanks for you thoughts. 

1) This is the order used in [spec-tools](https://cljdoc.org/d/metosin/spec-tools/0.10.0/doc/spec-coercion#spec-based-transformations) and it was then discussed to work this way. If the default branch would be used first, there would be no way to override any of the encode/decode that have base/default implementation, causing hard-to-debug errors:

```clj
;; encode would do nothing here (the default would be used)
[int? {:encode (constantly (comp inc mt/string->long))}] 

;; encode MIGHT work, depending if there is default `:encode` for `:and` or not
[:and {:encode (constantly (comp inc mt/string->long))} int?] 
```

2) the first branch uses `m/eval` as the data is read from schema properties, which can be serialized. the default branch doesn't `m/eval`, as the data originates from `Transformer`, which is a local concept. Currently, serializing Transformers is not supported.

RokLenarcic:
> Thanks for you thoughts.
> 
> 1. This is the order used in [spec-tools](https://cljdoc.org/d/metosin/spec-tools/0.10.0/doc/spec-coercion#spec-based-transformations) and it was then discussed to work this way. If the default branch would be used first, there would be no way to override any of the encode/decode that have base/default implementation, causing hard-to-debug errors:
> 
> ```clojure
> ;; encode would do nothing here (the default would be used)
> [int? {:encode (constantly (comp inc mt/string->long))}] 
> 
> ;; encode MIGHT work, depending if there is default `:encode` for `:and` or not
> [:and {:encode (constantly (comp inc mt/string->long))} int?] 
> ```
> 

Sorry, but I don't follow the reasoning. You call it the default branch, but the maps of encoders/decoders that you use when constructing transformers is keyed by schema name, which makes it just as specific as specifing encoder/decoder on the schema definition itself.

 In fact the branch you call default is more specific, since I can provide a different encoder/decoder for a particular schema name and transformer name every time I construct a transformer (which enables me to adjust encoder/decoder implementation based on a config file for instance). When you define encoder/decoder directly on schema itself, you can only have 1 implementation per schema name+transforer name pair, so it's more of a default behaviour.

 I mean, how is defining encoder/decoder on the schema itself override for default behaviour, since the definitions on both sites have the same scope (one implementation defined per transformer+schema names pair)? Defining encoder/decoder on the schema itself effectively deletes the same entry from the map you pass to transformer construction. And people are much more likely to vary transformations by constructing transformers with varying options than by passing schemas for same concept with varying options.

ikitommi:
> And people are much more likely to vary transformations by constructing transformers with varying options than by passing schemas for same concept with varying options.

Could you provide an example use case for this?

Problem with transformer options is that they work on schema types, not instances. If you have a transforming function defined for `:map`, it effects all maps, for example with `mt/strip-extra-keys-transformer`. If you have some map schema instances, that should transform differently, you are out of luck if the transformer option would win over schema defined one.

If, for some reason, one would need to transform data for example from JSON using two different ways (normal and legacy) one could create a new transformer with a new name, e.g. `:json_legacy` and it would look for `:encode/json_legacy` from the schema properties, not likely to exist - and the transformer defined one would be used instead. It that property existed in schema, the schema designer knows better and that is used instead.

Related: I was thinking one day about adding a support for chaining transformers easily into one. e.g. `strip-extra-keys` + `upper-case-keys`, both effecting `:map`. Currently, one needs to run them separately which is not optimal. Both transformers could have a separate name, and one could end up having a chain of (named) transformation functions per map schema instance. Schemas could override any of these transformations, using properties.

RokLenarcic:
> Could you provide an example use case for this?

An example would be when you have a transformator that transforms java.time object into a locale specific format, where locale is supplied at runtime (cannot be part of schema definition). Something like this.

ikitommi:
Thanks for the example! 

To support that use case better, I added support for optional `:opts` map for the transformers, see  https://github.com/metosin/malli/pull/99/commits/76744a4403b8d2982285d4b813f903c45bd51391. With that, one can customize the transforming functions using the transformer options. In your example, there could be a `:date-format` option that the date-formatting transformers use. One could still override/break this behaviour via Schema properties.

What do you think?

```clj
(deftest options-in-transformaton
  (let [schema [:and int? [any? {:decode/string '(fn [_ {::keys [increment]}] (partial + (or increment 0)))}]]
        transformer (mt/transformer mt/string-transformer)
        transformer1 (mt/transformer mt/string-transformer {:opts {::increment 1}})
        transformer1000 (mt/transformer mt/string-transformer {:opts {::increment 1000}})]
    (is (= 0 (m/decode schema "0" transformer)))
    (is (= 1 (m/decode schema "0" transformer1)))
    (is (= 1000 (m/decode schema "0" transformer1000)))))
```

ikitommi:
Fix for #98 included in this PR

ikitommi:
Merging this. All decisions will be revisited before initial release, tagged in #116.

eoliphan:
Just an FYI, this isn't checked off in #116

ikitommi:
Good catch

PR #106, opened 2019-11-01
allow generator with unsupported features to be overridden

loomis:
If a regular expression contains a feature that isn't supported by test.chuck, then an exception will be thrown even if a `:gen/elements` value is provided. The PR allows the default generator to be overridden in this case.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -47,12 +47,9 @@
 
 #?(:clj
    (defn -re-gen [schema opts]
-     (let [re (or (first (m/childs schema opts)) (m/form schema opts))]
-       (try
-         (gen2/string-from-regex (re-pattern (str/replace (str re) #"^\^?(.*?)(\$?)$" "$1")))
-         (catch Exception ex
-           #(throw ex)                                      ;; throw only if bad regex generator is actually used
-           )))))
+     (when-not (-> schema m/properties :gen/elements)
+       (let [re (or (first (m/childs schema opts)) (m/form schema opts))]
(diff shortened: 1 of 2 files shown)


src/malli/generator.cljc
```diff
@@ -48,7 +48,11 @@
 #?(:clj
    (defn -re-gen [schema opts]
      (let [re (or (first (m/childs schema opts)) (m/form schema opts))]
-       (gen2/string-from-regex (re-pattern (str/replace (str re) #"^\^?(.*?)(\$?)$" "$1"))))))
```
ikitommi: I think it would be better to return `nil` in case of failure. With this, the `m/generator` would fail fast on creation (as no generator is found), not at runtime when invoked. Not perfect as the error message is the generic one, instead of the actual `test.chuck` one, but still, better IMO.

ikitommi: or, actually - the `gen/elements` should be checked first, if it exists, the generator would not be looked from the schema.

loomis: @ikitommi I've added an explicit check for `:gen/elements`. If present, nil will be returned for the generator.

ikitommi: Thanks! Currently, the `:gen/elements` is already read in `-create`, could all the changes be applied there? e.g. re-order so that if the elements exists, the `-generator` would not be called at all. This would enforce the ordering of `elements` first for all schemas.

loomis: Sure. Done.

ikitommi:
Awesome, thanks!!

PR #112, opened 2019-11-05
Pass opts through to `schema` / `generator` calls

plexus:
Fix several places where recursive calls to `generator` or `schema` were not
passing through the opts map, thus breaking use of a custom registry.


ikitommi:
Thanks!

PR #113, opened 2019-11-05
s/child/children and pass through opts

plexus:
Rename "childs" to "children", since that is the standard plural for "child".

In `generator` fix more places where `opts` wasn't being passed through, mainly to `m/children` and `m/properties`.

Also fix a callsite to `-coll-distinct-gen` I forget to update in the previous PR which caused the tests to break. They're green again.

Sorry for putting these in one PR, was a bit tedious to split up because I did the rename first.


ikitommi:
Thanks!

PR #119, opened 2019-11-24
feat: call map-key w/ parent-schema information

rschmukler:
This commit passes the parent-schema (ie. the schema of the map) to get the transformer function used in map-keys.

This is useful if your map-keys wants to introspect the schema to determine
its behavior. An example usecase would be automatically remove the most
prevelant namespace of a key, and adding a string underscore encoding
for other namespaces. Eg. you can use this to write a json encoder that
does something like the following:

```clj
{:person/name "Bob" :person/age 36 :pet/name "Edgar"}
;; =>
{"name" "Bob" "age" 36 "pet_name" "Edgar"}
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -231,7 +231,7 @@
                       (explainer x in acc))
                     acc explainers)))))
           (-transformer [this transformer context]
-            (let [key-transformer (-transformer (map-key) transformer context)
+            (let [key-transformer (-transformer (map-key this) transformer context)
                   value-transformers (some->> entries
                                               (mapcat (fn [[k _ s]] (if-let [t (-transformer s transformer context)] [k t])))
                                               (seq)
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks for the PR. I'm not happy about this breaking the contract of having `this` as an argument to the transforming function (father schema would be used instead).

There is a discussion about having a link to the parent from all schemas, which would, I believe, solve this without breaking the contract.

Collected info into an new issue #120

rschmukler:
Makes sense! It definitely felt dirty but since `map-key` is already a bit of a hack (eg. not a truly traversed schema entry, nor `IntoSchema` I thought it might be acceptable.

For my use case, I can get away with not having it once #122 lands - moving the functionality into the `:after` of the `:map` interceptor. I would imagine that for many use cases, the interceptors solve needing parent information. Closing this for now.

Issue #114, opened 2019-11-11
Support :enter and :leave for transformers


ikitommi:
Currently, the schema-defined transforming functions are all applied before any other transformations. There should be a way to define the order they are run. Interceptors to the rescue?

## currently

```clj
(m/decode 
  [:and {:decode/string (constantly (partial str "olipa_"))} string?] 
  "kerran" 
  mt/string-transformer)
; "olipa_kerran"
```

## with interceptors

```clj
;; function maps to :enter
(m/decode 
  [:and {:decode/string (constantly (partial str "olipa_"))} string?] 
  "kerran" 
  mt/string-transformer)
; "olipa_kerran"

(m/decode
  [:and {:decode/string (constantly {:enter (partial str "olipa_")})} string?]
  "kerran"
  mt/string-transformer)
; "olipa_kerran"

(m/decode
  [:and {:decode/string (constantly {:leave (partial str "_ihminen")})} string?]
  "kerran"
  mt/string-transformer)
; "kerran_ihminen"

(m/decode
  [:and {:decode/string (constantly {:enter (partial str "olipa_")
                                     :leave (partial str "_ihminen")})} string?]
  "kerran"
  mt/string-transformer)
; "olipa_kerran_ihminen"
```

ikitommi:
maybe even support multiple interceptors via vector syntax?

```clj
(m/decode
  [:and {:decode/string '(constantly [str (partial str "id_") #(str % "23") keyword]} keyword?]
  "1"
  mt/string-transformer)
; :id_123
```

PR #122, opened 2019-11-24
feat: interceptor based transformers

rschmukler:
See the idea surfaced in #114.

This commit introduces interceptor based transformers by using the
`:enter` and `:leave` keys returned on the `-value-transformer`
function.

Naively it uses almost identical implementations for all enter and exit
applications, with the only exception being the collection-based
transformers which do look at the phase to avoid double wrapping
collections.

Ultimately there might be some occassions where we could speed things up
even more (eg. not reducing over a collection if there isn't a mapping
function anyway) but those optimizations don't exist in the current
implementation either, so this PR doesn't attempt to resolve any of
that.


ikitommi:
Looks good, I'll add some tests see if something is missing. Thanks!

PR #130, opened 2019-12-04
fix: custom registry on encode / decode

rschmukler:
Fixes an error where a schema would be resolved before options were
taken into account on `encode` and `decode` - causing custom registries
to be ignored.

Since `(schema ?schema opts)` is called in `encoder` and `decoder` it seemed redundant to call it at all in the outer functions.


ikitommi:
Thanks!

PR #132, opened 2019-12-05
feat: :map ::m/rename-keys transformer metadata

rschmukler:
Introduces the ability for map transformers to provide information about key-renames they perform by including the `::m/rename-keys` key-map in metadata attached to either a run-time value (ie. the data returned by the map transformer) or on the returned
interceptor functions themselves.

This allows map transformers to not break subsequent calls to child
value transformers.

Initially I tried to solve this generically with a `:leave` interceptor
in the map transformer, but this broke down with nested maps since
`:after` is currently executed as a prewalk.

It may be useful to also change the behavior of `:leave` to behave more
like a postwalk, but that requires more thought and is certainly a much
larger PR.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1,6 +1,7 @@
 (ns malli.core
   (:refer-clojure :exclude [-name eval name merge])
-  (:require [sci.core :as sci])
+  (:require [sci.core :as sci]
+            [clojure.set :as set])
   #?(:clj (:import (java.util.regex Pattern))))
 
 ;;
@@ -237,13 +238,15 @@
           (-transformer [this transformer context]
             (let [build-transformer
                   (fn [i-key]
-                    (let [key-transformer (i-key (-transformer (map-key) transformer context))
+                    (let [map-transformer (i-key (-value-transformer transformer this context))
+                          compile-time-key-renames (some-> map-transformer meta ::rename-keys)
+                          key-transformer (i-key (-transformer (map-key) transformer context))
                           value-transformers
                           (some->> entries
                                    (mapcat (fn [[k _ s]] (if-let [t (i-key (-transformer s transformer context))] [k t])))
                                    (seq)
-                                   (apply array-map))
-                          map-transformer (i-key (-value-transformer transformer this context))
+                                   (apply array-map)
+                                   (#(set/rename-keys % compile-time-key-renames)))
                           apply-key-transformers (fn [m k v]
                                                    (let [k' (key-transformer k)]
                                                      (-> m
@@ -285,7 +288,11 @@
                         (and (not key-transformer) value-transformers map-transformer)
                         (fn [x]
                           (if (map? x)
-                            (reduce-kv apply-value-transformers (map-transformer x) value-transformers)
+                            (let [map-transformed (map-transformer x)
+                                  value-transformers (if-some [renames (::rename-keys (meta map-transformed))]
+                                                       (set/rename-keys value-transformers renames)
+                                                       value-transformers)]
+                              (reduce-kv apply-value-transformers map-transformed value-transformers))
                             x))
 
                         (and key-transformer (not value-transformers) map-transformer)
@@ -299,6 +306,9 @@
                         (fn [x]
                           (if (map? x)
                             (let [map-transformed (map-transformer x)
+                                  value-transformers (if-some [renames (::rename-keys (meta map-transformed))]
+                                                       (set/rename-keys value-transformers renames)
+                                                       value-transformers)
                                   values-transformed (reduce-kv apply-value-transformers map-transformed value-transformers)]
                               (reduce-kv apply-key-transformers values-transformed values-transformed))
                             x)))))]
--- a/test/malli/transform_test.cljc
+++ b/test/malli/transform_test.cljc
@@ -1,5 +1,6 @@
 (ns malli.transform-test
   (:require [clojure.test :refer [deftest testing is]]
+            [clojure.set :as set]
             [malli.core :as m]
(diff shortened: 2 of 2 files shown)


ikitommi:
I see the problem, but not happy with the meta-data. Would a normal pre-walk work here?

```clj
(m/decode
                   [:and
                    {:decode/before (constantly {:leave (partial
                                                          walk/postwalk
                                                          #(cond-> % (map? %) (set/rename-keys {:foo :oof :bar :rab})))})}
                    [:map
                     [:foo [:map
                            [:foo int?]
                            [:bar int?]]]
                     [:bar int?]]]
                   {:foo {:foo 2 :bar 1} :bar 3}
                   (transformer
                     {:name :before
                      :decoders {'int? (constantly inc)}}))
; => {:oof {:oof 3, :rab 2}, :rab 4}
```

ikitommi:
or from outside:

```clj
(->> (m/decode
       [:map
        [:foo [:map
               [:foo int?]
               [:bar int?]]]
        [:bar int?]]
       {:foo {:foo 2 :bar 1} :bar 3}
       (transformer
         {:name :before
          :decoders {'int? (constantly inc)}}))
     (walk/postwalk #(cond-> % (map? %) (set/rename-keys {:foo :oof :bar :rab}))))
; => {:oof {:oof 3, :rab 2}, :rab 4}
```

ikitommi:
... or some way to backtrack changes keys. hmm. might be too complex.

rschmukler:
Backtracking key changes is easy if we have a key mapping function (such as `map-of`) but not when the transformer can arbitrarily change things.

A normal post-walk does work if you can modify the schema entry's decoder, but not if you want to put the functionality into a transformer. As far as I can tell, there is no way to perform the change in the test (as a `:map` transformer) without this type of functionality.

ikitommi:
what about:

```clj
(->> (m/decode
       [:map
        [:foo [:map
               [:foo int?]
               [:bar int?]]]
        [:bar int?]]
       {:foo {:foo 2 :bar 1} :bar 3}
       (transformer
         {:name :before
          :decoders {'int? (constantly inc)
                     :map (constantly {:leave (partial
                                                walk/postwalk
                                                #(cond-> % (map? %) (set/rename-keys {:foo :oof :bar :rab})))})}})))
; => {:oof {:oof 3, :rab 2}, :rab 4}
```

ikitommi:
... actually, should be encode as decode should transform the values into domain:

```clj
(->> (m/encode
       [:map
        [:foo [:map
               [:foo int?]
               [:bar int?]]]
        [:bar int?]]
       {:foo {:foo 2 :bar 1} :bar 3}
       (transformer
         {:encoders {'int? (constantly inc)
                     :map (constantly {:leave (partial
                                                walk/postwalk
                                                #(cond-> % (map? %) (set/rename-keys {:foo :oof :bar :rab})))})}})))
```

rschmukler:
The above solution makes it so that the root-most map transformer is responsible for all renames. To make things a bit less abstract, let me explain my intended use case.

I'd like to encode namespaced maps as unnamespaced entries, using no prefix if the namespace is the most common namespace in the schema, and otherwise prefixing with the namespace. Eg:

```clj
(def schema
  [:map 
     [:person/name string?]
     [:person/age pos-int?]
     [:person/location
       [:map [:location/city string?]
             [:location/state string?
              :location/coordinates
              [:map [:coordinate/lon float?]
                    [:coordinate/lat float?]]
     [:oauth/id int?]
     [:oauth/username string?]])

{:person/name "Bob Jones"
  :person/age 42
  :person/location {:location/city "Brooklyn" :location/state "NY"
                    :location/coordinates {:coordinate/lon 1.51
                                           :coordinate/lat 1.21}}
  :oauth/id 5
  :oauth/username "bobbyj"}

;; should become

{:name "Bob Jones"
  :age 42
  :location {:city "Brooklyn" :state "NY" :coordinates {:lon 1.51 :lat 1.21}
  :oauth-id 5
  :oauth-username "bobbyj"}
```

With key renames this can look like:

```clj
(defn generate-encode-key-rename-map
  "Function to generate a map of key renames for a provided
  schema with the given options."
  [schema opts]
  (let [allowed-keys (:keys (m/-parse-keys (m/children schema) opts))
        root-ns      (or (:json/root-namespace opts)
                         (->> (frequencies (map namespace allowed-keys))
                              (sort-by val)
                              (last)
                              key))
        rename-map   (into {}
                           (map
                            (fn [k]
                              [k (let [ns   (namespace k)
                                       name (name k)]
                                   (if (= root-ns ns)
                                     name
                                     (str (str/kebab ns) "-" name)))]))
                           allowed-keys)]
    rename-map))


(defn encode-namespaced-keys
  "Returns an interceptor that will encode keys as strings as well as filtering
  the map to only allowed keys."
  [schema opts]
  (let [rename-map   (generate-encode-key-rename-map schema opts)
        allowed-keys (keys rename-map)]
    (with-meta
      (fn [x]
        (-> x
            (select-keys allowed-keys)
            (set/rename-keys rename-map)))
      {::m/rename-keys rename-map})))

(def my-transformer
  {:name :customs 
    {:encoders {:map encode-namespaced-keys}})
```

Doing something like the above becomes extremely difficult with the current implementation.

ikitommi:
I think #140 will help here. 

Related to the `(m/-parse-keys (m/children schema) opts)`, that's horrible - the same code exists in the json-schema & generator namespaces. Maybe we could have a new Protocol that the `:map` schema implements and which would have the map-releated helpers? e.g. public api instead of impl details...

rschmukler:
Fixed with proper interceptor order. Put it in the `:leave` of a `:map` transformer!

PR #135, opened 2019-12-05
fix: check type before transforming collections

rschmukler:
Fixes an error where collection transformers could either inadvertantly
coerce seqable input into collections (eg. strings) or explode when
trying to coerce truly invalid input (which is bad if you use a
collection transformer in an `:or` schema with a non collection transformer).

This is also the same issue that @roklenarcic described in the slack today.


rschmukler:
Let me know if you want to keep the exception handling as it was written. I believe that by checking the collection type before calling the mapping function, we will only raise valid errors.

As it was written previously, an exception in the mapping function would simply be swallowed which could be very confusing to a user.

ikitommi:
Thanks!

Issue #136, opened 2019-12-06
Interceptors, Transformers, composition


ikitommi:
Related to the Slack conversion, thinking aloud how things should work. Transformation engine should be **simple** but **powerful**. We are at pre-alpha, so we can still toss away things if they get too complex.

My 2 cents:

* transformer should not have only one `-transformer-name`, but instead, a chain of names. `strip-extra-keys-transformer` should have an unique name (e.g. `:malli.transform/strip-extra-keys`)
* with `mt/transformer`, one can create a chain of transformations, where both names and encoders & decoders are chained together and the chain of names is used for schema property based lookups. Any schema-defined will override the one defined in the transformer
* encoders and decoders are defined as a transformation function of type `schema opts => IntoInterceptor`. This allows any transformation function to access the schema in question at “compile time”, allowing fast “drop extran keys” etc.
* interceptor is to describe a transformation step.  There is a `IntoInterceptor` Protocol, which is extended for a function (maps to :enter), a map, Interceptor Record or a chain of IntoInterceptor . Malli could use sieppari later, as it already defines all of these.

In most cases, users don’t have to worry anything about the interceptor machineny, just use plain functions and it works. When you need more batteries on what happens when, you can either
1) start using interceptors (enter & leave)
2) add more steps into the transformation chain
3) both

### Simplest thing that works

```clj
(m/decode
  [string? {:decode/string '(constantly #(str "olipa_" %))}]
  "kerran" mt/string-transformer)
; => olipa_kerran
```

### With Interceptor

```clj
(m/decode
  [string? {:decode/string '(constantly {:enter #(str "olipa_" %)
                                         :leave #(str % "_avaruus")})}]
  "kerran" mt/string-transformer)
; => "olipa_kerran_avaruus"
```

### A custom transformaton chain

```clj
(m/decode
  [:and
   {:decode/before '(constantly inc)
    :decode/after  '(constantly (partial * 2))}
   int?]
  1
  (mt/transformer {:name :before} {:name :after}))
; => 4
```

### both

```clj
(m/decode
  [:and
   {:decode/before '(constantly {:enter inc
                                 :leave (partial * 2)})
    :decode/after  '(constantly {:enter (partial * 4)
                                 :leave inc})}
   int?]
  1
  (mt/transformer {:name :before} {:name :after}))
; => 18
```

### using a chain

```clj
(m/decode
  [:and
   {:decode/before '(constantly [inc inc #(* % 2) inc])}
   int?]
  1
[...]

PR #137, opened 2019-12-07
Transformer Chains

ikitommi:
Related to #136:

* polish naming: 
  * `method` (internal), defines what we are about to do. Valid values: `:decode`, `:encode`
  * `phase`, interceptor pipeline, either `:enter` or `:leave`
  * `name`, a transformer name

* simplified `malli.core/Transformer` Protocol:

```clj
(defprotocol Transformer
  (-transformer-chain [this] "returns transformer chain as a vector of maps with :name, :encoders, :decoders and :opts")
  (-value-transformer [this schema method] "returns an value transforming interceptor for the given schema and method"))
```

* `malli.transform/transformer` allows a chain of transformers to be defined
    * `:encoders` and `:decoders` are chained together, in order (instead or merging per schemas as before)
    * each named transformer allow schema-based transformations that can override the default `:encoders` and `:decoders`

```clj
(m/decode
  [int? {:decode/before '(constantly {:leave inc})
         :decode/after '(constantly (partial * 2))}]
  "10"
  (mt/transformer
    {:name :before}
    mt/string-transformer
    {:decoders {'int? (constantly inc)}} ;; anonymous
    {:name :after}))
; => 23

;; :enter
;;  :string   "10" => 10
;;  anonymous   10 => 11
;;  :after      11 => 22
;; :leave
;;  :before     22 => 23
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -20,7 +20,7 @@
   (-form [this] "returns original form of the schema"))
 
 (defprotocol Transformer
-  (-transformer-chain [this] "returns transformer chain as a vector of maps")
+  (-transformer-chain [this] "returns transformer chain as a vector of maps with :name, :encoders, :decoders and :opts")
   (-value-transformer [this schema method] "returns an value transforming interceptor for the given schema and method"))
 
 (defrecord SchemaError [path in schema value type message])
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -137,10 +137,10 @@
                   acc explainers))))
           (-transformer [this transformer method]
             (let [build-transformer
-                  (fn [stage]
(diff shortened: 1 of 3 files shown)


rschmukler:
This looks great! Repeating comment from slack for posterity.

I believe the `transformer-chain` method can be removed from the `Transformer` protocol and instead just be accumulated in the `mt/transformer` function. This keeps the `Transformer` protocol simpler, and moves the concern of chaining to the `mt/transformer` function which, to me, makes sense since it's really the only thing concerned with composing the chains into a single transformer.

ikitommi:
Actuall the `transformer-chain` is needed to pull out chains out of existing instances, so we can say:

```clj
(mt/transformer mt/string-transformer {:name :after})
```

PR #142, opened 2019-12-12
fix: double transformer invokation + :leave order

### LINKED ISSUE #140: Interceptors out of phase with maps (& sequential schemas)
The transformation order in `:leave` should be the reverse of the order of `:enter`. So, with `:map` schemas:

* `:enter`
  1) `:map`
  2) `:malli.core/map-key`
  3) childs
* `:leave`
  1) childs
  2) `:malli.core/map-key`
  3) `:map`

... this solves the case of "renaming map keys on encode as the last thing":

```clj
(m/encode
  [:map
   [:foo [:map
          [:foo int?]
          [:bar int?]]]
   [:bar int?]]
  {:foo {:foo 2 :bar 1} :bar 3}
  (transformer
    {:encoders {'int? (constantly inc)
                :map (constantly {:leave #(set/rename-keys % {:foo :oof :bar :rab})})}}))
; => {:oof {:oof 3, :rab 2}, :rab 4}
```

Also, with the current impl, the `build-transformer` re-creates all three transformers for both phases (:enter & :leave), which is wrong: if a transformer has some state (e.g. a generated uuid), it is not shared between phases. All creation of transformers should be moved outside of the `build-transformer` fn and just invoke the `:enter` or `:leave` of it within the fn.

rschmukler:
This commit fixes two issues:

1) It no longer invokes `transformer` related methods twice while
creating parent transformers. This fixes errors with stateful
transformers that depend on sharing data between enter and leave.

2) It fixes the order in which interceptors call their children for the
`map-of`, `sequence`, `tuple` and `map` interceptors - causing the child
transformers to be invoked first on the `leave`.

Closes #140, #132

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -136,18 +136,20 @@
                         :else acc'')))
                   acc explainers))))
           (-transformer [this transformer method]
-            (let [build-transformer
+            (let [value-transformer  (-value-transformer transformer this method)
+                  child-transformers (map #(-transformer % transformer method) child-schemas)
+                  build-transformer
                   (fn [phase]
-                    (let [st (phase (-value-transformer transformer this method))
+                    (let [st  (phase value-transformer)
                           ?st (or st identity)
-                          tvs (into [] (keep #(phase (-transformer % transformer method)) child-schemas))]
+                          tvs (into [] (keep phase) child-transformers)]
                       (cond
                         (not (seq tvs)) st
                         short-circuit (fn [x]
                                         (let [x (?st x)]
                                           (reduce-kv
-                                            (fn [_ _ t] (let [x' (t x)] (if-not (identical? x' x) (reduced x') x)))
-                                            x tvs)))
+                                          (fn [_ _ t] (let [x' (t x)] (if-not (identical? x' x) (reduced x') x)))
+                                          x tvs)))
                         :else (fn [x] (reduce-kv (fn [x' _ t] (t x')) (?st x) tvs)))))]
               {:enter (build-transformer :enter)
                :leave (build-transformer :leave)}))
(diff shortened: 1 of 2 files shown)


src/malli/core.cljc
```diff
@@ -136,19 +136,21 @@
...
+                  child-transformers (map #(-transformer % transformer method) child-schemas)
+                  build-transformer
                   (fn [phase]
-                    (let [st (phase (-value-transformer transformer this method))
+                    (let [st  (phase value-transformer)
                           ?st (or st identity)
-                          tvs (into [] (keep #(phase (-transformer % transformer method)) child-schemas))]
+                          tvs (into [] (keep phase) child-transformers)]
                       (cond
-                        (not (seq tvs)) st
```
ikitommi: please reformat the code, lot's of whitespace changes (cond, let-aligns).

src/malli/core.cljc
```diff
@@ -341,40 +327,37 @@
...
+                                              (fn [x] (t (keyword->string x))))
+                          value-transformer (phase value-transformers)
+                          kv-transform      (cond
+                                              (and key-transformer value-transformer)
+                                              #(assoc %1 (key-transformer %2) (value-transformer %3))
+                                              key-transformer
+                                              #(assoc %1 (key-transformer %2) %3)
+                                              value-transformer
+                                              #(assoc %1 %2 (value-transformer %3))
+                                              :else nil)
```
ikitommi: redundant :else

src/malli/core.cljc
```diff
@@ -341,40 +327,37 @@
...
-                                    (assoc acc k' v))) (empty x) x))
-                            x))
-                        value-transformer
-                        (fn [x]
-                          (if (map? x)
-                            (let [x (?tt x)]
-                              (reduce-kv (fn [acc k v] (assoc acc k (value-transformer v))) x x))
-                            x)))))]
+                            (transform x)
+                            nil)))))]
```
ikitommi: should be `x`?

rschmukler:
Formatting applied. I also re-enabled some disabled `map-of` tests that I found because the bug you caught wasn't enabled in the test suite.

I also ended up renaming `k'` to `new-k`. This is because some editors (emacs) attempt to auto-align quotes with the trailing expression (ie. if you "auto-format-region" on `(k' foo)` it becomes `(k 'foo)` which leads to all sorts of bug hunting.

ikitommi:
Thanks!!

PR #146, opened 2019-12-13
BREAKING: Interceptor compilation

### LINKED ISSUE #145: Interceptors & :compile
Instead of having a verbose common case for interceptors:

```clj
;; the 99% case
[int? {:decode/math '(constantly inc)}]

;; the 1% case
[int? {:math/multiplier 10
       :decode/math '(fn [schema _]
                       (let [multiplier (:math/multiplier (m/properties schema))]
                         (fn [x] (* x multiplier))))}]
```

we should:

```clj
;; the 99% case
[int? {:decode/math 'inc}]

;; the 1% case
[int? {:math/multiplier 10
       :decode/math '{:compile (fn [schema _]
                                 (let [multiplier (:math/multiplier (m/properties schema))]
                                   (fn [x] (* x multiplier))))}}]
```

, inspired by [reitit](https://cljdoc.org/d/metosin/reitit-http/0.3.7/doc/ring/compiling-middleware)

ikitommi:
### Breaking Change (for the better)

Fixes #145  - tranformations can be declared as `value => value` functions instead of `schema opts => value => value` making the 99% case simpler and just the 1% bit harder (but still possible).

Schema properties can be used to override default transformations:

```clj
(m/decode
  [string? {:decode/string 'str/upper-case}]
  "kerran" mt/string-transformer)
; => "KERRAN"
```

Decoders and encoders as interceptors (with `:enter` and `:leave` stages):

```clj
(m/decode
  [string? {:decode/string '{:enter str/upper-case}}]
  "kerran" mt/string-transformer)
; => "KERRAN"
```

```clj
(m/decode
  [string? {:decode/string '{:enter #(str "olipa_" %)
                             :leave #(str % "_avaruus")}}]
  "kerran" mt/string-transformer)
; => "olipa_kerran_avaruus"
```

To access Schema (and options) use `:compile`:

```clj
(m/decode
  [int? {:math/multiplier 10
         :decode/math '{:compile (fn [schema _]
                                  (let [multiplier (:math/multiplier (m/properties schema))]
                                    (fn [x] (* x multiplier))))}}]
  12
  (mt/transformer {:name :math}))
; => 120
```

Going crazy:

```clj
(m/decode
  [:map
   {:decode/math '{:enter #(update % :x inc)
                   :leave #(update % :x (partial * 2))}}
   [:x [int? {:decode/math '{:enter (partial + 2)
                             :leave (partial * 3)}}]]]
  {:x 1} 
  (mt/transformer {:name :math}))
; => {:x 42}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,59 @@ Transformers are composable:
 ;           :lonlat [61.4858322 23.7854658]}}
 ```
 
+Schema properties can be used to override default transformations:
+
+```clj
+(m/decode
+  [string? {:decode/string 'str/upper-case}]
+  "kerran" mt/string-transformer)
+; => "KERRAN"
+```
+
+Decoders and encoders as interceptors (with `:enter` and `:leave` stages):
+
+```clj
+(m/decode
+  [string? {:decode/string '{:enter str/upper-case}}]
+  "kerran" mt/string-transformer)
+; => "KERRAN"
+```
+
+```clj
+(m/decode
+  [string? {:decode/string '{:enter #(str "olipa_" %)
+                             :leave #(str % "_avaruus")}}]
+  "kerran" mt/string-transformer)
+; => "olipa_kerran_avaruus"
+```
+
+To access Schema (and options) use `:compile`:
+
+```clj
+(m/decode
+  [int? {:math/multiplier 10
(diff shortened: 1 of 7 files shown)


rschmukler:
LGTM :shipit:

PR #150, opened 2019-12-16
fix: type changing transformers

rschmukler:
This commit introduces code and testing for two fixes:

1) It removes the check on the type of the data before calling the
`this-transformer` for the `tuple`, `map-of`, `map` and
`sequence`-related transformers. This is necessary to allow for encoded
/ decoded data to be in shapes that differ in type from their root
schema. Eg. a map schema that is encoded as a tuple of values (common in
financial candle data).

2) Introduces the ability for a `-chain` to take a tuple with a
`predicate` and a `sub-chain`. This allows schemas to specify groups of
child-transformations that depend on a certain predicate. This fixes an
error where post `this-transformer` functions would change their type to
a shape that would cause invoking the children to crash.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -76,11 +76,31 @@
     (seq children) (into [name] children)
     :else name))
 
+(defn- -transform-if
+  [pred tf]
+  (when tf
+    (fn [x]
+      (if (pred x)
+        (tf x)
+        x))))
+
 (defn- -chain [phase chain]
   (let [f (case phase
             :enter identity
             :leave reverse)]
-    (some->> chain (keep identity) (seq) (f) (reverse) (apply comp))))
+    (some->> chain
+             (keep (fn [item]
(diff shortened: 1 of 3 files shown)


src/malli/core.cljc
```diff
@@ -76,11 +76,31 @@
...
             :enter identity
             :leave reverse)]
-    (some->> chain (keep identity) (seq) (f) (reverse) (apply comp))))
+    (some->> chain
+             (keep (fn [item]
+                     (if (vector? item)
+                       (let [pred (first item)
+                             child-chain (second item)]
+                         (assert (fn? pred) "First item in tuple must be a predicate fn")
+                         (assert (seqable? child-chain) "Second item in tuple must be a sequence")
```
rschmukler: These assertions are useful because if you call `-chain` with an invalid format, the error becomes very difficult to track down.

PR #151, opened 2019-12-17
Rebased + Merged PRs

rschmukler:
I ended up needing this for local development. Feel free to discard if you don't want it.

This is a rebased / updated PR that includes the changes in #149, #147, and #150 (with the nested `-chain` removed, instead just using `-guard` as requested).


ikitommi:
Thanks! Will merge this when possible.

PR #152, opened 2019-12-27
New utilities & move things around

ikitommi:
* `malli.util/equals` (no-doc)
* `malli.util/simplify-map-entry` (no-doc)
* `malli.util/merge`
* `malli.util/union`

* `malli.core/required-map-entry?` (no-doc)
* `malli.core/map-entries` (`:map` schemas implement new `malli.core/MapSchema` protocol)

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,30 +404,60 @@ Single sweep of defaults & string encoding:
 ; :c {:x "42"}}
 ```
 
-## Merging Schemas
-
-Schemas can be deep-merged with `m/merge`:
+## Programming with Schemas
 
 ```clj
-(m/merge
-  Address
+(require '[malli.util :as mu])
+```
+
+Merging Schemas (last value wins):
+
+```clj
+(mu/merge
   [:map
+   [:name string?]
    [:description string?]
+   [:address
+    [:map
+     [:street string?]
+     [:country [:enum "finland" "poland"]]]]]
+  [:map
+   [:description {:optional true} string?]
    [:address
     [:map
      [:country string?]]]])
 ;[:map
-; [:id string?]
-; [:tags [:set keyword?]]
-; [:address 
-;  [:map 
-;   [:street string?] 
-;   [:city string?] 
-;   [:zip int?] 
-;   [:lonlat [:tuple double? double?]] 
-;   [:country string?]]]
-; [:description string?]]
+; [:name string?]
+; [:description {:optional true} string?]
+; [:address [:map
+;            [:street string?]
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -404,31 +404,61 @@ Single sweep of defaults & string encoding:
...
+     [:street string?]
+     [:country [:enum "finland" "poland"]]]]]
+  [:map
+   [:description {:optional true} string?]
+   [:address
+    [:map
+     [:country string?]]]])
+;[:map
+; [:name string?]
+; [:description string?]
```
rschmukler: Should this be optional?

rschmukler (changes requested):
Edit: Looks like I misread the code - Looks good!

I do have a question as to whether `union` should cause optional to be transferred to the new schema. I would expect it would since any value of `A` or `B`, including the map missing the field should be valid. The issue is then that incorrect values of `B` could be accepted. I'm wondering which behavior is more surprising. Maybe the only "safe" answer is using `[:or schema-1 schema-2]`?

src/malli/core.cljc
```diff
@@ -185,36 +175,33 @@
...
-     :forms (mapv (fn [[k p v]]
-                    (let [v' (-form v)]
-                      (if p [k p v'] [k v']))) entries)}))
+(defn- -parse-map-entries [children opts]
+  (->> children (keep identity) (mapv #(-expand-key % opts identity))))
+
+(defn ^:no-doc map-entry-forms [entries]
+  (mapv (fn [[k p v]] (let [v' (-form v)] (if p [k p v'] [k v']))) entries))
+
+(defn ^:no-doc required-map-entry [[_ ?p]]
```
rschmukler: Consider renaming `required-map-entry?` so it's obvious that it's a predicate. Also perhaps this should not be `no-doc`?

README.md
ikitommi: I think it should, fixed.

src/malli/core.cljc
ikitommi: renamed but left the `:no-doc` for now.

ikitommi:
moved `malli.core/equals` into a no-doc function in `malli.util` as it's only used in there. Might need to rethink Schema equality (equals & hash-code?), also less public api for now.

ikitommi:
btw, `util` is a bad name for namespace, but then again `mu` was not used as an ns alias...

``` clj
(require '[malli.util :as mu])
```

ikitommi:
thanks for the comments @rschmukler !

PR #156, opened 2019-12-31
Add support for closed maps

ikitommi:
Maps are open by default:

```clj
(m/validate
  [:map [:x int?]]
  {:x 1, :extra "key"})
; => true
```

Maps can be closed with `:closed` property:

```clj
(m/validate
  [:map {:closed true} [:x int?]]
  {:x 1, :extra "key"})
; => false

(m/explain
  [:map {:closed true} [:x int?]]
  {:x 1, :extra "key"})
;{:schema [:map {:closed true} [:x int?]],
; :value {:x 1, :extra "key"},
; :errors [#Error{:path []
;                 :in [:extra]
;                 :schema [:map {:closed true} [:x int?]]
;                 :type :malli.core/extra-key}]}
```

also:

```clj
(-> [:map {:closed true} [:x int?]]
    (m/explain {:x 1, :extra "key"})
    (me/humanize))
; => {:extra ["disallowed key"]}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,24 @@ Schemas can have properties:
 ;     :json-schema/example 20}   
 ```
 
+Maps are open by default:
+
+```clj
+(m/validate
+  [:map [:x int?]]
+  {:x 1, :extra "key"})
+; => true
+```
+
+Maps can be closed with `:closed` property:
+
+```clj
+(m/validate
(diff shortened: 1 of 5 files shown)


rschmukler (approved):
Looks awesome!

PR #170, opened 2020-01-26
prefer `npm ci` over `npm install`

dazld:
the npm package-lock is only respected when you run `npm ci`.

https://docs.npmjs.com/cli/ci


ikitommi:
Thanks!

PR #171, opened 2020-02-05
Pass options map to closes-schema and open-schema

olimsaidov:
Calling `malli.util/closed-schema` and `malli.util/open-schema` fails when using custom registry


ikitommi:
Thanks!

PR #172, opened 2020-02-10
Programming with schemas

### LINKED ISSUE #63: Programming with Schemas, part2
[schema-tools](https://github.com/metosin/schema-tools/tree/master/src/schema_tools) has nice utilities working with schemas. We should have some of those too, maybe in a separate ns? See also:

* #45 (fixed with #62)
* [spec2 select](https://github.com/clojure/spec-alpha2/wiki/Schema-and-select)
* [EQL](https://github.com/edn-query-language/eql)

```clj
(-> [:map
     [:a int?]
     [:b string?]
     [:c [:map
          [:x int?]
          [:y keyword?]]]]
    (m/select-keys [:a :c])
    (m/update-in [:c] dissoc :y))
;[:map
; [:a int?]
; [:c [:map
;      [:x int?]]]]
```

ikitommi:
Add new helpers into `malli.util`:
* [x] `select-keys`
* [x] `dissoc`
* [x] `get`
* [x] `assoc`
* [x] `update`
* [x] `get-in`
* [x] `assoc-in`
* [x] `update-in`
* [x] `optional-keys`
* [x] `required-keys`

resolved #63 

TODO:
* [x] `LensSchema` to composite schemas

Screenshot 2020-02-11 at 17 19 58


ikitommi:
code complete, need docs.

ikitommi:
docs done

rschmukler:
This all looks incredibly useful! The only thing I'm wondering is whether `get` and `get-in` should facilitate default values like `clojure.core/get` and `clojure.core/get-in`. I'm not sure that it's worth adding a 4 arity variant to facilitate both options and defaults, but it might be. If that's the case, it might be worth even giving the default to the 3 arity variant for true parity with `clojure.core` functions, and then allowing specifying options with `(get schema :key nil {:registry my-registry})`

rschmukler (approved):

ikitommi:
created a separate issue of the 3-arity get & get-in. good points, should be there.

rschmukler:
:+1:

PR #175, opened 2020-02-16
Allow custom bindings to :fn schema

jeans11:


ikitommi:
Instead of:

```clj
[:fn
 {:bindings {'dumb dumb}}
 '(fn [value] (some? (dumb value)))]
```

Coudn't you just:

```clj
[:fn (fn [value] (some? (dumb value)))]
```

you lose the serialization, but there is no guarantees for `dumb` being serializable anyway, right?

jeans11:
I didn't know that `:fn` support plain clojure function (without serialize it)! In this case, ok for lose the serialize part.

PR #176, opened 2020-02-16
Add :instance spec

piotr-yuxuan:
Thanks for this library which looks promising. Would you consider adding `instance` as a possible  spec definition? If so, I will further this, write tests and add transformers.


piotr-yuxuan:
Follow-up: a possible hindrance is the classes known from within `sci`. For example I believe the rather standard `java.time.*` would be unknown. How could it be overcome?

ikitommi:
the underlying code has changed so much, this would need tests, errors messages etc - so, closing thing.

but thanks for the PR!

Issue #180, opened 2020-02-17
Sequence/regex schemas


ikitommi:
From slack:

like clojure.spec.alpha/conform, there could be something for that? Explain could return either failures or real values if no errors exist.

```clj
 (->
   [:cat
    [:x int?]
    [:y int?]
    [:rest [:* string?]]]
   [1 2 "kikka" "kukka"]
  (m/explain)
  (mu/humanize))
 ; {:x 1
 ;  :y 2
 ;  :rest ("kikka "kukka")}
```

and/or, an anonymous variant:

```clj
 (m/valid?
   [:cat int? int? [:* string?]]
   [1 2 "kikka" "kukka"])
 ; => true
```

using for example https://github.com/cgrand/seqexp under the hoods

green-coder:
When designing it, the scenario of implementing the equivalent of `clojure.alpha.spec/unform` might also be considered.

For info, I would use both conform and unform if mali has it.

ikitommi:
Related issues:
* https://github.com/cgrand/seqexp/issues/11
* https://github.com/cgrand/seqexp/issues/9

ikitommi:
Initial internal impl done, just 123loc. need to add tests and figure out how to expose the parsing api in general. 

<img width="521" alt="Screenshot 2020-09-10 at 20 41 31" src="https://user-images.githubusercontent.com/567532/92779756-02ca7000-f3ab-11ea-8e48-dbe29458ee67.png">

ikitommi:
Master!

PR #187, opened 2020-03-08
Sequence/regex schema

green-coder:
This is a proposal for the format of data the users will provide to Malli. (Related to #180)

Notes:
- I think that users of `:cat` will use it more often *with* names on children rather that *without*.
- I chose to make the format anonymous/non-anonymous explicit, to avoid conflicts between names of children and other operators like `:*` or `:alt`.

### DIFF
--- a/test/malli/core_test.cljc
+++ b/test/malli/core_test.cljc
@@ -586,32 +586,32 @@
                                         [false [:sequential int?] #{1 2 3}]
                                         [false [:sequential int?] nil]]
 
-                          "cat" [; same as [:cat {:into nil} ...]
+                          "cat" [; [:seq-of ...] wrapper is implicitly used around [:cat ...]
                                  [true [:cat [:x int?]] '(1)]
                                  [true [:cat [:x int?]] [1]]
 
-                                 ; aka :cat*
-                                 [true [:cat {:into :any} [:x int?]] '(1)]
-                                 [true [:cat {:into :any} [:x int?]] [1]]
+                                 [true [:seq-of [:cat [:x int?]]] '(1)]
+                                 [true [:seq-of [:cat [:x int?]]] [1]]
 
-                                 ; aka :catl
-                                 [true [:cat {:into :list} [:x int?]] '(1)]
-                                 [false [:cat {:into :list} [:x int?]] [1]]
+                                 [true [:list-of [:cat [:x int?]]] '(1)]
+                                 [false [:list-of [:cat [:x int?]]] [1]]
 
-                                 ; aka :catv
-                                 [false [:cat {:into :vector} [:x int?]] '(1)]
-                                 [true [:cat {:into :vector} [:x int?]] [1]]
+                                 [false [:vector-of [:cat [:x int?]]] '(1)]
+                                 [true [:vector-of [:cat [:x int?]]] [1]]
 
                                  ; concatenated sequences
                                  [true [:cat [:a [:cat [:ax int?] [:ay int?]]]
                                              [:b [:cat [:bx int?] [:by int?]]]] [1 2 3 4]]
 
                                  ; sequence hierarchy
-                                 [true [:cat [:a [:cat {:into :any} [:ax int?] [:ay int?]]]
-                                             [:b [:cat {:into :any} [:bx int?] [:by int?]]]] ['(1 2) [3 4]]]
+                                 ; :seq-of is needed explicitly when used as element of a :cat sequence
+                                 [true [:cat [:a [:seq-of [:cat [:ax int?] [:ay int?]]]]
+                                             [:b [:seq-of [:cat [:bx int?] [:by int?]]]] ['(1 2) [3 4]]]]
+                                 [true [:cat [:a [:list-of [:cat [:ax int?] [:ay int?]]]]
+                                             [:b [:vector-of [:cat [:bx int?] [:by int?]]]] ['(1 2) [3 4]]]]
 
-                                 ; aka :acat
-                                 [true [:cat {:anonymous true} int? int? [:* string?]] [1 2 "a" "b"]]
+                                 ; nameless variant, aka :cat-
+                                 [true [:cat {:named false} int? int? [:* string?]] [1 2 "a" "b"]]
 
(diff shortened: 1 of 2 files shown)


test/malli/core_test.cljc
```diff
@@ -586,6 +586,74 @@
                                         [false [:sequential int?] #{1 2 3}]
                                         [false [:sequential int?] nil]]
 
+                          "cat" [; same as [:cat {:into nil} ...]
+                                 [true [:cat [:x int?]] '(1)]
+                                 [true [:cat [:x int?]] [1]]
+
+                                 ; aka :cat*
+                                 [true [:cat {:into :any} [:x int?]] '(1)]
```
ikitommi: does `:into` mean just type validation? If so, the name is bit misleading.

test/malli/core_test.cljc
```diff
@@ -586,6 +586,74 @@
...
+                                 ; concatenated sequences
+                                 [true [:cat [:a [:cat [:ax int?] [:ay int?]]]
+                                             [:b [:cat [:bx int?] [:by int?]]]] [1 2 3 4]]
+
+                                 ; sequence hierarchy
+                                 [true [:cat [:a [:cat {:into :any} [:ax int?] [:ay int?]]]
+                                             [:b [:cat {:into :any} [:bx int?] [:by int?]]]] ['(1 2) [3 4]]]
+
+                                 ; aka :acat
+                                 [true [:cat {:anonymous true} int? int? [:* string?]] [1 2 "a" "b"]]
```
ikitommi: currently, both `:or` and `:and` are anonymous by default, should `:cat` be too? Once there is some real use for the named branches, I though of adding variants supporting with names. My initial hunch was to use different schema names, something like:

```clj
;; anonymous
[:or int? string?]

;; named
[:orn [:i int?] [:s str?]]

;; .. or via props
[:or {:named true} [:i int?] [:s str?]]
```

If there are two different things that contribute to schema names, it starts to be awkward:

```clj
;; anonymous
[:cat ...]

;; named branches
[:catn ...]

;; named + vector
[:catnv ...]

;; anonymous + vector
[:catv ...]
```

my 2 cents this morning: the "named branches" would go to name (`:cat` & `:catn`) and the type could be a property. What do you think?

ikitommi:
> I think that users of :cat will use it more often with names on children rather that without.

I think so too, but would be good to be coherent with `:or`. Well, pre-alpha, so we can change things.

test/malli/core_test.cljc
green-coder: It is meant for validation and generation. I took this wording from an option of https://clojure.org/guides/spec#_collections and would not mind changing it.

test/malli/core_test.cljc
green-coder: The names on `:and`, `:or`, `:cat` and `:alt` do not provide the same kind of information to the user:
- Names on `:and` and `:or` help reporting on which predicates invalidate (`:and`) or validate (`:or`) the data. They are used for indicating **properties** of the data. I guess that the user does not need this information most of the time.
- Names on `:cat` and `:alt` are providing the information on the **identity** of the data. Users would need it most of the time.

To me, it does not look strange for Malli to have by default no names on `:and` and `:or` and have some on `:cat` and `:alt`. Clojure Spec also have a mix of the 2, `s/and` has no name while `s/or`, `s/cat` and `s/alt` have them.

`{:anonymous true/false}` and `{:named: false/true}` encode the same information, I prefer `{:named false/true}` as it is shorter to read and write.

green-coder:
> Tests look great! As the destructuring is a valid application for sequence/regex schemas, ideas how to implement the s/conform with malli?
>
> My initial thought was to make m/explain explain also successful values, so one would get both the values and the branching information from it's result (if not errors).

I see 2 different usages there:
1. The human who occasionally wants to get information of what went wrong in a validation, possibly via a REPL.
2. The program who would do serial data transformation for further processing and that would not be interested to know why some of the transformation failed.

In the future, the API for `1.` would likely evolve toward better explanations, even if it costs more CPU, while the API for `2.` would likely evolve toward something that run faster. For this reason, I would recommend having different functions for the 2 usages. In other words, I think that explain, conform and unform should be separated in the API. They could however use common functions in their implementation.

At the moment, I don't know where would be the best place for the implementation of conform. Maybe as new functions in the Schema protocol, or as a Transformer, or as something else.

green-coder:
I am going to implement a naive implementation first: Because its implementation will be simple it will be easier to adapt to support recursive models later on.

test/malli/core_test.cljc
green-coder: @ikitommi Do you have a suggestion for an alternative to `:into`?

test/malli/core_test.cljc
```diff
@@ -586,32 +586,32 @@
...
-                                 [true [:cat {:into :list} [:x int?]] '(1)]
-                                 [false [:cat {:into :list} [:x int?]] [1]]
+                                 [true [:list-of [:cat [:x int?]]] '(1)]
+                                 [false [:list-of [:cat [:x int?]]] [1]]
 
-                                 ; aka :catv
-                                 [false [:cat {:into :vector} [:x int?]] '(1)]
-                                 [true [:cat {:into :vector} [:x int?]] [1]]
+                                 [false [:vector-of [:cat [:x int?]]] '(1)]
+                                 [true [:vector-of [:cat [:x int?]]] [1]]
```
ikitommi:
```clj
[:vec-of [:cat [:x int?]]
```
can be done with:
```clj
[:and vector? [:cat [:x int?]]
```

green-coder: That's fixed in the new commits.

green-coder:
Things done in this PR:
- Some tests for the validators of the sequence schemas.
- Validators for the sequence schemas are implemented and working.

Things to be done in this PR:
- Finish the implementation for the transformers,
- review the LensSchema implementation which may not be appropriate (copied from -map-schema),
- review the implementation of the visitors (copied from -map-schema),
- add generators for the sequence schemas.

ikitommi:
thanks once more, the clean slate impl is around the corner.

green-coder:
You're welcome, I am happy it helped.

PR #193, opened 2020-04-05
Add validation for -parse-map-entries

kwrooijen:
## Problem
I have the following (invalid) schema:

``` clojure
[:map 
 [:human/name]]
```
`:human/name` is missing the `string?` schema in this case. Resulting in the following error:

Screenshot 2020-04-05 at 16 47 15

This error is technically speaking correct, the schema of `:human/name` is in fact missing. It is also very misleading because the end-user might assume the _entire_ schema is missing.

## Solution

Validate the children before parsing them. Thow a new error if one or more children are invalid.

Screenshot 2020-04-05 at 16 47 53

## Notes

* The validation function is a bit premature, maybe you have some suggestions to improve this? We could destructure the vector in the arguments, and add a `:pre` condition? This would mean it would crash when it's not a vector though.
```clojure
(defn- -valid-child? [[k p s]]
  (and (some? k)
       (and (some? s)
            (or (map? p)
                (nil? p)))))
```
* Possibly improve the ex-info map? There isn't much data at this point to return though.
* A better name could be `-valid-map-entry?` ?

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -192,7 +192,15 @@
   (let [[p v] (if (or (nil? ?p) (map? ?p)) [?p ?v] [nil ?p])]
     [k p (f (schema v options))]))
 
+(defn- -valid-child? [child]
+  (or (== 2 (count child))
+      (and (== 3 (count child))
+           (or (nil? (second child))
+               (map? (second child))))))
+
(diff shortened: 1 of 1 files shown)


ikitommi:
I think this is good, same as with other syntax validators.

ikitommi:
Thanks!

Issue #14, opened 2019-07-13
Support qualified keys in maps


ikitommi:
Spec showed the way to reuse map key specs. This is kinda nice and would be easy to support: one could use explicitly registered leaf specs as maps key-value tuples, so:

```clj
(m/schema
  [:map
   [:id int?]
   [:name string?]])
```

could be written as:

```clj
(def registry
  (merge m/default-registry {::id int?}))

(m/schema
  [:map
   ::id
   [:name string?]]
  {:registry registry})
```

ikitommi:
closed, via #194, merged via #239 .

```clj
(m/validate
  [:map {:registry {::id int?
                    ::country string?}}
   ::id
   [:name string?]
   [::country {:optional true}]]
  {::id 1
   :name "kikka"})
; => true
```

ikitommi:
<img width="623" alt="Screenshot 2020-08-06 at 17 38 49" src="https://user-images.githubusercontent.com/567532/89546483-55919480-d80d-11ea-80f7-70b843e94051.png">

Issue #86, opened 2019-10-10
Feature request or bug: adding properties directly to map keys.


johesoman:
Hi Tommi!

I was surprised that it is not possible to add properties to keys directly, but I don't know if it is intended behavior or not. I found this when I was playing around with error messages, so the example focuses on that. As you can see the first expression evaluates with the error message, while the second one does not.
```clj
(-> [:map [:leet [:re {:error/message "should be '1337'"} #"^1337$"]]]
    (m/explain {:leet "1338"})
    (me/humanize {:wrap :message})) ;; => {:leet "should be '1337'"}

(-> [:map [:leet {:error/message "should be: '1337'"} #"^1337$"]]
    (m/explain {:leet "1338"})
    (me/humanize {:wrap :message})) ;; => {:leet "unknown error"}
```
If this is intended behavior, maybe the the fact that the second example does not work could be included in the docs.

ikitommi:
Excellent observation, I was sure there was an issue about it, but seems there wasn't. 

My first thought was that the key properties could be merged into child (here: value) properties, but it would raise an ambiguity issue: `:optional` should not be merged as it would mask the potentially relevant `:optional` in the value schema itself.

Another option to resolve this would be to introduce parent/child relationship between schema instances. This way, we could look parent properties at runtime when needed. Similar example with `:and`, where the error should be looked from parent if not present at child:

```clj
[:and {:error/message "fail"} int? [:> 18]]
```

for maps, we could create `key-schema` behind the scenes that is marked as parent for the value.

what do you think?

ikitommi:
related #80 (the parent/child)

johesoman:
I think that creating a `key-schema` behind the scenes and marking it as a parent for the value seems like a good idea!

ikitommi:
Should be resolved for first alpha.

rschmukler:
@ikitommi are you still thinking the parent / child relationship is the way to go on this? Would this mean that transformers would end up being able to reference them via `m/-parent` or something similar? And then getting properties becomes `(m/properties (m/-parent schema))`?

It may be worth stating that this could end up leading to inconsistent behavior for transformers, specifically if some of them end up checking parent properties, while others don't. I imagine it will be easy enough to get them to be consistent in `malli.core` but I do think it may lead to unintentional bugs in libraries / applications defining their own transformers. It also may be surprising for users that this is malli's behavior in core, but when consuming a library that didn't implement parent checking, the behavior is different.

[...]

PR #194, opened 2020-04-05
Add :map qualified keyword child syntax

kwrooijen:
Fixes: https://github.com/metosin/malli/issues/14


kwrooijen:
Maybe I could clean this up by creating a few predicate functions?
`-map-entry-qualified-keyword?`
`-map-entry-qualified-keyword-with-options?`
`-map-entry-regular?`
`-map-entry-regular-with-options?`

This might help the readability of the code. Right now it's just a bunch of predicates everywhere. We could also use these functions in both `-expand-key` and  `-valid-child?`.

kwrooijen:
Applied my comment in the latest commit. It's a bit more code but I hope you can agree it's more organized as well.

kwrooijen:
A thought: Would it be interesting to allow property inheritance?
Current:

``` clojure
;; Add to registry
(m/def ::id [int? {:description "Unique identifier"}])

[:map
 ::id]

;; Results in >>>

[:map 
 [::id nil [int? {:description "Unique identifier"}]]]
```

I can image that inheritance could be interesting in some cases.

``` clojure
;; Add to registry
(m/def ::id [int? 
  ^:inherit
  {:description "Unique identifier"}])

[:map
 ::id]

;; Results in >>>

[:map 
 [::id {:description "Unique identifier"} [int? {:description "Unique identifier"}]]]
```

Just an idea. But if you're interested, maybe we could create an issue for this?

kwrooijen:
On the same topic as my last comment; I also noticed that this doesn't work:

```clojure
(def s1 [:map [:user/name {:default "Anonymous"} string?]])
(m/decode s1 {} mt/default-value-transformer)
;; >> {}
```

But this does:

```clojure
(def s2 [:map [:user/name [string? {:default "Anonymous"}]]])
(m/decode s2 {} mt/default-value-transformer)
;; >> {:user/name "Anonymous"}
```

Maybe you could also pass properties the other way around? Something like...

```clojure
[:map 
 [:user/name
  ^:provide
  {:default "Anonymous"}]]
```
That way you could add a default to the value from the map's perspective. You would be able to share types between maps, but they would have their own usage for them.

ikitommi:
Will get back to this after #86 is resolved. Sorry for taking so long.

ikitommi:
pulled the code into current (changed) master, but broke it. will fix. But nice using a local registry:

```clj
(-> [:schema {:registry {:user/id int?
                         :user/name string?
                         :user/gender [:enum "male" "female" "other"]}}
     [:map {:closed true}
      ;; naked references to registry
      :user/id
      :user/name
      ;; can be optional
      [:user/gender {:optional true}]
      ;; local
      [:skills [:set [:enum "clj" "cljs"]]]
      [:user/skills [:set [:enum "clj" "cljs"]]]]]
    (m/explain
      {:user/id 1
       :user/genderz ""})
    (me/with-spell-checking)
    (me/humanize))
;{:user/name ["missing required key"],
; :skills ["missing required key"],
; :user/skills ["missing required key"],
; :user/genderz ["should be spelled :user/gender"]}
```

ikitommi:
merged in master via #239. Added some parsing tests and option to create `:map` without support of this. Thanks!

Issue #174, opened 2020-02-11
mu/get and mu/get-in should support setting the default value


ikitommi:
from #172 :

> The only thing I'm wondering is whether get and get-in should facilitate default values like clojure.core/get and clojure.core/get-in. I'm not sure that it's worth adding a 4 arity variant to facilitate both options and defaults, but it might be. If that's the case, it might be worth even giving the default to the 3 arity variant for true parity with clojure.core functions, and then allowing specifying options with (get schema :key nil {:registry my-registry})

PR #195, opened 2020-04-05
Add default value argument for mu/get and mu/get-in

kwrooijen:
Fixes: https://github.com/metosin/malli/issues/174

Note that since `options` is the last argument in all functions, this is technically a breaking change. (Which should be fine, since Malli is still pre-alpha)


ikitommi:
thanks!

PR #197, opened 2020-04-08
Allow specifying gen/gen and friends on map entries

plexus:
Map entry properties are a bit different from properties on regular schemas,
which means the :gen/gen, :gen/elements and :gen/fmap don't work there the way
you might expect. This adds support for these.


ikitommi:
Thanks for the PR. This got finally resolved via redesigning how the map-entries are handled, see #212

Issue #37, opened 2019-08-15
Derive schemas out of value, e.g. schema-provider


ikitommi:
like https://github.com/stathissideris/spec-provider, but for malli. This is needed already in figuring out `:enum` schema types in #30 (currently `nil` transformer`)

ikitommi:
example with json-schema: https://jsonschema.net/

ikitommi:
done

PR #200, opened 2020-05-03
#37 decode int? on cljs different from clj

geraldodev:
Javascript  parseInt is lenient with "+1-2" "1abc" resulting integer one.
Long/parseLong recognizes the sign at the start of the string.
So I've made a \D expression ignoring the first character and return ##Nan if it's found.
Just to be more like clojure. I think we can do this because we are parsing base 10 numbers.
The if string? already guards for string so its safe to call subs and re-find.

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -82,7 +82,9 @@
   (if (string? x)
     (try
       #?(:clj  (Long/parseLong x)
-         :cljs (let [x' (js/parseInt x 10)]
+         :cljs (let [x' (if (re-find #"\D" (subs x 1))
+                          ##NaN
+                          (js/parseInt x 10))]
                  (if (js/isNaN x') x x')))
       (catch #?(:clj Exception, :cljs js/Error) _ x))
     x))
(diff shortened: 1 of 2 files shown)


src/malli/transform.cljc
```diff
@@ -82,7 +82,9 @@
   (if (string? x)
     (try
       #?(:clj  (Long/parseLong x)
-         :cljs (let [x' (js/parseInt x 10)]
+         :cljs (let [x' (if (re-find #"\D" (subs x 1))
```
ikitommi, line 85: I think it would be better to use `js/Number` here, and test the result with `js/Number.isInteger`.

ikitommi: Would guess the performance is better but also the number logic stays elsewhere. What do you think.

geraldodev: After making the pull. I've tried isFinite to get rid of the re-find. But I've abandoned because it let numbers like '2e2' pass. And it's not on par with clojure parseLong. Number/isInteger suffers from the same problem. Number('2e2') = 200 .

ikitommi:
thanks!!

PR #205, opened 2020-06-08
welcome :string schema

ikitommi:
```clj
(m/validate :string "")
; => true

(m/validate [:string {:min 1}] "")
; => false

(json-schema/transform [:string {:min 1, :max 4}])
; => {:type "string", :minLength 1, :maxLenght 4}

(mg/sample [:string {:min 1, :max 4}])
; => ("RMmR" "5" "oL" "G7" "ENo" "cAh" "iOb" "jG" "l8" "kuo")

(-> [:map
     [:a :string]
     [:b [:string {:min 1}]]
     [:c [:string {:max 4}]]
     [:d [:string {:min 1, :max 4}]]]
    (m/explain
      {:a 123
       :b ""
       :c "invalid"
       :d ""})
    (me/humanize))
;{:a ["should be string"],
; :b ["should be at least 1 characters"],
; :c ["should be at most 4 characters"],
; :d ["should be between 1 and 4 characters"]}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,24 @@ Maps keys are not limited to keywords:
 ; => true
 ```
 
+## String schemas
+
+Using a predicate:
+
+```clj
+(m/validate string? "kikka")
+```
+
+Using `:string` Schema:
+
+```clj
+(m/validate :string "kikka")
+; => true
+
+(m/validate [:string {:min 1, :max 4}] "")
+; => false
+```
+
 ## Function schemas
 
 `:fn` allows any predicat function to be used:  
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -727,7 +727,7 @@
           (-explainer [this path]
             (fn explain [x in acc]
               (if (or (not (string? x)) (and count-validator (not (count-validator x))))
-                (conj acc (error path in this x ::string)) acc)))
+                (conj acc (error path in this x)) acc)))
           (-transformer [this transformer method options]
             (-value-transformer transformer this method options))
           (-accept [this visitor in options]
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -726,9 +726,8 @@
           (-validator [_] validator)
           (-explainer [this path]
             (fn explain [x in acc]
-              (cond
-                (not (string? x)) (conj acc (error path in this x ::invalid-type))
-                (not (and count-validator (count-validator x))) (conj acc (error path in this x ::limits)))))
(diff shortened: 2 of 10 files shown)


pithyless:
Any chance for adding an option along the lines of `blank?` or `present?`? Or is that too specific to add to malli-core? It's the first thing I have to add to all my string specs, since you almost never want to allow user input of just blank characters.

ikitommi:
`[:string {:min 1}]` is effectively a non-blank string.

pithyless:
I specifically want to avoid situations where user inputs `" "` to get around `{:min 1}`. My most common clojure.core spec in any project is `(s/and string? (complement str/blank?))`

pithyless:
I use a version of `:string` that applies `str/trim` before checking `:min` and `:max`, but not sure if that's something that is appropriate to `malli.core`.

rschmukler:
Just chiming in that a `:trim true` could be added, along with the `:min` and `:max` options. Potentially also a `:trim-l`, `:trim-r`. The only downside with this is there's some nuance of where it ends, one could argue that padding could also be supported with this. Eg. `:pad-l "0"`.

ikitommi:
Thanks for explaining. There are two things here:

1) validation of trimmed strings
2) trimming/padding strings

the latter can be done using a (new) transformer, the first would be currently:

```clj
[:and :string [:fn (complement str/blank?)]]
```

agree it's not pretty.

ikitommi:
JSON Schema has `:format` for strings. Would that be ok here, with value `:non-empty`?

https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats

ikitommi:
example trimmer:

```clj
(require '[malli.transform :as mt])
(require '[malli.core :as m])

(defn str-transformer []
  (mt/transformer
    {:decoders {:string {:compile (fn [schema _]
                                    (let [{:string/keys [trim]} (m/properties schema)]
                                      (when trim #(cond-> % (string? %) str/trim))))}}}))

(m/decode [:string {:min 1}] "    " (str-transformer))
; => "    "

(m/decode [:string {:string/trim true, :min 1}] "    " (str-transformer))
; => ""

(m/decoder :string (mt/str-transformer))
; => #object[clojure.core$identity] (no-op)
```

ikitommi:
Happy to take of a tested and lean string-trimmer PR with all the trims, pads etc.

ikitommi:
@pithyless could you write an separate issue of the trim-validated thing? Will merge this now.

miikka:
Post-merge review: the user guide -style example in the README is good, but a reference style documentation that lists the possible options would be also needed.

Issue #20, opened 2019-07-13
Support recursion in Malli Schema definitions


ikitommi:
Registry should support recursion, e.g. 

```clj
{::age int?
 :child-age [:and ::age [:> 18]] ;; should be expanded
 :child [:map 
         [:name string?] 
         [:age :child-age] 
         [:friends [:set :child]]]} ;; recursion
```

RafaeLeal:
Hello!
I have a different proposal, to be more flexible. So we don't require to be on registry to support recursion.
https://github.com/metosin/malli/pull/117

ikitommi:
merged in master

PR #209, opened 2020-06-22
Recursive Schemas Impl

ikitommi:
Displaces impl from #117, which will remain a place to discuss & history. FIxs #20 .

* new `m/Schema` method `-children` to get child schemas instances too
* `:multi` implements `m/MapSchma`
* new `m/RefSchma` protocol with the following impls:
  * `:ref` - a lazy reference, enabling recursion
  * `:schema`, like `m/schema` (entity) but in data, eager 
  * `malli.core/schema`, eager schema (value) ref, used internally. if no properties and child is a qualified keyword, derived form from the child
* support for local `:registry` in any element
* validators, explainers etc are moved from schema creation to actual `m/Schema` methods, to support laziness
* `m/-distance` and `m/-lookup` helpers
* `m/from-map-syntax`  works with non-map child's (enums, refs)
* updated README (local registry, recursive schemas)

# TODO

* [x] `:schema` for non-lazy references (retains original information)
* [x] local recursion using `:id`
* [x] local mutual recursion using `:registy`
* [x] new element to ensure local registries as data
* [x] ambiguity rules
* [x] ensure it's done right

# Current syntax:

```clj
(testing "recursion limiting"
  (are [schema]
    (is (every? (partial m/validate schema) (mg/sample schema {:size 1000})))

    ;; maybe
    [:schema {:registry {::rec [:maybe [:ref ::rec]]}}
     ::rec]

    ;; maps
    [:schema {:registry {::rec [:map [:rec {:optional true} [:ref ::rec]]]}}
     ::rec]

    ;; or
    [:schema {:registry {::tuple [:tuple boolean? [:ref ::or]]
                         ::or [:or int? ::tuple]}}
     ::or]

    ;; multi
    [:schema {:registry {::multi
                         [:multi {:dispatch :type}
                          [:int [:map [:type [:= :int]] [:int int?]]]
                          [:multi [:map [:type [:= :multi]] [:multi [:ref ::multi]]]]]}}
     [:ref ::multi]]))
```

# Initial Syntax:

(support for local recursion removed)

malli-rec

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -93,7 +93,7 @@
 (defmethod -schema-generator :string [schema options] (-string-gen schema options))
 
 ;; TODO: ref-max need to favor the non-recursive part, not to fail
-(defmethod -schema-generator :ref [schema {::keys [ref-max] :or {ref-max 10} :as options}]
+(defmethod -schema-generator :ref [schema {::keys [ref-max] :or {ref-max 50} :as options}]
   (let [ref (first (m/children schema options))
         ref-count (get-in options [::ref-count ref] 0)]
     (if (< ref-count ref-max)
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -91,7 +91,7 @@
 (defmethod -schema-generator :tuple [schema options] (apply gen/tuple (mapv #(generator % options) (m/children schema options))))
 #?(:clj (defmethod -schema-generator :re [schema options] (-re-gen schema options)))
 (defmethod -schema-generator :string [schema options] (-string-gen schema options))
-(defmethod -schema-generator :ref [schema {::keys [ref-max] :or {ref-max 10} :as options}]
+(defmethod -schema-generator :ref [schema {::keys [ref-max] :or {ref-max 100} :as options}]
   (let [ref (first (m/children schema options))
         ref-count (get-in options [::ref-count ref] 0)]
     (if (< ref-count ref-max)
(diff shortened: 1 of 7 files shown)


ikitommi:
Recursive schema resolution is too eager now:

```clj
(m/form
  [:and
   {:registry {::ping ::pong2
               ::pong2 ::pong
               ::pong [:and pos-int?]}}
   ::ping])
;[:and {:registry {:user/ping [:and pos-int?]
;                  :user/pong2 [:and pos-int?]
;                  :user/pong [:and pos-int?]}} 
; [:and pos-int?]]
```
, should retain the pointer information.

src/malli/generator.cljc
```diff
@@ -40,16 +46,23 @@
...
+(defn -or-gen [schema options]
+  (gen/one-of (keep #(some->> (-recursion-options % options) (generator %)) (m/children schema options))))
+
+(defn -multi-gen [schema options]
+  (gen/one-of (keep #(some->> (-recursion-options (last %) options) (generator (last %))) (m/map-entries schema options))))
+
 (defn -map-gen [schema options]
   (let [entries (m/map-entries schema)
-        value-gen (fn [k s] (gen/fmap (fn [v] [k v]) (generator s options)))
+        options' (-recursion-options schema options)
```
rschmukler, line 57: Perhaps this is better filed as a bug with them, but just worth noting that the "prime" convention isn't well supported in some editors' understanding of the clojure grammar. Emacs will attempt to convert `options'` into `options '` causing the expression to rebind `options` to the quoted expression `'(-recursion-options schema options)`. Your haskell is showing ;)

rschmukler (commented):
Awesome! Left a super nitpicky comment that really doesn't warrant any changes. Code looks really nice, as always.

Issue #182, opened 2020-02-22
Rough edges for swagger documentation


piotr-yuxuan:
Hi, thanks for this project which definitely is awesome! I've stumbled across this behaviour which one might counter-intuitive:

One can't describe a map attribute like this:
``` clojure
(malli.swagger/transform
  [:map [:my-uuid {:optional true
                   :swagger/example #uuid"eb4acd14-5321-4317-82e2-5895269f18a8"
                   :swagger/description "your most awesome uuid"}
                  uuid?]])

;; =>
{:type "object", 
 :properties {:my-uuid {:type "string", :format "uuid"}}, 
 :required []}
```
`:optional` is taken into account but `:swagger/*` keywords are ignored. As a result one has to follow this syntax with two attribute maps:
``` clojure
(malli.swagger/transform
  [:map [:my-uuid {:optional true}
                  [:and {:swagger/example #uuid"eb4acd14-5321-4317-82e2-5895269f18a8"
                         :swagger/description "your most awesome uuid"}
                        uuid?]]])
;; =>
{:type "object",
 :properties {:my-uuid {:type "string",
                        :format "uuid",
                        :x-allOf [{:type "string", :format "uuid"}],
                        :example #uuid"eb4acd14-5321-4317-82e2-5895269f18a8",
                        :description "your most awesome uuid"}},
 :required []}
```
For the sake of completion, this following syntax doesn't produce the expected result either as `:optional` now is ignored:
``` clojure
(malli.swagger/transform
  [:map [:my-uuid [:and {:optional true
                         :swagger/example #uuid"eb4acd14-5321-4317-82e2-5895269f18a8"
                         :swagger/description "your most awesome uuid"}
                        uuid?]]])

;; =>
{:type "object",
 :properties {:my-uuid {:type "string",
                        :format "uuid",
                        :x-allOf [{:type "string", :format "uuid"}],
                        :example #uuid"eb4acd14-5321-4317-82e2-5895269f18a8",
                        :description "your most awesome uuid"}},
 :required [:my-uuid]}
```
Would you agree with me to qualify this as a bug? would you be interested in fixing it, or open to pull requests about that?

Using version `"0.4.2"`

Cheers!

ikitommi:
Thanks for reporting, this is duplicate to #86. Need to figure out a clean solution for this, options being:

1) implement the parent/child relationship between entry & child schema (complex?)
[...]

Issue #120, opened 2019-11-24
Define Child/Parent relationship to Schemas


ikitommi:
If schemas knew their parent, we would be able to do:

* use properties of mother in children, e.g. error definitions (#86) 
```clj
;; use error message of a parent (here: map-entry schema)
[:map [:x {:error/message "not an int"} int?]]
```

* use mother for complex transformations #119 

Is this good or bad?

rschmukler:
Just echoing some conversation seen in other issues.

For complex transformations, having interceptors (#122, #114) takes care of most things. Any transformer that requires parent access can simply become an `:after` transformer on the parent type. I will admit that this, theoretically, doesn't compose as well, nor remain as focused - however I can't think of any _real_ use cases where this ends up being a big deal.

flow-danny:
I think I have a use-case for this. 

In `reitit` nesting of routes allows one to define what sub-routes share access roles. I would like to define in a similar way the access level for a group of attributes. Then I could filter/transform data (anonymize, or omit keys) based on the user access level.

ikitommi:
#212 is merged in master. Is allows all but error messages to be read from entry schemas.

```clj
(require '[malli.json-schema :as json-schema])

(json-schema/transform
  [:map
   [:x {:json-schema/example 42} int?]
   [:y {:json-schema {:type "string"}} [:fn (fn [x] (string? x))]]])
;{:type "object"
; :properties {:x {:type "integer"
;                  :format "int64"
;                  :example 42}
;              :y {:type "string"}}
; :required [:x :y]}
```

because of that, we don't need childs to know about their parents.

Issue #116, opened 2019-11-12
Things to be decided before initial stable release


ikitommi:
* [x] Recursive schemas #96 
* [x] Inherit schema properties from map entries? #86
* [x] explain data to be symmetric with `LensScema` options #238 
* [x] Does value parsers (e.g. spec conform) effect the api? --> can be done later
* [x] Support by default type-schemas, predicate-schemas or both? #25-> both, for alpha
* [x] Customizable default options #228 --> later
* [x] Open or Closed maps by default #31
* [x] Re-visit priority of encoders & decoders #99

green-coder:
I suggest to add "recursive models" to the list.

ikitommi:
Added that and #86 to the list.

ikitommi:
one to go.

ikitommi:
two to go.

green-coder:
Something special with sequences,  it's that parsing a data structure can result in multiple ways to match the data, i.e. to put a name on the parts of the data.

It is useful to keep that in mind and see if it affects the API before making a release.

ikitommi:
Parsing (and Schema-variants with named branches) could be implemented as an increment using either `-explain` or `-transform`. Both could be non-breaking increments on top of the initial api, but might not -> I'll add to the list.

Using `-explain` (to optionally collect both the successful parse-tree & errors) would be interesting, as we could return partial parse trees with errors in place. 

compared to spec:

```clj
(require '[clojure.spec.alpha :as s])

(def spec (s/+ (s/cat :odds (s/+ odd?)
                      :even (s/? even?))))

(s/conform spec [1 3 5 100 1 20])
; [{:odds [1 3 5], :even 100} {:odds [1], :even 20}]

(s/explain-data spec [1 3 5 100 1 20 40])
;#:clojure.spec.alpha{:problems ({:path [:odds], :pred clojure.core/odd?, :val 40, :via [], :in [6]}
;                                {:path [:odds], :pred clojure.core/odd?, :val 40, :via [], :in [6]}
;                                {:path [:odds], :pred clojure.core/odd?, :val 40, :via [], :in [6]}
;                                {:path [:odds], :pred clojure.core/odd?, :val 40, :via [], :in [6]}),
;                     :value [1 3 5 100 1 20 40]}
```

ikitommi:
closing this, there is a internally working prototype of regeexs, work to get integrated into `explain` and potentially new `destructure` and `restructure`. but can be a later additon to the api,

Issue #80, opened 2019-10-07
More Robust Humanized Errors


ikitommi:
`malli.error/error-message` is great, but not complete. It should yield basic errors for all core error conditions. Fix at least the following:

* default messages for all built-in schemas:

```clj
(-> [:maybe int?]
    (m/explain "1")
    (me/humanize {:wrap :message}))
; "unknown error"
```

* error properties should be looked from father schemas in case of aggregates:

```clj
(-> [:and {:error/message "over 18"} int? [:> 18]]
    (m/explain 12)
    (me/humanize {:wrap :message}))
; "unknown error"
```

* messages for type mismatch:

```clj
(-> [:list int?]
    (m/explain #{1 2 3})
    (me/humanize {:wrap :message}))
; "unknown error"
```

rlovtangen:
```
(-> [:and {:error/message "over 18"} int? [:> 18]]
    (m/explain 12)
    (me/humanize {:wrap :message}))
```

Is this example now supposed to return the provided message "over 18"? 
It returns only the message from `[:> 18]`: `["should be larger than 18"]`
Malli 0.9.2

PR #212, opened 2020-07-04
Schema Applications can use MapEntry properties

ikitommi:
Spike on whether and how `MapEntry` properties should used in schema applications:

* [x] forms
* [x] transformation
* [x] default transformer
* [x] ~error messages~ (postponed)
* [x] value generation
* [x] json schema transformation
* [x] swagger2 transformation

Related issues: #80, #86, #116,  #120, #182
Related PRs: #119, #197

### forms

* should not change (e.g. not merging data to children)

```clj
(m/form [:map [:x {:encode/math inc} int?]])
; => [:map [:x {:encode/math inc} int?]]
```

### transformation 

* should allow extra property-based transformations
* should have a target key for `:encoders` & `:decoders`?

```clj
(m/decode 
  [:map [:x {:encode/math inc} [int? {:encode/math inc}]]]
  {:x 1}
  (mt/transformer {:name :math}))
; => {:x 3}
```

### error messages

* ~should use the parent props if defined~
* despite the orginal cause (#86), not resolving this with this PR. Handling errors is a special case, will resolve that later

```clj
(-> [:map [:x {:error/message "over 18"} [:and int? [:> 18]]]
    (m/explain {:x 12})
    (me/humanize))
; {:x ["over 18"]}
```

* also for composite schemas

```clj
(-> [:and {:error/message "over 18"} int? [:> 18]]
    (m/explain 12)
    (me/humanize))
; ["over 18"]
```

### value generation

* should use the parent props if defined

```clj
(mg/generate [:map [:foo {:gen/elements [1 2]} int?]])
; => {:foo 1}
```

### json-schema

* should use merged properties (parent wins)

```clj
(json-schema/transform [:map [:x {:json-schema {:type "numberz"}} int?]])
;{:type "object",
; :properties {:x {:type "numberz"},
; :required [:x]}
```

```clj
(json-schema/transform [:map [:x {:json-schema/type "numberz"} [:schema {:json-schema/format "int32"} [int? {:json-schema/example 1}]]]])
;{:type "object",
; :properties {:x {:type "numberz", :format "int32", :example 1},
; :required [:x]}
```

### swagger

* like with json-schema

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ We use [Break Versioning][breakver]. The version numbers follow a `<major>.<mino
 
 ## breaking changes in the pre-alpha era:
 
+* ????
+  * `m/children` returns 3-tuple (key, properties, schma) for `MapSchema`s
+  * `m/map-entries` is removed, `m/entries` returns 2-tuple (key, entry-schema)  
 * 4.8.2020
   * `:path` in explain is re-implemented: map keys by value, others by child index
   * `m/-walk` and `m/Walker` uses `:path`, not `:in`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ We use [Break Versioning][breakver]. The version numbers follow a `<major>.<mino
 
 * ????
   * `m/children` returns 3-tuple (key, properties, schma) for `MapSchema`s
-  * `m/map-entries` is removed, `m/entries` returns 2-tuple (key, entry-schema)  
+  * `m/map-entries` is removed, `m/entries` returns a `MapEntry` of key & `m/-val-schema`  
 * 4.8.2020
   * `:path` in explain is re-implemented: map keys by value, others by child index
   * `m/-walk` and `m/Walker` uses `:path`, not `:in`
--- a/docs/tips.md
+++ b/docs/tips.md
@@ -20,8 +20,8 @@ Schemas can be walked over recursively using `m/walk`:
   (fn [schema _ children options]
     ;; return nil if Schema has the property 
     (when-not (:deleteMe (m/properties schema))
-      ;; there are two syntaxes: normal and the map-entry, handle separatly
-      (let [children (if (m/map-entries schema) (filterv last children) children)]
+      ;; there are two syntaxes: normal and the entry, handle separatly
+      (let [children (if (m/entries schema) (filterv last children) children)]
         ;; create a new Schema with the updated children, or return nil
         (try (m/into-schema (m/type schema) (m/properties schema) children options)
              (catch #?(:clj Exception, :cljs js/Error) _))))))
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -54,7 +54,7 @@
 ;; impl
 ;;
 
-(declare schema schema? into-schema into-schema? eval default-registry -schema-schema -registry)
+(declare schema schema? into-schema into-schema? eval default-registry -schema-schema -entry-schema -registry)
 
 (defn -keyword->string [x]
   (if (keyword? x)
@@ -96,7 +96,7 @@
   (mapv (fn [[i c]] (-inner walker c (conj path i) options)) (map-indexed vector children)))
 
 (defn -inner-entries [walker path entries options]
-  (mapv (fn [[k p s]] [k p (-inner walker s (conj path k) options)]) entries))
+  (mapv (fn [[k p s]] [k p (-inner walker (-entry-schema s p) (conj path k) options)]) entries))
 
 (defn -get-entries [schema key default]
   (or (some (fn [[k _ s]] (if (= k key) s)) (-map-entries schema)) default))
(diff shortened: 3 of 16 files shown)


ikitommi:
Tested if there is a quick fix for this using `m/parent-properties` and generic parent handling. Not looking good.

Idea was, that for contained schemas (`:and`, `:or`, `:map/entry` etc) one can look up the parent properties. This would resolve all the cases to some degree but the code and whole contract seems hacky.

For example, one should check `:map/entry` and `:and` & `:or` differently in JSON Schema. Generic solution would be good target but too greedy:

```clj
(defn- -json-schema-visitor [schema children _in options]
  (let [p (m/properties schema)
        pp (m/parent-properties schema)]
    (or (unlift pp :json-schema)
        (unlift p :json-schema)
        (merge (select p)
               (select pp)
               (if (satisfies? JsonSchema schema)
                 (-accept schema children options)
                 (accept (m/type schema) schema children options))
               (unlift-keys p :json-schema)
               (unlift-keys pp :json-schema)))))
```

=>

```clj
(json-schema/transform
  [:and {:json-schema/title "and"}
   int?
   boolean?])
;{:allOf [{:type "integer", :format "int64", :title "and"} 
;         {:type "boolean", :title "and"}]
; :title "and"}
```

Just adding the parent handling to `:map/entry` (not `:and` and `:or`) would not resolve the error cases and would be a new mechanism just for `:map/entry`.

Pushing the special code into JSON Schema `:map` would also be bad, expecially as `:swagger` composes on top of `:json-schema`: 

```clj
(defmethod accept :map [_ _ children _]
  (let [required (->> children (filter (comp not :optional second)) (mapv first))]
    {:type "object"
     :properties (apply array-map (mapcat (fn [[k p s]]
                                            [k (or (unlift p :json-schema)
                                                   (merge (select p)
                                                          (unlift-keys p :json-schema)
                                                          s))]) children))
     :required required}))
```

and copy-paste (previously not neede code!) to swagger:

```clj
(defmethod accept :map [_ _ children _]
  (let [required (->> children (filter (comp not :optional second)) (mapv first))]
    {:type "object"
     :properties (apply array-map (mapcat (fn [[k p s]]
                                            [k (or (json-schema/unlift p :swagger)
                                                   (json-schema/unlift p :json-schema)
                                                   (merge
                                                     (json-schema/select p)
                                                     (json-schema/unlift-keys p :json-schema)
                                                     (json-schema/unlift-keys p :swagger)
                                                     s))]) children))
     :required required}))
```

Based on the few hours spent on this, it looks like we should have first class (internal) `MapEntry` Schema that is returned in walking. So, intead:

```clj
(m/to-map-syntax
  [:map
   [:x int?]
   [:y {:optional true} int?]])
;{:type :map
; :children [[:x nil {:type int?}] 
;            [:y {:optional true} {:type int?}]]}
```

this:

```clj
(m/to-map-syntax
  [:map
   [:x int?]
   [:y {:optional true} int?]])
;{:type :map
; :children [[:x {:type :map/entry
;                 :children [{:type int?}]}]
;            [:y {:type :map/entry
;                 :properties {:optional true}
;                 :children [{:type int?}]}]]}
```

Hmm.

ikitommi:
Heureka! Conditional walking.

ikitommi:
All work but the error messages.

ikitommi:
challenge with errors is that errors are reported currently on the failing schema, for composite schemas like `:and`,  `:or` and `::m/entry`, the error messages should be looked from parent schemas. From how far? just from one level up? Should be collected top-down (like everything else)?

top-down would mean:

```clj
(-> [:map 
     [:x {:error/message "from entry, over 18"} 
      [:and {:error/message "from and, over 18"} 
       int? [:> 18]]]
    (m/explain {:x 12})
    (me/humanize))
; {:x ["from entry, over 18"]}
```

should it be always looked up?

```clj
(-> [:map 
     [:x {:error/message "from entry, over 18"} 
      [:and {:error/message "from and, over 18"} 
       int? [:> {:error/message "from :>, over 18"} 18]]]
    (m/explain {:x 12})
    (me/humanize))
; {:x ["from entry, over 18"]}
```

... currently taken from the schema itself, if present. unlike everything else.

ikitommi:
Reletivaly happy with this now:

* real, optionally walkable `-entry-schema`s
* `m/children` returns 3-tuple (key, properties, schma) for `MapSchema`s
* `m/map-entries` is removed, `m/entries` returns 2-tuple (key, entry-schema)

This would break everyone using `m/map-entries`. Should we add it back, using the old contract, on top of the new apis?

rschmukler:
@ikitommi Thanks for all your thought on this issue. Got a chance to play with the latest code and it's working great.

Re. your question, I would break `m/map-entries` while we can. The difference between `m/entries` and `m/children` is already somewhat confusing (at first) - to me, keeping old API baggage from alpha doesn't seem worth it. 

I think as we near release it will be worth adding doc-strings to both functions that explain when to use which one. My understanding (and please, correct me if I am wrong) is that most transformers would likely use `m/children` since it handles unwrapping the internals of `MapEntry` schema for you, while walking and schema transformations are better served by having direct access to the MapEntry AST node directly.

Let me know if you're close to merging and I'd be happy to do a code review if you want one :beer:

rschmukler (approved):

ikitommi:
Thanks for the review! Will add the docstrings.

ikitommi:
some docs in https://github.com/metosin/malli/commit/3443411a3ada8809937efd25de6a12bce1899188

PR #213, opened 2020-07-07
Fix default `false` value when applying transformations

mgxm:
The default transformation was not applied when a `false` value was expected.


ikitommi:
thanks!

PR #216, opened 2020-07-09
Replace Schema Visitor with Schema Walker

ikitommi:
Re-implement Schema Visitor using Walker. Enables `find-first` to (almost) short-circuit when a match is found.

```clj
(m/find-first
  [:map
   [:x int?]
   [:y [:vector [:tuple
                 [:maybe int?]
                 [:or [:and {:salaisuus "turvassa"} boolean?] int?]
                 [:schema {:salaisuus "vaarassa"} false?]]]]
   [:z [:string {:salaisuus "piilossa"}]]]
  (fn [s _in _options]
    (some-> s m/properties :salaisuus)))
; => "turvassa"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -629,7 +629,7 @@ Schema unions (merged values of both schemas are valid for union schema):
 Adding generated example values to Schemas:
 
 ```clj
-(m/accept
+(m/walk
   [:map
    [:name string?]
    [:description string?]
@@ -637,7 +637,7 @@ Adding generated example values to Schemas:
     [:map
      [:street string?]
      [:country [:enum "finland" "poland"]]]]]
-  (m/schema-visitor
+  (m/schema-walker
     (fn [schema]
       (mu/update-properties schema assoc :examples (mg/sample schema {:size 2, :seed 20})))))
 ;[:map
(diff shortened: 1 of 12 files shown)


src/malli/core.cljc
```diff
@@ -935,6 +948,32 @@
...
+  ([?schema f options]
+   (-walk
+     (schema ?schema options)
+     (reify Walker
+       (-accept [_ s _ _] s)
+       (-inner [this s in options] (-walk s this in options))
+       (-outer [_ s c in options] (f s c in options)))
+     [] options)))
+
+(defn ^:no-doc find-first
```
rschmukler, line 963: This seems like a useful / worth exporting into `clj-doc` - but maybe it should live under `malli.utils` like`merge`, `assoc`, etc?

README.md
```diff
@@ -925,9 +925,9 @@ The identity visitor:
 Adding `:title` property to schemas:
 
 ```clj
-(m/accept
+(m/walk
   Address
-  (m/schema-visitor #(mu/update-properties % assoc :title (name (m/type %)))))
+  (m/schema-walker #(mu/update-properties % assoc :title (name (m/type %)))))
```
rschmukler, line 930: Could this be wrapped by `m/walk` - ie. If we are always going to have to wrap the function with `m/schema-walker`, why not just have `m/walk` do it for us?

Similar to how the `m/encode` will auto coerce non-reified schemas using `m/schema`

rschmukler (commented):
A few random comments but overall this looks great!

rschmukler (approved):
A few random comments but overall this looks great!

README.md
ikitommi: I think it covers the different cases nicely right now:

* `m/-walk` for any kind of walking, for library extensions
* `m/walk` with 4-arity fn for advanced postwalking (e.g. to map-syntax)
* `m/walk` + `m/schema-walker` to postwalk Schema instances to Schema instances, easy for many cases

src/malli/core.cljc
ikitommi: good point! Will move it to `malli.util` and make public & add to README.

ikitommi: done

PR #221, opened 2020-07-14
improved english enum error message

stevebuik:


stevebuik:
no probs. I had a check for 2 already but added 1, 2, 3 & 4 with better tests. 

Is that what you meant?

ikitommi:
Looks great, thanks!

Issue #218, opened 2020-07-09
:default false doesn't work


kwrooijen:
```clojure
(require '[malli.core :as m])
(require '[malli.transform  :as mt])
(m/decode [:map [:person/verified [:and {:default false} boolean?]]] {} mt/default-value-transformer)
; => {}
(m/decode [:map [:person/verified [:and {:default true} boolean?]]] {} mt/default-value-transformer)
; => #:person{:verified true}
```
I expected the first decode to return `#:person{:verified false}`

This is probably because `:default` is checked somewhere to see if it's a truthy value. The question is, is this a bug, or intended? Seems a bit misleading. Maybe we should check if the key `:default` exists rather than check if it's a truthy value?

rschmukler:
In an attempt to verify (and fix) this, I added a test to `transformers_test.cljc` that looked like:

```clj
(testing "falsey defaults"
    (let [schema [:map [:person/verified [:and {:default false} boolean?]]]]
      (is (= {:person/verified false}
             (m/decode schema {} mt/default-value-transformer)))))
```

The test passed for me. Which version of malli did you run into this on? Can you confirm that it doesn't behave as expected on master.

kwrooijen:
You're right, it works. I should've updated before reporting this issue, sorry about that. Funny how it was fixed so recently, thanks!

PR #223, opened 2020-07-15
Return false with default-value-transformer

kwrooijen:
Fixes https://github.com/metosin/malli/issues/218


ikitommi:
What if someone would like to set `nil` as a default? For maps keys, it might make sense. Could be a separate issue to expand this so that `nil` would work too. e.g. ANY default value would be ok.

kwrooijen:
In the issue I mentioned that maybe it would be better if we check for the existence of the `:default` key, rather than checking if it's truthy. I think that would solve your last comment. Maybe make that a separate issue? I'll fix the indentation for now.

kwrooijen:
@ikitommi fixed indentation

Issue #203, opened 2020-06-05
Make SCI dependency optional for smaller JS builds


Naomarik:
Currently in the context of my application I will never have a need for it and it adds 93kb gzipped at compression level 6.

ikitommi:
Did a quick test using shadow-cljs:

```clj
metosin/malli 43.1 KB
borkdude/sci  63.6 KB
```

The malli is currently not optimized either for bundle size. Will make sci optional and try to make malli smaller too.

ikitommi:
Added shadow-cljs as dev-dependency and guide how to measure the bundle sizes. With the latest versions (sci got docstrings, making it a lot bigger):

```
metosin/malli 50KB -> 11.8KB GZIP
borkdude/sci  564KB - > 120KB GZIP
```

ikitommi:
Not sure how correctly the bundle size is calculated, but here goes the report:

<img width="1052" alt="Screenshot 2020-06-12 at 8 54 51" src="https://user-images.githubusercontent.com/567532/84469881-c04da600-ac8a-11ea-8f03-66d5ca81381c.png">

borkdude:
Work in progress at #210. It works, but the tests need to be refactored to account for the optionality of sci.

borkdude:
I have closed the PR. Instead, I'm offering you this library https://github.com/borkdude/dynaload. If you need any help, let me know. :)

ikitommi:
thanks, will try that :)

ikitommi:
Will be fixed in #227 .

ikitommi:
sci is now disabled by default. README has guide how to enable it for cljs. With the minimal dummy sample app:

* js: 757,8kb -> 136kb
* gzipped: 161kb -> 31,5kb

# with sci:

<img width="911" alt="with-sci" src="https://user-images.githubusercontent.com/567532/88318717-d340a500-cd23-11ea-88ae-08e86600ad69.png">

# without (default):

<img width="910" alt="no-sci" src="https://user-images.githubusercontent.com/567532/88318733-d9cf1c80-cd23-11ea-8673-d222407839bd.png">

PR #227, opened 2020-07-23
Optional SCI

ikitommi:
load `borkdude/sci` using `borkdude/dynaload` to enable:
* faster standalone usage for JVM (2.5sec -> 0.5sec to load `malli.core`)
* smaller js-bundles: 120kb -> 7kb

bit hacky, but works. in future sci should be enabled using explicit option, like:

```clj
(m/schema 
  [:fn '(fn [x] (> x 10))] 
  {:evaluator malli.sci/evaluator})
```

# example app

https://github.com/metosin/malli/blob/master/app/malli/app2.cljc

## with sci

with-sci

## without sci

no-sci


ikitommi:
Will fix #203.

PR #232, opened 2020-07-23
Add Swagger Enum type to the schema.

### LINKED ISSUE #231: Enums need Swagger type "enum" set to render on the UI
Hi,

Currently, if using enums, such as

`[:enum  "A" "B" "C" "D"]`

The Swagger UI will not show the values A, B, C and D in the list of permittable options.

Adding this will allow Swagger to do so:

`[:enum  {:swagger/type "enum"} "A" "B" "C" "D"]`

I have a PR upcoming which will add in the swagger type automatically if "enum" is the type.

-=david=-

dharrigan:
This will allow the Swagger UI to display the list of valid elements.

Fixes #231

-=david=-

### DIFF
--- a/src/malli/swagger.cljc
+++ b/src/malli/swagger.cljc
@@ -22,6 +22,8 @@
 
 (defmethod accept :tuple [_ _ children _] {:type "array" :items {} :x-items children})
 
+(defmethod accept :enum [_ _ children _] {:type "enum" :enum children})
+
 (defn- -swagger-walker [schema children _in options]
   (let [p (m/properties schema)]
     (or (json-schema/unlift p :swagger)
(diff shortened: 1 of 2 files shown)


ikitommi:
I don't think `:enum` is a valid type. Using swagger editor:

Screenshot 2020-07-23 at 19 37 55

currently, there is no inferring of enum type (https://github.com/metosin/malli/blob/master/src/malli/core.cljc#L572), so unless we implement such, one needs to hint the type manually:

```clj
(swagger/transform [:enum {:swagger/type "string"} "1" "2"])
; => {:enum ["1" "2"], :type "string"}
```

which is kinda ok IMO.

dharrigan:
Hi,

Interesting, I wonder why when I did a type hint of `:swagger/type "enum"` values came out...<shrug> I'm happy enough to use a type hint `:swagger/type "string"`, so feel free to close this PR if you wish :-)

PR #233, opened 2020-07-23
Improve regex validation with humanized message

### LINKED ISSUE #226: Improve regex validation humanized message
When a regex validation fails, it reports the generic `"unknown error"`:

```Clojure
(malli.error/humanize (malli.core/explain [:re #"bla"] "bla"))
=> nil
(malli.error/humanize (malli.core/explain [:re #"bla"] "gogo"))
=> ["unknown error"]
```

I believe a more focused error would be more appropriate.

Candidates might include `"should match a regex"`,  `"should be a match"`, `"invalid value"`, etc.

dharrigan:
Fixes #226

-=david=-


ikitommi:
thanks!

PR #234, opened 2020-07-24
Add in :length as a option for String.

dharrigan:
This option sets the exact length that a String must be.

-=david=-

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,14 @@ Using `:string` Schema:
 (m/validate [:string {:min 1, :max 4}] "")
 ; => false
 ```
+(m/validate [:string {:length 3}] "ab")
+; => false
+
+(m/validate [:string {:length 3}] "abc")
+; => true
+
+(m/validate [:string {:length 3}] "abcd")
+; => false
 
 ## Function schemas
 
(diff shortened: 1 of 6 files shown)


ikitommi:
copied from slack:

---

You can already say `[:string {:min 4, :max 4}]` , which is not that much longer than `[:string {:length 4}]`. And you can always uses aliases to save on typing:

```clj
[:schema 
 {:registry {::str4 [:string {:min 4, :max 4}]}} 
 ::str4]
```

or:

```clj
(def string4 [:string {:min 4, :max 4}])
(m/validate string4 "kikka")
; => false
```

there could be a pretty error formatter for that, e.g.

```clj
(and min (= min max)) (str "should be " min " characters")
```

=>

```clj
(-> [:schema {:registry {::str4 [:string {:min 4, :max 4}]}} ::str4]
    (m/explain "kikka")
    (me/humanize))
; => ["should be 4 characters"]
```

ikitommi:
but thanks for suggesting this.

ikitommi:
https://github.com/metosin/malli/commit/22df982f3428b4ee2b0f7ca72e9885edc318728c

PR #235, opened 2020-07-25
DRAFT: Default options

ikitommi:
Instead of JVM/closure-defines overridable default registry, we can swap whole default options. This allows easy way to customize (sci) evalutor, in future default localizations etc.

Compiler/jvm bootstrap:
   * cljs: `:closure-defines {malli.options/type "custom"}`
   * clj: `:jvm-opts ["-Dmalli.options/type=custom"]`

Available types:
  * `default` -> the default registry + dynamic sci
  * `custom` -> empty registry + dynamic sci
  * `empty` -> empty registy + no sci

### DIFF
--- a/app/malli/app2.cljc
+++ b/app/malli/app2.cljc
@@ -3,24 +3,24 @@
 ;; - npx shadow-cljs release app2 --pseudo-names
 (ns malli.app2
   (:require [malli.core :as m]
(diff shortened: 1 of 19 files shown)


ikitommi:
this has diverged too much from the current malli, issue (#236) still open to resolve this.

PR #237, opened 2020-07-26
Improve string validation humanized messages

### LINKED ISSUE #225: Improve string validation humanized messages
Without `min` and/or `max`, when the value is not a String, it reports correctly `"should be a string"`:

```Clojure
(malli.error/humanize (malli.core/explain [:string] "Bla"))
=> nil
(malli.error/humanize (malli.core/explain [:string] 1))
=> ["should be a string"]
```

With `min` and/or `max`, when the value is not a String, it reports a message indicating invalid length of the String:

```Clojure
(malli.error/humanize (malli.core/explain [:string {:max 3}] "Bla"))
=> nil
(malli.error/humanize (malli.core/explain [:string {:max 3}] "Gogo"))
=> ["should be at most 3 characters"]
(malli.error/humanize (malli.core/explain [:string {:max 3}] 1))
=> ["should be at most 3 characters"]
```

I believe this is misleading, and the `"should be a string"` message should be reported in this case, too.

(Thanks for the malli library! Finally something that can actually be used for user input validation in Clojure.)

ivos:
Fixes #225

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -67,10 +67,10 @@
                                   (first (m/children schema))
                                   (str "either " (->> (m/children schema) butlast (str/join ", "))
                                        " or " (last (m/children schema))))))}}
-   :string {:error/fn {:en (fn [{:keys [schema]} _]
+   :string {:error/fn {:en (fn [{:keys [schema value]} _]
                              (let [{:keys [min max]} (m/properties schema)]
                                (cond
-                                 (not (or min max)) "should be a string"
+                                 (not (string? value)) "should be a string"
                                  (and min max) (str "should be between " min " and " max " characters")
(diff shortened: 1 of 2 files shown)


ikitommi:
thanks!

ivos:
Happy to help.

Just a side note, having this function in `default-errors` feels a bit odd to me. I would rather expect `default-errors` to be a simple map of error codes to error messages. The string validation would then have to return different error codes, depending on which sub-validation failed (either not a string or invalid string length).

Also, this `(not (string? value))` fragment is actually repeating part of the validation logic in error reporting code, which I'd say is wrong. Having the validation itself return failure codes would resolve both these concerns, but will pro'lly require some redesign :(

PR #240, opened 2020-08-09
Document `:map-of` to model homogenous maps

teodorlu:
- I didn't put any thought into where in the readme this belongs, feel free to reorganize.
- The explanations also feel a bit wordy to me, they might benefit from some slimming down.


ikitommi:
thanks!

PR #245, opened 2020-08-18
fix a warning

lucywang000:
When using latest master, shadow-cljs reports:

```
Resource: malli/core.cljc:819:9
Bad method signature in protocol implementation, RefSchema does
not declare method called -keep
```


ikitommi:
thx!

PR #247, opened 2020-08-29
feat: add `pos?` and `neg?` to `predicate-schemas`

rschmukler:
Adds the core `pos?` and `neg?` functions to the predicate schemas. This
is useful for when you don't care about the type of numeric, (ie. int vs
double vs float) and only the sign of it.

Also useful for constructing `[:and pos? float?]` if you want positive floats.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1118,7 +1118,7 @@
 ;;
 
 (defn predicate-schemas []
-  (->> [#'any? #'some? #'number? #'integer? #'int? #'pos-int? #'neg-int? #'nat-int? #'float? #'double?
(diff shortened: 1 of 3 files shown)


ikitommi (commented):
For completeness, there should be json-schema mappings & generators for these too.

rschmukler:
@ikitommi Thanks for the review. Updated w/ generators and json-schema mappings.

ikitommi:
Thanks!

ikitommi:
there was one more place: https://github.com/metosin/malli/commit/a6d38fa538cd6f0f7ccd58c938a80fc82f1df844

ikitommi:
@rschmukler coudn't make you a reviewer on #212 , but would you have time to take a look at it? if the diff is too messy (it is), I could cherry pick the code formatting & moving around into separate issue.

rschmukler:
@ikitommi will take a look! Thanks for the help getting this merged.

PR #267, opened 2020-10-02
Minor: Fix misspelling in default error message for some?

tekacs:
I spotted this in an error presented to me on my client. :)


ikitommi:
thanks!

PR #277, opened 2020-10-11
explicitly disabling sci & sci options, fixes #276

### LINKED ISSUE #276: Make sci explicitly optional
Currently, if sci is a transitive dependency or is required by some other component, it will be enabled for Malli too. This is not good. Real solution would be to make installation of sci explicit, e.g. via #235: 

## 1) explicit sci via options

```clj
(require '[malli.sci :as ms])
(require '[malli.core :as m])

(def options
  {:evaluator (ms/evaluator)
   :registry (ms/default-registry)})

(def Schema (m/schema [:fn '(fn [x] (string? x))] options))

(m/validate Schema "kikka")
; => true
```

## 2) disable via flag

, but, as it would be a breaking change (should wait for 1.0.0?), we could just add a flag wether to use sci or not.

```clj
(m/schema [:fn '(fn [x] (string? x))] {::m/enable-sci false})
; Exectution error: :malli.core/sci-not-available {:code (fn [x] (string? x))}
```

ikitommi:
Option 2 for #276. A non-breaking change (opposed to option1).

Disabling `sci`:

```clj
(m/validate [:fn 'int?] 1)
; => true

(m/validate [:fn 'int?] 1 {::m/disable-sci true})
; Execution error
; :malli.core/sci-not-available {:code int?}
```

Custom `sci` options:

```clj
(m/validate [:fn '(fn [x] (< x (last (range 10))))] 1)
; => true

(m/validate [:fn '(fn [x] (< x (last (range 10))))] 1 {::m/sci-options {:realize-max 1}})
; => false
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ We use [Break Versioning][breakver]. The version numbers follow a `<major>.<mino
 
 Malli is in [alpha](README.md#alpha).
 
+## UNRELEASED
+
+* New options
+  * `:malli.core/disable-sci` for explicitly disabling `sci`, fixes [#276](https://github.com/metosin/malli/issues/276)
+  * `:malli.core/sci-options` for configuring `sci`
+
 ## 0.1.0 (2020-10-08)
 
 First stable release.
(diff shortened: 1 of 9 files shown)


miikka:
Doing this still keeps the door open for a breaking change

PR #280, opened 2020-10-14
Timestamp frac precision

katox:
I have been bitten by round-trip `java.time.Instant` -> `string` -> malli -> `#inst`. The problem manifests in java9+ because of increased precision from miliseconds to microseconds. I noticed there is a broader plan for time handling code but in the meantime I patched the `string->date` transformer.


miikka (approved):

miikka:
Thanks for the fix!

PR #291, opened 2020-10-30
Set actual Oslo coordinates

teodorlu:
I set the old coordinates just to have an example, but re-reading the docs now, I feel that it would be more proper not to confuse people!

So here's a coordinate-fixing PR for just docs.

Have a nice weekend!

Teodor


ikitommi:
Thanks!

PR #293, opened 2020-11-03
Add example for test.check generator integration

solatis:
Small snippet that documents test.check generator integration by using `mg/generator`.


ikitommi:
thanks!

PR #296, opened 2020-11-09
Add JSON Schema conversion for `:=`

dakra:
See https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

`const` got added in JSON Schema version 6. I couldn't find the minimum version that malli supports and if depending on version 6+ is ok or not.


ikitommi:
Looks good, I guess it won't work with swagger, which defaults to JSON Schema mappings, but, most likely the swagger ui respects this with swagger.

Thanks!

PR #299, opened 2020-11-10
Memoized schema

solatis:
First version, still needs some tests, but let me know whether this is in line with what you had in mind, and whether there are any code style problems.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1064,7 +1064,7 @@
              (-> ?schema (-schema options) (schema options))))))
 
 (defn memoized-schema [s]
-  (-memoized-schema (-schema s)))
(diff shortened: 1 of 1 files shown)


ikitommi:
Hi. looks good. I now think that this should be in `malli.util` as it's not a mandatory part of the core, but an add-on. About the name, should it be `memoizing-schema` or `caching-schema`? as the schema itself is not memoized, just parts of it.

solatis:
Oh sorry I completely missed your feedback -- I'll incorporate it ASAP!

ikitommi:
The malli codebase has evolved a lot since the PR. There is still a need to get fast validators, but I think we can bake that into malli core functions - and enable that via options. Thanks anyway.

PR #311, opened 2020-12-02
Apply multi dispatch function only once on enter during transform

### LINKED ISSUE #309: Weird interaction between :multi, :map and encoders
Hi,

I am seeing some weird behavior around `:multi` but maybe I'm holding something wrong. I was able to reduce it to the following reproduction case. Consider the following definitions:

```clj
(require '[malli.core :as m]
         '[malli.transform :as mt])

(defn test-transformer []
  (mt/transformer
   {:name :test
    :encoders {:map {:leave #(assoc % :map-encoder-called true)}}}))

(def test-key-schema
  [:keyword {:encode/test {:enter (constantly :key-encoder-called)}}])
```

The following evaluates as I would expect:

```clj
(m/encode [:map
           [:type test-key-schema]]
          {:type :foo}
          test-transformer)

=> 

{:type :key-encoder-called, :map-encoder-called true}
```

However, when I wrap the same schema in a `:multi`, this happens:

```clj
(m/encode [:multi {:dispatch :type}
           [:foo [:map
                  [:type test-key-schema]]]]
          {:type :foo}
          test-transformer)

=>

{:type :key-encoder-called}
```

So the transformer's map encoder is not called anymore for some reason. Turns out that this is apparently caused by the map schema containing a key that matches the one used as the `:dispatch` function, because changing it to some other key makes it work again:

```clj
(m/encode [:multi {:dispatch :type}
           [:foo [:map
                  [:some-other-key test-key-schema]]]]
          {:type :foo
           :some-other-key :foo}
          test-transformer)

=>

{:type :foo,
 :some-other-key :key-encoder-called,
 :map-encoder-called true}
```

Also note that it doesn't matter whether there are any other keys next to the dispatch key:

```clj
(m/encode [:multi {:dispatch :type}
           [:foo [:map
                  [:type test-key-schema
                   :some-other-key test-key-schema]]]]
          {:type :foo
           :some-other-key :foo}
          test-transformer)

=>

{:type :key-encoder-called, :some-other-key :foo}
```

So am I doing / expecting something wrong here or is this a bug?

DerGuteMoritz:
Fixes #309.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -908,15 +908,19 @@
                    child-transformers (reduce-kv
                                         (fn [acc k s] (assoc acc k (-transformer s transformer method options)))
                                         {} dispatch-map)
-                   build (fn [phase]
+                   build (fn [phase ->dispatch]
                            (let [->this (phase this-transformer)
                                  ->children (->> child-transformers
                                                  (keep (fn [[k v]] (if-let [t (phase v)] [k t])))
                                                  (into {}))
-                                 ->child (if (seq ->children) (fn [x] (if-let [t (->children (dispatch x))] (t x) x)))]
-                             (-chain phase [->this ->child])))]
-               {:enter (build :enter)
-                :leave (build :leave)}))
+                                 ->child (if (seq ->children)
+                                           (fn [[x k]]
+                                             (if-let [t (->children k)]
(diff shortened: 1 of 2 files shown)


nilern:
#309 was fixed by #314, so closing this.

PR #312, opened 2020-12-03
Regex schemas

nilern:
Based on [Seqexp](https://github.com/cgrand/seqexp) but much optimized and extended.

WIP:
- [ ] transformers
- [ ] sufficient tests
- [ ] benchmarking
- [ ] (more) optimization
- [ ] cleanup
- [ ] docstrings
- [ ] documentation (README etc.)
- [ ] tree parsing (for non-regex schemas as well, e.g. `:or`)

PR #314, opened 2020-12-04
Rewrite `-transform`

### LINKED ISSUE #309: Weird interaction between :multi, :map and encoders

nilern:
Conventionalize traversal order; encoder/decoder `:enter` is called before descending into children and `:leave` after transforming children. Change `-transform` to return a nilable encoder/decoder function instead of an interceptor. Encoders and decoders can
still be interceptors; there are no changes to the `malli.transform` namespace.

* Simpler to understand and implement.
* (Potentially) faster.
    - In particular `:or` and #312 transformers can be implemented without hacks that do redundant transformations and intermediate allocations. The only way I could find to make #312 fit the current `-transform` would have exponential time (and space) complexity.
    - May also enable more no-op optimizations or at least they can be spotted more easily, YMMV.
* Caveat: the return type of `-transform` and the resulting traversal order are breaking changes. However I expect most users know almost nothing about these and do not rely on them.

Incidentally, fixes #309 :smile:


ikitommi (approved):
looks much simpler.

PR #316, opened 2020-12-09
malli.transform: add a guard against decoding unrepresentable ints (fixes #315)

### LINKED ISSUE #315: Perform a guard against Number.MAX_SAFE_INTEGER before using parseInt in the int? transformer on JS?
Today, Malli decoding of integers using the string transformer yield incorrect transformations above JS's `MAX_SAFE_INTEGER` because `parseInt` exhibits the same behaviour.

I would find it preferable to fail to parse above this threshold as though the input were invalid rather than a broken decode -- for obvious reason the below doesn't roundtrip successfully through a m/decode, m/encode pair.

With other encode/decode situations, running `(m/validate type decoded-value)` works to check whether the transform succeeded, but here the resultant value even passes validation, so there's really no way to know that the decode gave you an invalid result other than guarding /before/ decode.

```clojure
cljs.user> (malli.core/decode int? "11111111111111111" malli.transform/string-transformer)
11111111111111112
cljs.user> (js/parseInt "11111111111111111" 10)
11111111111111112
cljs.user> js/Number.MAX_SAFE_INTEGER
9007199254740991
cljs.user> (malli.core/validate int? 11111111111111112)
true
```

Would you be open to a PR to guard this inside the string transformer for integers on CLJS?

tekacs:
This branch makes it so that Malli won't try to decode strings in JS that don't have a representable value in JS.

Any number larger than `js/Number.MAX_SAFE_INTEGER` isn't reliably representable in JS, so whilst "9007199254740992" decodes to 9007199254740992 as you'd expect, "9007199254740993" (one greater) decodes to 9007199254740992 as well.

This results in a disparity in what can be decoded as an integer between JS and Java but given that JS can't represent certain numbers that's what we have to accept, I think. :)

There's more context on the fact that for example this is hard to detect in #315.

Hopefully this isn't a breaking change, since the previous behaviour could be considered 'incorrect' (the decodes resulted in unexpected values).

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -60,6 +60,7 @@
                  (cond
                    (js/isNaN x') x
                    (> x' js/Number.MAX_SAFE_INTEGER) x
+                   (< x' js/Number.MIN_SAFE_INTEGER) x
                    :else x')))
       (catch #?(:clj Exception, :cljs js/Error) _ x))
     x))
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -56,7 +56,11 @@
   (if (string? x)
     (try
       #?(:clj  (Long/parseLong x)
-         :cljs (let [x' (if (re-find #"\D" (subs x 1)) ##NaN (js/parseInt x 10))] (if (js/isNaN x') x x')))
+         :cljs (let [x' (if (re-find #"\D" (subs x 1)) ##NaN (js/parseInt x 10))]
+                 (cond
+                   (js/isNaN x') x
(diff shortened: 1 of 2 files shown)


tekacs:
The failing test run seems to be a transient one, in a failing pull of Guava due to an SSLException, rather than because of a failing test.

ikitommi:
I was wondering, should the `:int` schema be limited to js min & max so that any given `:int` works exactly the same in js & in jvm.

ikitommi:
and, re-run the tests, passed now. Thanks!

PR #322, opened 2020-12-21
mu/find

ikitommi:
Fix broken entry-finding convention, `:malli.core/find` instead of `malli.core/val`.

```clj
(mu/get [:map [:b {:optional true} int?]] [::m/find :b])
; => [:b {:optional true} int?]

(mu/find [:map [:b {:optional true} int?]] :b)
; => [:b {:optional true} int?]

(-> [:map [:a [:map [:b {:optional true} Int]]]]
    (mu/get :a)
    (mu/find :b))
; => [:b {:optional true} int?]

(-> [:map [:a [:map [:b {:optional true} Int]]]]
    (mu/get-in [:a [::m/find :b]]))
; => [:b {:optional true} int?]
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -135,10 +135,10 @@
   (-set-children schema (assoc (-children schema) key value)))
 
 (defn -get-entries [schema key default]
-  (or (if (and (vector? key) (= ::val (nth key 0)))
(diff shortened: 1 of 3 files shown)


RutledgePaulV:
Thank you!

PR #323, opened 2020-12-27
document regex string schemas

euccastro:


README.md
```diff
@@ -172,10 +172,35 @@ Using `:string` Schema:
 
 ```clj
 (m/validate :string "kikka")
-; => true
+;; => true
```
euccastro, line 175: Drive-by: other examples include double semicolons too.

ikitommi:
Thanks!

Issue #324, opened 2020-12-30
Better support for namespaced transformer names


DerGuteMoritz:
When a transformer's name is a namespaced keyword, it is currently not possible to include the namespace in schema-hinted transformations. For example, in order to add a schema-hint for the following transformer:

```clojure
(def my-transformer
  (mt/transformer {:name :my/kikka}
                  mt/string-transformer))
```

One has to omit the `my` namespace like this:

```clojure
(def my-schema
  [keyword? {:decode/kikka {:enter 'str/lower-case}}])

(m/decode my-schema "KUKKA" my-transformer) ; => :kukka
```

I feel it would be desirable to support namespaced transformer names in this situation, too, to prevent accidental name clashes.

In practice, it is possible to construct a keyword which contains both namespaces:

```clojure
(keyword (name :decode/my/kikka)) ; => :my/kikka
```

But I guess that this only works by accident. At least the [EDN spec](https://github.com/edn-format/edn#symbols) explicitly disallows it. Also, it's incompatible with auto-resolved keywords, so one would have to resort to hacks like:

```clojure
(keyword "encode" (str (symbol ::kikka)))
```

How about something like this instead:

```clojure
(def my-schema
  [keyword? {:decode {:my/kikka {:enter 'str/lower-case}}}])
```

The only drawback I can think of is that it's a little less convenient when merging property maps. Thoughts?

ikitommi:
Thanks for the PR. There is already similar method with JSON Schema & Swagger:

```clj
{:json-schema {:type "string"} ;; 1
{:json-schema/type "string"} ;; 2
```

as property merge uses a simple clojure merge, the last submap wins with 1, but using 2, all the unique properties are merged together. This is intented and by design. With transformations, not sure what is mental default for people, the submap approach looks nice, and I would assume many would use that as the default way to define things, causing potential issues later. But, as transformers are chained, this would still work:

```clj
[:map [:x {:decode {:string my-decode} [:string {:decode {:json my-decode}}]]]
```

what is your use case for needing qualified transformer names?

nilern:
`:decode/...` always looked suspicious to me. Shouldn't we just completely switch to using the submap?

Of course it is always possible to make do without qualified names, but is it ever truly necessary?

ikitommi:
let's not break the current api, at least before 1.0.0.

DerGuteMoritz:
[...]

PR #325, opened 2020-12-30
Support schema-hints for namespaced transformer names

DerGuteMoritz:
FWIW, here's an implementation of the suggestion I made in #324.


ikitommi:
this is really good, totally forgot to merge, merged via #619 Thanks!

Issue #241, opened 2020-08-10
Schema Parsing


ikitommi:
Currently, `m/explain` parses the values in align to the Schemas, producing `:in` and `:path` paths pointing paths in both values and schemas. This is a good base for generic parsing, e.g. Clojure Spec conform. We could reuse the `m/-explain` to collect paths for valid & invalid errors. The resulting *~conformed* data structure could be self-describing, so that no `m/-unexplain` would be needed to turn the parsed data structure back into value, like spec has `s/unform`. This could be achieved using special types/records to mark branching, instead of just maps and vectors (which can't be found from values without per-schema method).

From spec:

```clj
(s/def ::name-or-id (s/or :name string?
                          :id   int?))

(s/conform ::name-or-id "abc")
;;=> [:name "abc"]

(s/conform ::name-or-id 100)
;;=> [:id 100]
```

Would be ~:

```clj
(m/parse [:or string? int?] "abd")
;; => #Branch{:path 0, :value "abc"}

(m/parse [:or string? int?] 100)
;; => #Branch{:path 1, :value 100}
```

And with named branches:

```clj
(m/parse [:orn [:string string?] [:int int?]] "abd")
;; => #Branch{:path :string, :value "abc"}

(m/parse [:orn [:string string?] [:int int?]] 100)
;; => #Branch{:path :int, :value 100}
```

As a bonus, we could present errors in the parse tree in-place.

See also #180.

borkdude:
I would also like to see something like `s/conformer` in malli:

https://clojuredocs.org/clojure.spec.alpha/conformer

ikitommi:
ported the `conformer` example against the current api: validate + encode + validate + encode.

```clj
(def Schema1 [:string {:encode/conform #(mapv str (seq %))}])
(def Schema2 [:sequential {:encode/conform #(apply str %)} [:enum "0" "1"]])

(def conformer (mt/transformer {:name :conform}))

(as-> "1011" $
      (do (println $ "=>" (m/validate Schema1 $)) $)
      (m/encode Schema1 $ conformer)
      (do (println (pr-str $) "=>" (m/validate Schema2 $)) $)
      (m/encode Schema2 $ conformer)
      (= "1011" $))
; 1011 => true
; ["1" "0" "1" "1"] => true
; => true
```

nilern:
#312 has most of the required logic, but only for regex schemas. Destructuring other schemas should be much more straightforward though.

nilern:
[...]

PR #330, opened 2021-01-08
Schema parsing / conform

nilern:
Something like this for #241 and the rest of #180?

Builds on top of #317.

* On failure, `(m/conform s v)` returns `:malli.core/nonconforming`
* For most schemas `(m/conform s v)` just returns something that is `=` to `v` (optimally, `v` itself).
* But for the (new) `:or*` (tagged `:or`) `(= (m/conform [:or* [:s string?] [:n int?]] 0) (miu/-tagged :n 0))`
    - Non-identity conforming like this propagates: `(= (m/conform [:tuple [:or* [:s string?] [:n int?]]] [0]) [(miu/-tagged :n 0)])`
* Seqexes conform much as in Spec:

```clj
(m/conform [:alt string? int?] [0]) ; => 0
(m/conform [:alt* [:s string?] [:n int?]] [0]) ; => (miu/-tagged :n 0)

(m/conform [:cat [:* string?] [:* int?]] ["foo" "bar" 1 2]) ; => [["foo" "bar"] [1 2]]
(m/conform [:cat* [:s [:* string?]] [:n [:* int?]]] ["foo" "bar" 1 2]) ; => {:s ["foo" "bar"], :n [1 2]}

(m/conform [:? string?] ["foo"]) ; => "foo"
(m/conform [:? string?] []) ; => nil

(m/conform [:* string?] ["foo" "bar"]) ; => ["foo" "bar"]
(m/conform [:* string?] []) ; => []

(m/conform [:+ string?] ["foo" "bar"]) ; => ("foo" "bar")
(m/conform [:+ string?] []) ; => :malli.core/nonconforming

(m/conform [:repeat {:min 1, :max 3} string?] []) ; => :malli.core/nonconforming
(m/conform [:repeat {:min 1, :max 3} string?] ["foo" "bar"]) ; => ["foo" "bar"]
(m/conform [:repeat {:min 1, :max 3} string?] ["foo" "bar" "baz" "quux"]) ; => :malli.core/nonconforming

(m/conform [:cat* [:s [:schema [:* string?]]]] [["foo" "bar"]]) ; => {:s ["foo" "bar"]}
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1224,7 +1224,7 @@
           (-regex-transformer [_ transformer method options]
             (re-transformer properties (map #(-regex-transformer % transformer method options) children))))))))
 
-(defn -sequence-entry-schema [{:keys [type child-bounds re-validator re-explainer re-transformer] :as opts}]
+(defn -sequence-entry-schema [{:keys [type child-bounds re-validator re-explainer re-conformer re-transformer] :as opts}]
   ^{:type ::into-schema}
   (reify IntoSchema
     (-into-schema [_ properties children options]
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -391,7 +391,6 @@
           (-get [_ key default] (get children key default))
           (-set [this key value] (-set-assoc-children this key value)))))))
 
-;; TODO: [:or* [:b boolean?] [:s string?]]
 (defn -or-schema []
   ^{:type ::into-schema}
   (reify IntoSchema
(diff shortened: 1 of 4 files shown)


nilern:
At this point it should be quite simple to adjust the result format to taste.

nilern:
Also, `match`, `parse` or even `destructure` would be a more usual name, especially if it is not going to be as similar to Spec as it is now.

ikitommi:
This is really good. Few suggestions:

1. the conform should not throw, should fail with `::m/conforming` instead, have a code for this (on top of this PR), makes the backtracking order of magnitude faster
2. still think that we should use `explain` for `conform`, the code looks like 90% of c&p now. Can be done later
3. the `conform`, I suggest parse, can do a PR on top of this.

ikitommi (approved):

PR #332, opened 2021-01-11
Add default errors for comparator schemas

hansbugge:
This PR adds default error messages for comparator schemas.  Examples:
```clojure
(-> (m/explain [:> 4] 4)
    me/humanize)
;; => ["should be larger than 4"]

(-> (m/explain [:> 4] "")
    me/humanize)
;; => ["should be a number"]

(-> (m/explain [:= :foo] :bar)
    me/humanize)
;; => ["should be :foo"]

(-> (m/explain [:not= :foo] :foo)
    me/humanize)
;; => ["should not be :foo"]
```


ikitommi:
Thanks!

PR #333, opened 2021-01-11
Fix crash in humanize for maps

### LINKED ISSUE #321: malli.error/humanize throws ClassCastException when processing errors at different depths
Hi, and thanks for malli!

There seems to be a bug in humanize:

```clj
(require '[malli.core :as m])
(require '[malli.error :as me])

(def person
  [:and
   seq?
   [:map [:name string?]]])


(me/humanize (m/explain [:map [:person person]] {:person {}}))
; => error: java.lang.ClassCastException: class clojure.lang.Keyword cannot be cast to class java.lang.Number clojure.lang.Keyword is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')
```

The issue seems to be that humanize borks when dealing with errors at different levels of the value. In the above the seq? predicate fails the entire object, and the :map fails because the :name key is missing.

```clj
(m/explain [:map [:person person]] {:person {}})
; =>
{:schema [:map [:person [:and seq? [:map [:name string?]]]]],
 :value {:person {}},
 :errors
 ({:path [:person 0],
   :in [:person],
   :schema seq?,
   :value {},
   :type nil,
   :message nil}
  {:path [:person 1 :name],
   :in [:person :name],
   :schema [:map [:name string?]],
   :value nil,
   :type :malli.core/missing-key,
   :message nil})}
```

Also note that reversing the order of :and's children produces the following output.

```clj
(require '[malli.core :as m])
(require '[malli.error :as me])

(def person
  [:and
   [:map [:name string?]]
   seq?])


(me/humanize
 (m/explain
  [:map [:person person]]
  {:person {}}))
; =>
{:person [[:name ["missing required key"]]
          "should be a seq"]}
```

soulplant:
This change makes it so that humanize always "descends" into maps, so errors that used to reside "on" a map now go inside in the :malli/error key. The alternative would be to have top level errors go "on" a map, but then move them inside the map as the need arises, but it seemed bad to have the format change based on what other errors are found.

Fixes #321

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -132,8 +132,10 @@
 (defn- -assoc-in [acc value [p & ps] error]
   (cond
     p (let [acc' (-ensure (or acc (empty value)) p)
-            value' (if ps (-assoc-in (-get acc p) (-get value p) ps error) error)]
-        (-put acc' p value'))
+            error' (if (or ps (map? value))
+                     (-assoc-in (-get acc p) (-get value p) ps error)
+                     error)]
+        (-put acc' p error'))
     (map? value) (recur acc value [:malli/error] error)
     acc acc
     :else error))
--- a/test/malli/error_test.cljc
+++ b/test/malli/error_test.cljc
@@ -148,7 +148,20 @@
            (me/humanize
              {:value {},
               :errors [{:in [:x], :schema [:map [:x int?]], :type ::m/missing-key}
-                       {:in [:x], :schema [:map [:x int?]], :type ::m/missing-key}]})))))
+                       {:in [:x], :schema [:map [:x int?]], :type ::m/missing-key}]}))))
+
+  (testing "maps can have top level errors and key errors"
+    (is (= {:person {:malli/error ["should be a seq"],
+                     :name ["missing required key"]}}
+           (-> [:map [:person [:and seq? [:map [:name string?]]]]]
+               (m/explain {:person {}})
+               (me/humanize)))))
+
+  (testing "maps have errors inside"
+    (is (= {:person {:malli/error ["should be a seq"]}}
+           (-> [:map [:person seq?]]
+               (m/explain {:person {}})
+               (me/humanize))))))
 
 (deftest humanize-customization-test
   (let [schema [:map


src/malli/error.cljc
```diff
@@ -132,8 +132,10 @@
 (defn- -assoc-in [acc value [p & ps] error]
   (cond
     p (let [acc' (-ensure (or acc (empty value)) p)
-            value' (if ps (-assoc-in (-get acc p) (-get value p) ps error) error)]
-        (-put acc' p value'))
+            error' (if (or ps (map? value))
```
ikitommi, line 135: does this work correctly? doesn't it check just that the top-level value is a map? e.g. does this give correct errors:

```clj
(m/explain 
  [:vector [:map [:person [:and seq? [:map [:name string?]]]]]] 
  [{:person {}}])
```

soulplant: Thanks for the review.

This change does handle the case you suggest correctly[1] - we aren't checking the top-level value, in fact we are checking the bottom-level value, as (map? value) is only reached when we have no more elements in the error path to traverse into.

I'm not sure this is the desired solution to the problem though. This change has the effect of always putting errors inside maps and never on them, which seems to be inconsistent with the other collection types, e.g.:

```
;; For vectors and sets we place errors 'on' the collection.  
(-> map?
    (m/explain [])
    (me/humanize))
;; => ["should be a map"]

(-> map?
    (m/explain #{})
    (me/humanize))
;; => ["should be a map"]

;; But for maps we place errors 'in' the collection. 
(-> seq?
    (m/explain {})
    (me/humanize))
;; => #:malli{:error ["should be a seq"]}
```

However, the other collections have problems as well:

```
(-> [:and map? [:vector int?]]
    (m/explain [1 2 "a"])
    (me/humanize))
;; => ["should be a map" nil ["should be an int"]]

(-> [:and [:vector int?] map?]
    (m/explain [1 2 "a"])
    (me/humanize))
;; => [nil nil ["should be an int"]]

(-> [:and map? [:set int?]]
    (m/explain #{1 2 "a"})
    (me/humanize))
;; => "error: java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number (java.lang.String and java.lang.Number are in module java.base of loader 'bootstrap')"
```

In general what should the behaviour be for representing errors at both [:foo] and [:foo :bar] (or [:foo 0])? Maybe it is preferable to put errors on a map if there are no errors that are explicitly in the map? That has the unfortunate side effect of changing the shape of humanize's result depending on what errors are found in it, but maybe is the lesser of two evils here.

[1]
```
(me/humanize
 (m/explain 
  [:vector [:map [:person [:and seq? [:map [:name string?]]]]]] 
  [{:person {}}]))
;; =>
;; [{:person
;;   {:malli/error ["should be a seq"], :name ["missing required key"]}}]
```

ikitommi: What if:

* `humanize` would tag the error vectors with metadata `:malli.error/error`, so we would know if it's a vector of values or vector of errors

* in case a humanized path already contains an error vector,  we do not concat container (map with error behind key or vector with error on index) errors into it, just new top-level errors, ignore all others

```clj
(-> [:and map? [:vector int?]]
    (m/explain [1 2 "a"])
    (me/humanize))
;; => ["should be a map"]

(-> [:and seq? [:map [:x int?]]
    (m/explain {})
    (me/humanize))
;; => ["should be a seq"]
```

* if there is already a map container error in path, all top-level errors would accumulate under `:malli/error`.

```clj
(-> [:and [:vector int?] map?]
    (m/explain [1 2 "a"])
    (me/humanize))
;; => [nil nil ["should be int"]]

(-> [:and [:map [:x int?] seq?]
    (m/explain {})
    (me/humanize))
;; => {:a ["missing required key"], :malli/error ["should be a seq"]}
```

this would give consistent errors in all cases. Another option is always to put all top-level map-errors under `:malli.error`:

```clj
(-> [:and seq? [:map [:x int?]]
    (m/explain {})
    (me/humanize))
;; => {:malli/error ["should be a seq"]}
```

what do you think?

soulplant: Yeah, I think dropping errors in certain cases sounds good. It's a bit of a gotcha, I think, to have to check the metadata, but probably most of the time people can just ignore it until they start doing more fancy stuff. Or possibly they will use me/error-path and me/error-message if they need more precision.

It's a bit of a toss up in my mind as to whether to treat maps the same as everything else, or always put errors inside them. One thing you might consider is returning set errors in a map as well so we can tell which elements had which errors. In fact, that would work for sequences too - to have their errors returned in maps from index to error vector.

ikitommi: I think we should go with first approach: the first erros defines the form of the error:

```clj
(-> [:and seq? [:map [:x int?]]
    (m/explain [1 2 "a"])
    (me/humanize))
;; => ["should be a seq"]

(-> [:and [:map [:x int?] seq?]
    (m/explain {})
    (me/humanize))
;; => {:a ["missing required key"], :malli/error ["should be a seq"]}
```

would you like to finish the PR with that?

ikitommi: Do you still have time to finish this? Bitten by this too.

soulplant: Hey, sorry for the slow reply. This fell off my radar a bit.

I think probably it's good to land the patch as it is, as I believe it makes the code more correct and handles a particular edge case, but making the other changes may be a bit more work than I can commit to at the moment.

If anybody else, or yourself, would like to take this change over please feel free!

lsevero: I'm also using this fix as is with success. I think it could be merged as well.

ikitommi: good point, created #476 for the rest. Thanks for the PR, sorry for taking so long.

PR #335, opened 2021-01-15
Expose function to register a function schema.

ageneau:
Use case: I want to be able to store my function schemas in an edn file and to generate a clj-kondo config as part of my build process. Currently the only way to register a new function schema to the registry is using the "=>" macro which assumes the schema definition is in the same compilation unit as the function it specifies.
This patch exposes a new function  m/=>register-schema! which allows registering function schemas from a different compilation unit.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1509,11 +1509,14 @@
                (-fail! ::overlapping-arities {:arities arity->=>})) s)
        (-fail! :invalid-=>schema {:type (type s), :schema s})))))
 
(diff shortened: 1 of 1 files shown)


ikitommi:
Thanks! 

Renamed the function to conform the currrent (spartan?) naming convention:

```clj
-(defn =>register-schema! [ns name value]
+(defn -register-=>schema! [ns name value]
```

* public internal, starts with `-`  
* it does `register` 
* for `=>schema`
* side-effecting `!`

PR #336, opened 2021-01-16
Fix `m/walk` docstring

DerGuteMoritz:
Got thrown off track for a bit by the "properties" there :slightly_smiling_face:


ikitommi:
thanks!

PR #338, opened 2021-01-16
malli.util: add a rename-keys utility, similar to clojure.set

tekacs:
It looked to me that renaming keys isn't a completely trivial operation for a user to implement themselves (without a series of `dissoc`/`assoc` pairs that would be pretty inefficient), so I added in a utility for performing it using `transform-entries`.

This... may not be the best implementation possible, so if there's a better way that's preferred I can do that or it can just be swapped out.

Also I may be missing an existing simple solution to this. :)


ikitommi:
Looks good, thanks!

PR #340, opened 2021-01-17
malli.util: use passed options in into-schema in transform-entries

tekacs:
It's... not clear if this is in the spirit of consistency within malli, so despite having opened this I'm not sure of it myself now. :)

Without this change, the following code throws an error:

```clojure
(def reg (mr/composite-registry {:a/x 'int?} (m/default-schemas)))

(def sch [:map [:a/x {:a/b 1} [:a/x {}]]])

(defn key-transform [entries]
  (map (fn [[k m _]] [k m k]) entries))

(mu/transform-entries sch key-transform {:registry reg})
;; Execution error (ExceptionInfo) at malli.core/-fail! (core.cljc:93).
;; :malli.core/invalid-schema {:schema :a/x}
```

With this change it returns the following result:

```clojure
[:map [:a/x #:a{:b 1} :a/x]]
```

With the latter change (in 7674d48f1fcce947a6ca38a3570126ac0e298d6b), it also works to do this:

```clojure
(def sch-with-attached-options (m/schema sch {:registry reg}))

(mu/transform-entries sch-with-attached-options key-transform nil)
```

Tests for both cases are attached.

One currently-working way to solve this might be:

```clojure
(defn key-transform-alt [entries]
  (map (fn [[k m v]] [k m (m/schema k (m/options v))]) entries))
```

Transformations as they stand don't receive the options of the top-level type (I believe?), so whilst in this case I can steal the options from a value, it's not always possible (such as if transforming starting from an empty map, e.g. `(m/schema :map {:registry ...})`) -- sometimes the transform would need to be constructed with `(m/options top-level-schema)` out of band, something like:

```clojure
(defn key-transform-manual [options]
  (fn [entries]
    (map (fn [[k m v]] [k m (m/schema k options)]) entries)))

(mu/transform-entries sch-with-attached-options (key-transform-manual (m/options sch-with-attached-options)) nil)
; or
(mu/transform-entries sch (key-transform-manual options) nil)
```

... or I could be missing a much easier solution.

The goal here is that transformation functions should be able to be written once by using the context of the outer type at the last step when interpreting the type.

Another possible solution would be to have `transform-entries` pass `options` to the transformation function, although this would be a breaking change for all consumers of `transform-entries`, which is public and documented.


ikitommi:
Thanks! And sorry for taking so long.

PR #342, opened 2021-01-20
Extend `:qualified-keyword` predicate with `{:namespace <keyword>}` properties

yanlobkarev:
* Test against `[:qualified-keyword {:namespace <keyword>}]` schema
* Add schema example in `README.md`


ikitommi:
thanks!

PR #347, opened 2021-01-25
Fix interceptor composition

### LINKED ISSUE #346: Interceptor chains composition is suspect
`transform.cljc`:
```clj
...
(let [enter (if (and enter new-enter) #(new-enter (enter %)) (or enter new-enter))
     leave (if (and leave new-leave) #(new-leave (leave %)) (or leave new-leave))]
...
```
Seems that either `#(new-enter (enter %))` or `#(new-leave (leave %))` should be reversed. Probably the latter.

nilern:
Fixes #346.


miikka (approved):
Looks correct, although maybe there should be a new test.

PR #354, opened 2021-01-26
Set JSON Schema types and formats for numbers properly

Deraen:
- Note that number formats aren't mentioned in JSON Schema spec, but are
  from OpenAPI spec.
- Comparators shouldn't automatically set format to double.
- Set format double for :double and double?.
- decimal? is BigDecimal, so that is not a double.
- Fix rational? type to number, double isn't valid type
- Fix ratio? type to number, ratios aren't integers.

### DIFF
--- a/src/malli/json_schema.cljc
+++ b/src/malli/json_schema.cljc
@@ -22,17 +22,21 @@
 
 (defmulti accept (fn [name _schema _children _options] name) :default ::default)
 
+;; Note, format value for integer/number is from OpenAPI spec.
+;; JSON Schema spesifies formats for strings like uuid and date-time, but
+;; not numbers, but custom values are allowed.
+
 (defmethod accept ::default [_ _ _ _] {})
 (defmethod accept 'any? [_ _ _ _] {})
 (defmethod accept 'some? [_ _ _ _] {})
 (defmethod accept 'number? [_ _ _ _] {:type "number" :format "double"})
-(defmethod accept 'integer? [_ _ _ _] {:type "integer"})
+(defmethod accept 'integer? [_ _ _ _] {:type "integer" :format "int32"})
 (defmethod accept 'int? [_ _ _ _] {:type "integer" :format "int64"})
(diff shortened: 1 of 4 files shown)


src/malli/json_schema.cljc
```diff
@@ -105,7 +105,8 @@
   (merge {:type "integer"} (-> schema m/properties (select-keys [:min :max]) (set/rename-keys {:min :minimum, :max :maximum}))))
 
 (defmethod accept :double [_ schema _ _]
-  (merge {:type "number"} (-> schema m/properties (select-keys [:min :max]) (set/rename-keys {:min :minimum, :max :maximum}))))
+  (merge {:type "number"}
+         (-> schema m/properties (select-keys [:min :max]) (set/rename-keys {:min :minimum, :max :maximum}))))
```
ikitommi, line 109: thks could be a helper in `json-schema` as it's now copied 3 times already.

src/malli/swagger.cljc
```diff
@@ -22,6 +21,25 @@
...
+(defmethod accept 'int? [_ _ _ _] {:type "integer" :format "int64"})
+(defmethod accept 'pos-int? [_ _ _ _] {:type "integer", :format "int64", :minimum 1})
+(defmethod accept 'neg-int? [_ _ _ _] {:type "integer", :format "int64", :maximum -1})
+(defmethod accept 'nat-int? [_ _ _ _] {:type "integer", :format "int64" :minimum 0})
+(defmethod accept 'float? [_ _ _ _] {:type "number" :format "float"})
+(defmethod accept 'double? [_ _ _ _] {:type "number" :format "double"})
+
+(defmethod accept :int [_ schema _ _]
+  (merge {:type "integer" :format "int64"}
+         (-> schema m/properties (select-keys [:min :max]) (set/rename-keys {:min :minimum, :max :maximum}))))
```
ikitommi, line 37: Just double-check: is this inclusive or exclusive? Json schema separates the two.

ikitommi (approved):

src/malli/swagger.cljc
Deraen: Inclusive.

PR #355, opened 2021-01-26
Add test for nested multi schema with decode/string

Deraen:
Why does this fail?

### DIFF
--- a/test/malli/core_test.cljc
+++ b/test/malli/core_test.cljc
@@ -835,6 +835,36 @@
               [:human [:map [:type 'keyword?] [:name 'string?] [:address [:map [:country 'keyword?]]]]]]
              (m/form schema)))))
 
+  (testing "nested multi schema"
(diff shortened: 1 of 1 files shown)


ikitommi:
dispatch values are keywords (`:int`), but in values, they are strings (`"int"`). Try `[:multi {:dispatch :type} ["int" ...]]`

Deraen:
That's why I'm using :decode/string.

It has worked in some other cases.

Deraen:
The previous test cases also test `decode/string` to update value before dispatch.

But doesn't work with nested multi schemas.

Deraen:
Oh, I need to of course decode values first. Seems to work here, something else broken on the real project.

Deraen:
Okay, Malli itself works.

Reitit Malli coercion closes all the schemas, and that breaks `[:and multi [:map ...]]` case because maps on the both and branches will be closed.

PR #357, opened 2021-01-27
Adds unparsing

nilern:
To complement #330.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1,7 +1,7 @@
 (ns malli.core
   (:refer-clojure :exclude [eval type -deref deref -lookup -key])
   (:require [malli.sci :as ms]
-            [malli.impl.util :as miu]
+            [malli.impl.util :as miu :refer [-invalid?]]
             [malli.impl.regex :as re]
             [malli.registry :as mr])
   #?(:clj (:import (java.util.regex Pattern)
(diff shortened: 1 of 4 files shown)


src/malli/core.cljc
```diff
@@ -830,6 +905,20 @@
                                          (identical? v* v) x
                                          :else (assoc x i v*))))
                                    x parsers)))))
+          (-unparser [_]
```
ikitommi, line 908: the `-unparser` code are basically 95% of copy-paste from `-parser`. we could share code in the future.

src/malli/impl/regex.cljc
```diff
@@ -289,7 +345,7 @@
                      (-park-validator! driver
                                        (fn [driver stack pos coll k]
                                          (compulsories driver (conj (pop stack) (inc (peek stack))) pos coll k))
-                                       regs pos coll k))) ; TCO
+                                       regs pos coll k)))   ; TCO
```
ikitommi, line 348: whitespace

ikitommi (approved):
Looks good.

src/malli/core.cljc
nilern: I made a sketch https://github.com/metosin/malli/compare/master...nilern:coercer a while back about coercion (validation + transform like in Plumatic Schema) and I believe that approach could also yield `parse`, `unparse`, `encode` and `decode` as special cases... but it gets complicated.

ikitommi: Having one generic solution would be great - if that emits code that is not (much) slower than the current solution. Currently, for example. `-validate´ tries to be close to optimal clojure code having no extra args or function calls. JVM inlining works well with small and simple methods.

nilern: There shouldn't be such issues if it is the code that constructs the parsers, unparsers etc. that is shared instead of constructing one coercer function that implements everything by branching on extra arguments or something.

Issue #334, opened 2021-01-15
Generator size should be larger for more variety


sunnygunnar:
This is not necessarily a bug, but it's unexpected behavior so I'm bringing it up FWIW :)

As we see from the following:

https://github.com/metosin/malli/blob/19fb277f39bcd86d585ecbd6b7c280acaa343e6a/src/malli/generator.cljc#L190-L195

`malli.generator/generate` defaults to `size` = 1, whereas `clojure.test.check.generators/generate` defaults to `size` = 30:

https://github.com/clojure/test.check/blob/d0e179720fbb29c97345f04f2e76fceff956dcc8/src/main/clojure/clojure/test/check/generators.cljc#L182-L194

I came across this issue when trying to generate some `juxt/tick` instants and was only seeing dates around 1970 even when scaling up the jitter to be in days instead of milliseconds. When `size` is 30 just like `test.check`, I see a wide variety of dates. Might be worth being consistent with `test.check` and defaulting to `size` = 30.

ikitommi:
Can't recall why it defaults to 1. 30 seems a good way to go. PR welcome.

sunnygunnar:
Sounds good! PR at https://github.com/metosin/malli/pull/364 @ikitommi.

PR #364, opened 2021-02-04
Ensure we use size 30 for generator (for more variety)

sunnygunnar:
This is in line with `clojure.test.check.generators/generate`. See also https://github.com/metosin/malli/issues/334.


ikitommi:
Thanks!

Issue #363, opened 2021-01-29
Resolve minimum size for sequence-type schemas


ikitommi:
this is needed to figure out the minimum arity of function schemas with varargs. Could be a new Protocol like:

```clj
(defprotocol SequenceSchema
  (-min-size [this] "returns minimum size of thee sequence"))
```

implemented to `:tuple`, `:vector`, ... `:cat` ....

**EDIT** I think we can limit this to regex/sequence schemas. So:

```clj
(require '[malli.core :as m])

(m/minimum-size [:? int?]) ;=> 0
(m/minimum-size [:cat [:? string?] [:+ int?] [:repeat {:min 1} boolean?]) ;=> 2
(m/minimum-size [:alt [:+ int?] [:? boolean?]]) ;; => 0
```

miikka:
A couple of examples would be nice to illustrate what is wanted here.

ikitommi:
Added examples. Also, this is needed for #360.

miikka:
Solved by #365.

PR #365, opened 2021-02-05
Get minimum size for sequence schemas

ilmoraunio:


miikka:
Partial solution for https://github.com/metosin/malli/issues/363

nilern:
I don't think we actually want the public wrapper and documentation, now this only works for seqex schemas and only exists for function schemas...

ikitommi (approved):
elegantly compact.

PR #376, opened 2021-02-22
Regex clarifications

nilern:
Simplifications to `regex.impl`:

* Streamline `:cat` and `:alt` "compilations"
* Improve naming


ikitommi:
great

PR #377, opened 2021-02-22
Fix #353: Remove redundant `satisfies?` call

### LINKED ISSUE #353: cljs: Cannot infer target type in expression (. this -malli$core$protocol$)
Got this warning in shadow-cljs with a local malli checkout. It seems safe to put `(satisfies? protocol this)` into the `:clj` guard since its containing function is never called from cljs, but not sure what the intent is.
```
--------------------------------------------------------------------------------
 253 |                          (extend-protocol Schemas (class this)
 254 |                            (-schema? [_] s?)
 255 |                            (-into-schema? [_] is?))))
 256 |                (satisfies? protocol this))]
----------------------^---------------------------------------------------------
 Cannot infer target type in expression (. this -malli$core$protocol$)
--------------------------------------------------------------------------------
 257 |   (extend-protocol Schemas
 258 |     nil
 259 |     (-schema? [_] false)
 260 |     (-into-schema? [_] false)
--------------------------------------------------------------------------------
```

nilern:
Fixes #353.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -302,8 +302,7 @@
                              is? (satisfies? IntoSchema this)]
                          (extend-protocol Schemas (class this)
                            (-schema? [_] s?)
-                           (-into-schema? [_] is?))))
-               (satisfies? protocol this))]
+                           (-into-schema? [_] is?)))))]
   (extend-protocol Schemas
     nil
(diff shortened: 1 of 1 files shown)


ikitommi (approved):

ikitommi:
thx!

PR #378, opened 2021-02-28
renamed named variants of :cat, :alt and :or

ikitommi:
* `:cat*`, `:alt*`, `:or*` => `:cat-named`, `:alt-named`, `:or-named`


ikitommi:
not 100% happy with either one.

borkdude:
@ikitommi Maybe:

```
:catn, :altn, :orn 
``` 
?

ikitommi:
@borkdude that's more concise. Was one of the original thoughts, but got buried. I think i's the best of the bunch.

ikitommi:
changed to that. thanks!

PR #379, opened 2021-03-01
Fix clj-kondo ret schema

mynomoto:
It got broken on
https://github.com/metosin/malli/commit/0a22ccb1036c25b12f4b8c0d0f1dc20682aa6ca1#diff-0c47d0cd013f6cb845efdab593e740088b1cbcdb0704959422641cdffd9b7b01R162


ikitommi:
thanks!

PR #388, opened 2021-03-08
Fix mu/update for when key doesn't exist in schema

DerGuteMoritz:
Passing an empty map schema by default doesn't really make sense here. Probably a copy & paste leftover from `update-in`? :-)


ikitommi:
Good catch, thanks!

PR #391, opened 2021-03-11
:m/default for :multi

ikitommi:
Default branch with `:m/default`:

```clj
(def valid?
  (m/validator
    [:multi {:dispatch :type}
     ["object" [:map-of :keyword :string]]
     [:m/default :string]]))

(valid? {:type "object", :key "1", :value "100"})
; => true

(valid? "SUCCESS!")
; => true

(valid? :failure)
; => false
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1087,7 +1087,7 @@
              form (-create-form type properties forms)
              dispatch (eval (:dispatch properties) options)
              dispatch-map (->> (for [[k s] entries] [k s]) (into {}))
-             finder (fn [{:keys [:m/default] :as m}] (fn [x] (m x default)))]
+             finder (fn [{:keys [::default] :as m}] (fn [x] (m x default)))]
          (when-not dispatch
            (miu/-fail! ::missing-property {:key :dispatch}))
          ^{:type ::schema}
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1086,7 +1086,8 @@
              {:keys [children entries forms]} (-parse-entries children opts' options)
              form (-create-form type properties forms)
              dispatch (eval (:dispatch properties) options)
-             dispatch-map (->> (for [[k s] entries] [k s]) (into {}))]
+             dispatch-map (->> (for [[k s] entries] [k s]) (into {}))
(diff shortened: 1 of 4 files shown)


src/malli/core.cljc
```diff
@@ -1087,7 +1087,8 @@
              {:keys [children entries forms]} (-parse-entries children opts' options)
              form (-create-form type properties forms)
              dispatch (eval (:dispatch properties) options)
-             dispatch-map (->> (for [[k s] entries] [k s]) (into {}))]
+             dispatch-map (->> (for [[k s] entries] [k s]) (into {}))
+             finder (fn [{:keys [:m/default] :as m}] (fn [x] (m x default)))]
```
ilmoraunio, line 1091: [aside] TIL you can destructure & reuse keys like that.

test/malli/core_test.cljc
```diff
@@ -981,7 +981,34 @@
               {:dispatch :type, :decode/string '(fn [x] (update x :type keyword))}
               [:sized [:map [:type 'keyword?] [:size 'int?]]]
               [:human [:map [:type 'keyword?] [:name 'string?] [:address [:map [:country 'keyword?]]]]]]
-             (m/form schema)))))
+             (m/form schema))))
+
+    (let [schema [:multi {:dispatch first}
```
ilmoraunio, line 986: [nit] Could test that dispatching `:m/default` works both with `{:dispatch :type}` and `{:dispatch first}`.

ilmoraunio (approved):
Looks good!

DerGuteMoritz:
I vote for using `:malli.core/default` / `::m/default`.

ilmoraunio:
Good catch. Looks like [grouped keywords](https://blog.jeaye.com/2017/10/31/clojure-keywords/) can be too opinionating especially going forward with such a highly common prefix like `m`.

DerGuteMoritz:
Exactly. Also, there even is some prior use of that in Malli itself already with the various options for `m/walk` :slightly_smiling_face:

ikitommi:
`::m/default` it is then. for humanized errors, the top-level key is `:malli/error`, which should be changed too, but a breaking change in the public api we can postpone. thanks for the comments.

PR #397, opened 2021-03-18
Ignore optional properties in `default-value-transformer`

### LINKED ISSUE #351: Ignore optional schemas when generating default values
default-value-transformer is handy, but I want it to ignore schemas with an `{:optional true}` property in the outputted value. I made a pass at writing a variant that does so:
```clojure
(defn default-value-optional-transformer
  ([]
   (default-value-optional-transformer nil))
  ([{:keys [key defaults] :or {key :default}}]
   (let [get-default  (fn [schema] (let [properties (m/properties schema)]
                                   (when-not (:optional properties)
                                     (if-some [default (get properties key)]
                                       default
                                       (some->> schema m/type (get defaults) (#(% schema)))))))
         set-default  {:compile (fn [schema _]
                                  (when-some [default (get-default schema)]
                                    (fn [x] (if (nil? x) default x))))}
         add-defaults {:compile (fn [schema _]
                                  (let [defaults (into {}
                                                       (keep (fn [[k {default key optional :optional} v]]
                                                               (when-not optional
                                                                 (when-some [default (if (some? default) default (get-default v))]
                                                                   [k default]))))
                                                       (m/children schema))]
                                    (when (seq defaults)
                                      (fn [x]
                                        (if (map? x)
                                          (reduce-kv
                                           (fn [acc k v]
                                             (if (contains? x k)
                                               acc
                                               (assoc acc k v)))
                                           x defaults)
                                          x)))))}]
     (transformer
      {:default-decoder set-default
       :default-encoder set-default}
      {:decoders {:map add-defaults}
       :encoders {:map add-defaults}}))))
```

Not sure if this is behavior is better as a separate fn or an option in default-value-transformer.

nilern:
Closes #351

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -381,22 +381,21 @@
    (default-value-transformer nil))
   ([{:keys [key defaults] :or {key :default}}]
    (let [get-default (fn [schema]
-                       (let [default (some-> schema m/properties key)]
-                         (if (some? default)
-                           default
-                           (some->> schema m/type (get defaults) (#(% schema))))))
+                       (if-some [default (some-> schema m/properties key)]
+                         default
+                         (some->> schema m/type (get defaults) (#(% schema)))))
          set-default {:compile (fn [schema _]
-                                 (if-some [default (get-default schema)]
+                                 (when-some [default (get-default schema)]
                                    (fn [x] (if (nil? x) default x))))}
          add-defaults {:compile (fn [schema _]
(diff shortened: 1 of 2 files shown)


nilern:
Seems that tests are broken by https://github.com/metosin/malli/commit/0ad965a5ec5a866b06ae2412e3de166cbfaab8ac, not the commits in this PR.

ikitommi (approved):

Issue #402, opened 2021-03-22
parse on :multi should tag on branch


ikitommi:
should work like `:orn`:

```clj
(def Multi
  [:multi {:dispatch :type}
   [:user [:map [:size :int]]]
   [::m/default :any]])

(m/parse Multi {:type :user, :size 1})
; => [:user {:type :user, :size 1}]

(m/parse Multi {:type "sized", :size 1})
; => [:malli.core/default {:type "sized", :size 1}]

(->> {:type :user, :size 1}
     (m/parse Multi)
     (m/unparse Multi))
; => {:type :user, :size 1}
```

currently:

```clj
(m/parse Multi {:type :user, :size 1})
; => {:type :user, :size 1}
```

ikitommi:
closed by #403

PR #403, opened 2021-03-22
parse & unparse on :multi

ikitommi:
```clj
(def Multi
  [:multi {:dispatch :type}
   [:user [:map [:size :int]]]
   [::m/default :any]])

(m/parse Multi {:type :user, :size 1})
; => [:user {:type :user, :size 1}]

(m/parse Multi {:type "sized", :size 1})
; => [:malli.core/default {:type "sized", :size 1}]

(->> {:type :user, :size 1}
     (m/parse Multi)
     (m/unparse Multi))
; => {:type :user, :size 1}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1301,7 +1301,7 @@ Schemas can be used to parse values using `m/parse` and `m/parser`:
 ;               :children [[:primitive [:text "Hello, world of data"]]]}]]}]
 ```
 
-Parsing returns tagged values for `:orn`, `:catn`, `:altn` and `:multi`. For others it returns the value.
+Parsing returns tagged values for `:orn`, `:catn`, `:altn` and `:multi`.
 
 ```clj
 (def Multi
--- a/README.md
+++ b/README.md
@@ -1301,6 +1301,21 @@ Schemas can be used to parse values using `m/parse` and `m/parser`:
 ;               :children [[:primitive [:text "Hello, world of data"]]]}]]}]
 ```
 
+Parsing returns tagged values for `:orn`, `:catn`, `:altn` and `:multi`. For others it returns the value.
+
+```clj
+(def Multi
+  [:multi {:dispatch :type}
+   [:user [:map [:size :int]]]
+   [::m/default :any]])
+
+(m/parse Multi {:type :user, :size 1})
+; => [:user {:type :user, :size 1}]
+
+(m/parse Multi {:type "sized", :size 1})
+; => [:malli.core/default {:type "sized", :size 1}]
+```
+
 ## Unparsing values
 
 The inverse of parsing, using `m/unparse` and `m/unparser`:
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1098,10 +1098,7 @@
            (-type-properties [_] (:type-properties opts'))
            (-validator [_]
              (let [find (finder (reduce-kv (fn [acc k s] (assoc acc k (-validator s))) {} dispatch-map))]
-               (fn [x]
-                 (if-let [validator (find (dispatch x))]
-                   (validator x)
-                   false))))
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1301,6 +1301,21 @@ Schemas can be used to parse values using `m/parse` and `m/parser`:
 ;               :children [[:primitive [:text "Hello, world of data"]]]}]]}]
 ```
 
+Parsing returns tagged values for `:orn`, `:catn`, `:altn` and `:multi`. For others it returns the value.
```
nilern, line 1304: The value is also not returned unchanged if there are e.g. nested `:orn`:s: `[:tuple [:orn [:foo ...` => `[(miu/-tagged :foo ...`

src/malli/core.cljc
```diff
@@ -1110,19 +1107,14 @@
                    (explainer x in acc)
                    (conj acc (miu/-error (->path path) (->path in) this x ::invalid-dispatch-value))))))
            (-parser [_]
-             (let [find (finder (reduce-kv (fn [acc k s] (assoc acc k (-parser s))) {} dispatch-map))]
-               (fn [x]
-                 (if-some [parser (find (dispatch x))]
-                   (parser x)
-                   ::invalid))))
+             (let [parse (fn [k s] (as-> (-parser s) $ (fn [x] (miu/-map-valid #(miu/-tagged k %) ($ x)))))
```
nilern, line 1110: `as->` has no advantage over `let` here.

src/malli/core.cljc
```diff
@@ -1110,19 +1107,14 @@
...
-             (let [unparsers (mapv (fn [[_ _ s]] (-unparser s)) children)]
-               (fn [x]
-                 ;; Can't use `dispatch` as `x` might not be valid before it has been unparsed:
-                 (reduce (fn [_ unparser] (miu/-map-valid reduced (unparser x)))
-                         ::invalid unparsers))))
+             (let [unparsers (reduce-kv (fn [acc k s] (assoc acc k (-unparser s))) {} dispatch-map)]
+               (fn [x] (if (miu/-tagged? x) (if-some [f (unparsers (key x))] (f (val x)) ::invalid) ::invalid))))
            (-transformer [this transformer method options]
-             ;; FIXME: Probably should not use `dispatch`, see comment in `-unparser` above.
+             ;; FIXME: Probably should not use `dispatch`
```
nilern, line 1117: Add the deleted comment here instead of just removing the reference as well.

test/malli/core_test.cljc
```diff
@@ -944,14 +944,14 @@
...
+      (is (= (miu/-tagged :sized valid1) (m/parse schema valid1)))
+      (is (= (miu/-tagged :human valid2) (m/parse schema valid2)))
       (is (= ::m/invalid (m/parse schema invalid1)))
       (is (= ::m/invalid (m/parse schema invalid2)))
       (is (= ::m/invalid (m/parse schema invalid3)))
       (is (= ::m/invalid (m/parse schema "not-a-map")))
-      (is (= valid1 (m/unparse schema valid1)))
-      (is (= valid2 (m/unparse schema valid2)))
+      (is (= valid1 (m/unparse schema (m/parse schema valid1))))
+      (is (= valid2 (m/unparse schema (m/parse schema valid2))))
```
nilern, line 954: I would prefer just `(m/unparse schema (miu/-tagged :sized valid1))`, I get a bad vibe from using the SUT when creating the test inputs (even if it is a different part, tested separately).

nilern (changes requested):
The README needs to be fixed at least.

nilern:
Let's not forget that this fixes #402

README.md
ikitommi: good catch.

ikitommi:
thanks for the review! fixed all but the test.

PR #429, opened 2021-04-27
Fix #427

### LINKED ISSUE #427: Generated sets of `:ref` are always empty
Sets that contains a `:ref` are always generated empty, for example:

```clojure
(malli.gen/sample [:set
                   {:registry {::foo :int}}
                   [:ref ::foo]])
```

No such problem with `:vector`. I tried a couple other ways in my (more complex) use case but it always boils down to this fact.

helins:
Attention, second commit introduces a small breaking change (fix typo in function name) and can be discarded if it is intolerable. Important commit is the first one which fixes #427.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -50,7 +50,7 @@
   (let [{:keys [min max]} (-min-max schema options)
         [continue options] (-recur schema options)
         child (-> schema m/children first)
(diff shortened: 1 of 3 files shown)


ikitommi:
Thanks! It's a private var, so rename doesn't break any contracts.

PR #430, opened 2021-04-27
Do not ignore double properties at generation

helins:
Previous to this commit, properties of `:double` such as `:NaN?` were
discarded for generation meaning the user could not apply such
parameters.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -215,12 +215,8 @@
 (defmethod -schema-generator :double [schema options]
   (gen/double* (merge (let [props (m/properties schema
                                                 options)]
-                        {:infinite? (or (:gen/infinite? props)
-                                        (:infinite? props)
-                                        false)
-                         :NaN?      (or (:gen/NaN? props)
-                                        (:NaN? props)
-                                        false)})
+                        {:infinite? (get props :gen/infinite? false)
+                         :NaN?      (get props :gen/NaN? false)})
                       (-min-max schema options))))
 (defmethod -schema-generator :boolean [_ _] gen/boolean)
 (defmethod -schema-generator :keyword [_ _] gen/keyword)
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -32,8 +32,6 @@
       (miu/-fail! ::invalid-property {:key :gen/min, :value gen-min, :min min}))
     (when (and max gen-max (> gen-max max))
       (miu/-fail! ::invalid-property {:key :gen/max, :value gen-min, :max min}))
-           :min (or gen-min min)
-           :max (or gen-max max)
     {:min (or gen-min min)
      :max (or gen-max max)}))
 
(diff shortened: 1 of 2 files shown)


src/malli/generator.cljc
```diff
@@ -32,6 +32,8 @@
       (miu/-fail! ::invalid-property {:key :gen/min, :value gen-min, :min min}))
     (when (and max gen-max (> gen-max max))
       (miu/-fail! ::invalid-property {:key :gen/max, :value gen-min, :max min}))
+           :min (or gen-min min)
```
ikitommi, line 35: dead code?

test/malli/generator_test.cljc
```diff
@@ -40,6 +40,17 @@
...
       (is (every? (partial m/validate schema) (mg/sample schema {:size 1000})))))
 
+  (testing "double properties"
+    (is (pos? (count (filter (fn [x]
+                               (or (#?(:clj Double/isNaN
+                                       :cljs js/isNaN)
+                                    x)
+                                   (#{##Inf ##-Inf} x)))
+                             (mg/sample [:double
+                                         {:infinite? true
```
ikitommi, line 50: instead of reserving new plain keywords for generation, I suggest that these keys are namespaced with `gen`, e.g. `:gen/infinite?` and we could have an utility fn `-gen-properties` in `malli.generator` that filters out all keywords that start with `gen` and remove the ns, e.g. `[:double {:gen/infinite? true, :gen/kikka "kukka"}]` would emit `{:infinite? true, :kikka "kukka"}`.

ikitommi (changes requested):
looks good but let's make it better.

test/malli/generator_test.cljc
helins: `:gen/infinite?` and `:gen/NaN?` defaulting to their corresponding unqualified keywords or ultimately to `false` (current behavior) is a good idea as it matches the behavior in `-min-max`.

However, is there a rationale for having a general purpose utility that extends this to arbitrary properties? Isn't it best to let generator preparing functions pick what they need? (No firm opinion on the matter, just thinking about performance and avoiding unnecessary looping, albeit small)

ikitommi: both ways are ok as:

- haven't seen any other need for generic `:gen` properties -> custom thing here is ok
- so the looping would happen only when a generator is created, so the actual generator would still be fast

btw, there are also `-double-gen`, which is used with `:>` etc.

helins: All right, I will favor the first option to keep it simple (no over-engineering unless needed :) )

I didn't think too much about other gen utils that rely on `-double-gen` because those schemas do not target doubles specifically (eg. `:>` validates all numbers). The fact they generate some looks more like an implementation detail.

src/malli/generator.cljc
helins: My bad, indeed!

src/malli/generator.cljc
```diff
@@ -214,9 +212,16 @@
...
 (defmethod -schema-generator :string [schema options] (-string-gen schema options))
 (defmethod -schema-generator :int [schema options] (gen/large-integer* (-min-max schema options)))
-(defmethod -schema-generator :double [schema options] (gen/double* (merge {:infinite? false, :NaN? false}
-                                                                          (m/properties schema options)
-                                                                          (-min-max schema options))))
+(defmethod -schema-generator :double [schema options]
+  (gen/double* (merge (let [props (m/properties schema
+                                                options)]
+                        {:infinite? (or (:gen/infinite? props)
+                                        (:infinite? props)
```
ikitommi, line 219: this reserves 2 new plain keyword keys, let's just support the namespaced versions? If  `:double` uses the plain keys, we could use them here. Currently, it doesn't.  Same with `:NaN?`

Otherwise, looks good.

helins: Right, got slightly confused by a previous comment but your are right, this implies that the plain kws have an effect on validation as well. It might be useful some day to put that kind of restriction in `:double` (no Nan or Inf) but that's another topic.

ikitommi:
Thanks!

Issue #435, opened 2021-05-01
`:re` ignores `:gen/xxx` properties in absence of `:gen/gen`


helins:
Something like this will throw `:test-chuck-not-available` on generation:

```clojure
[:re
 {:gen/fmap (fn [x] ...)
  :gen/schema :int}
  #"some-regex"]
```

I guess in Clojurescript it will throw even sooner because it will look for the `:re` method on `-schema-generator` which is only defined on the JVM.

Adding a `:gen/gen` value, even completely unrelated, fixes this. So the problem is broader. It looks like this is due to how `malli.generator/-create`, in absence `:gen/gen`, calls `-schema-generator` even when it is not actually needed.

If you confirm this, I can try to PR a rewrite of `-create`.

ikitommi:
Please do.

ikitommi:
fixed in master

PR #436, opened 2021-05-01
Refactor creating generators

helins:
Starting from #435 , this PR solves that kind of problems by refactoring `malli.generator/-create`.


helins:
Weird, tests pass on my machine, something must be out of sync.

ikitommi:
Refactoring looks really good, thanks!

PR #440, opened 2021-05-05
DRAFT: First class protocol & value class-based value inferring

### LINKED ISSUE #191: malli.providers/provide takes "forever" for large samples
Hi there,

I'm using Malli `0.0.1-SNAPSHOT`.

I have 15 samples of approximately 200KB each (when saved as EDN to disk). Malli is unable to provide a schema for those - it seems to run forever. It would probably have stopped at some point, but I lost my patience.

I tried to run the provider for just one of those, it actually finished, but it took several minutes on my i7 w/6 physical cores (don't know if you have any parallelization yet).

Would it be possible in some way to speed this up, or at least provide some peek into the progress (via logging, async or similar)?

Looks promising though, will use Malli for smaller stuff for now.

Best regards,
Henrik

ikitommi:
fixes #191.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -26,6 +26,7 @@
   (-type-properties [this] "returns schema type properties")
   (-properties-schema [this options] "maybe returns :map schema describing schema properties")
   (-children-schema [this options] "maybe returns sequence schema describing schema children")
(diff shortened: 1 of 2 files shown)


ikitommi:
#191 is closed already, closing this

PR #454, opened 2021-05-20
Correct example in README

jeroenvandijk:
The partial function is now used correctly in the example. The example will fail as the `s/valid?` only expects 2 arguments.

Another option is to remove the partial application of `s/valid?`
```clojure
valid? (partial s/valid? spec)
```

I don't know if this change has any effect on the mentioned benchmarks. I'm guessing the partial application is not faster.


ikitommi:
thanks!

PR #458, opened 2021-05-28
Faster map schema validator

### LINKED ISSUE #455: Possibly faster map schemas on the JVM
Hello,
I've been poking at the `-map-schema` implementation and looking for ways to make it go slightly faster and be even more amenable to the JIT compiler.
Some ideas I managed to come up with which may be worth looking into:
- `-validator` checks for `map?` first, which guarantees the object is `IPersistentMap` and `Associative`. This lets us cut down on the overhead of `find` by dispatching directly to `.entryAt()`. For  this schema `(schema [:map [:a int?] [:b int?] [:c int?]])` I measured a 2x improvement just from this single change
- another option to look at which plays into the JIT is the option of composing validators instead of iterating over them, something like `(reduce (fn [agg f] (fn [m] (and (agg m) (f m)))) validators)`. From preliminary tests it looks like it gives a speed up of at least ~20%, more for larger maps

What do you think? Does it seem viable?

bsless:
Replace find with .entryAt in clj map validator

Since the map validator always checks for `map?` first, we have a guarantee the object is `IPersistentMap` which is an `Associative`, thus has an `.entryAt` method.
Calling it instead of using `find` is significantly faster.

The accompanying gist specifies the methods used to test the change and the results
https://gist.github.com/bsless/c50ee40f9734d6fe59e6dc36930d78c0

Closes #455

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -651,7 +651,8 @@
                                         (fn [[key {:keys [optional]} value]]
                                           (let [valid? (-validator value)
                                                 default (boolean optional)]
-                                            (fn [m] (if-let [map-entry (find m key)] (valid? (val map-entry)) default))))
+                                            #?(:clj (fn [^clojure.lang.Associative m] (if-let [map-entry (.entryAt m key)] (valid? (val map-entry)) default))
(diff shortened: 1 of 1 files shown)


ikitommi:
🙇

PR #460, opened 2021-05-30
Unroll validators iteration

bsless:
Schemas for and, or, orn and map combine several validators

Instead of closing over the collection with a function and iterating
over it at invoke time, we can unroll it into a single function
by utilizing a little code generation macros

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -385,7 +385,10 @@
           Schema
           (-validator [_]
             (let [validators (distinct (map -validator children))]
-              (if (second validators) (apply every-pred validators) (first validators))))
+              #?(:clj
+                 (miu/-every-pred validators)
+                 :cljs
+                 (if (second validators) (apply every-pred validators) (first validators)))))
           (-explainer [_ path]
             (let [explainers (mapv (fn [[i c]] (-explainer c (conj path i))) (map-indexed vector children))]
               (fn explain [x in acc] (reduce (fn [acc' explainer] (explainer x in acc')) acc explainers))))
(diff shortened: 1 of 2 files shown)


bsless:
Results for the following test case:
```clojure
(def xs [1 2 3 5 7 11 13 17 23])

(defn divisible-by?
  [^long n ^long m]
  (if (zero? (rem n m)) true false))

(def divisible-by
  (m/-simple-schema
   (fn [_ [child]] {:type :rem, :pred #(divisible-by? % child), :min 1, :max 1})))

(def reg (mr/registry (merge (m/default-schemas) {:rem divisible-by})))

(def s (m/schema (into [:and] (map (fn [n] [:rem n])) xs) {:registry reg}))
(def v (m/validator s {:registry reg}))
(def n (reduce * xs))
(cc/quick-bench (v n))
```
Before:
```
Evaluation count : 2757552 in 6 samples of 459592 calls.
             Execution time mean : 212.621220 ns
    Execution time std-deviation : 0.419191 ns
   Execution time lower quantile : 212.195284 ns ( 2.5%)
   Execution time upper quantile : 213.114226 ns (97.5%)
                   Overhead used : 2.066616 ns
```
After:
```
Evaluation count : 39869724 in 6 samples of 6644954 calls.
             Execution time mean : 13.043325 ns
    Execution time std-deviation : 0.234305 ns
   Execution time lower quantile : 12.775744 ns ( 2.5%)
   Execution time upper quantile : 13.289920 ns (97.5%)
                   Overhead used : 2.066616 ns
```

ikitommi:
tested this with simpler data sets, performance gains are massive. Curious, is the code inlined from some other library or did you create it from scratch? Anyway, big thanks!

ikitommi:
added tests, CHANGELOG and reformatted some code: https://github.com/metosin/malli/commit/8631e2c9a665bb8e01344f110fd2f883ecec09c7

bsless:
I think I did it from scratch. Once I learned to recognize this pattern it turned out to be ubiquitous and very easy to recreate.
It's just the same loop-unrolling and closing over which can be done in so many cases.
Another implementation but slightly less JIT friendly is reducing with a combiner over all the predicates. I guess for complicated combinations which will be hard to write with macros it's still a preferable case.

bsless:
IIRC #458 should increase the speedups for maps. Could be JIT friendlier?

ikitommi:
Will add `find` to list of avoid-these-if-possible...

bsless:
I wish I knew how to explain why `RT/find` introduces so much overhead although it does less type checking than `RT/get`

PR #462, opened 2021-06-03
Humanized errors can be looked from parent schemas

### LINKED ISSUE #86: Feature request or bug: adding properties directly to map keys.

ikitommi:
using a custom `:resolve` impl with `me/humanize`:

```clj
(-> [:map
     [:foo {:error/message "entry-failure"} :int]]
    (m/explain {:foo "1"})
    (me/humanize {:resolve me/resolve-root-error}))
; => {:foo ["entry-failure"]}
```

fixes #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,11 +535,11 @@ Errors can be targeted using `:error/path` property:
 ; {:password2 ["passwords don't match"]}
 ```
 
-By default, only erronous schema properties are used:
+By default, only direct erronous schema properties are used:
 
 ```clj
 (-> [:map
-     [:foo {:error/message "entry-failure"} :int]]
+     [:foo {:error/message "entry-failure"} :int]] ;; here, :int fails, no error props
     (m/explain {:foo "1"})
     (me/humanize))
 ; => {:foo ["should be an integer"]}
--- a/README.md
+++ b/README.md
@@ -535,6 +535,26 @@ Errors can be targeted using `:error/path` property:
 ; {:password2 ["passwords don't match"]}
 ```
 
+By default, only erronous schema properties are used:
+
+```clj
+(-> [:map
+     [:foo {:error/message "entry-failure"} :int]]
+    (m/explain {:foo "1"})
+    (me/humanize))
+; => {:foo ["should be an integer"]}
+```
+
+Looking up humanized errors from parent schemas with custom `:resolve`:
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -535,6 +535,26 @@ Errors can be targeted using `:error/path` property:
 ; {:password2 ["passwords don't match"]}
 ```
 
+By default, only direct erronous schema properties are used:
```
viesti, line 538:
```suggestion
By default, only direct erroneous schema properties are used:
```

src/malli/error.cljc
```diff
@@ -267,13 +283,21 @@
                    $))))))))
 
 (defn humanize
+  "Humanized a explanation. Accepts the following optitons:
+
+  - `:wrap`, a function of `error -> message`, defaulting ot `:message`
```
viesti, line 288:
```suggestion
  - `:wrap`, a function of `error -> message`, defaulting to `:message`
```

ikitommi:
thanks!

src/malli/error.cljc
```diff
@@ -134,7 +136,7 @@
 (defn- -put [x k v]
   (cond
     (set? x) (conj x v)
-    (associative? x) (update x k (fn [e] (if (-just-error? v) (into (vec e) v) v)))
+    (associative? x) (update x k (fn [e] (if (-just-error? v) (vec (distinct (into (vec e) v))) v)))
```
ikitommi, line 139: this is slow, but thinking we'll change this to (ordered?) set later anyways.

PR #468, opened 2021-06-21
malli.plantuml

ikitommi:
```clj
(require '[malli.plantuml])

(malli.plantuml/transform
 [:schema
  {:registry {"Country" [:map
                         [:name [:enum :FI :PO]]
                         [:neighbors [:vector [:ref "Country"]]]]
              "Burger" [:map
                        [:name string?]
                        [:description {:optional true} string?]
                        [:origin [:maybe "Country"]]
                        [:price pos-int?]]
              "OrderLine" [:map
                           [:burger "Burger"]
                           [:amount int?]]
              "Order" [:map
                       [:lines [:vector "OrderLine"]]
                       [:delivery [:map
                                   [:delivered boolean?]
                                   [:address [:map
                                              [:street string?]
                                              [:zip int?]
                                              [:country "Country"]]]]]]}}
  "Order"])
 ;"@startuml
; entity \"Burger\" {
;  :name string?
;  :description string?
;  :origin [:maybe \\\"Country\\\"]
;  :price pos-int?
; }
; 
; entity \"Country\" {
;  :name [:enum :FI :PO]
;  :neighbors [:vector [:ref \\\"Country\\\"]]
; }
; 
; entity \"Order\" {
;  :lines [:vector \\\"OrderLine\\\"]
;  :delivery Order$Delivery
; }
; 
; abstract \"Order$Delivery\" {
;  :delivered boolean?
;  :address Order$Delivery$Address
; }
; 
; abstract \"Order$Delivery$Address\" {
;  :street string?
;  :zip int?
;  :country Country
; }
; 
; entity \"OrderLine\" {
;  :burger Burger
;  :amount int?
; }
; 
; \"Burger\" <|-- \"Country\"
; \"Country\" <|-- \"Country\"
; \"Order\" <|-- \"OrderLine\"
; \"Order\" *-- \"Order$Delivery\"
; \"Order$Delivery\" *-- \"Order$Delivery$Address\"
; \"Order$Delivery$Address\" <|-- \"Country\"
; \"OrderLine\" <|-- \"Burger\"
; @enduml
; "
```  

malli


brdloush (commented):
Should there really be a `<|--` (class inheritence) at line 21?

It results in following UML diagram. 

image

- Burger "class" is not supposed to be a subclass of OrderLine
- Country is definitely not supposed to be a subclass of Burger :-)

I tried to locally change line 21 to directed association arrow (and I also changed the direction of that arrow)
```
         (println from (if (entity? to) "-->" "*--") to))
```

And it generates following diagram which I believe might be more correct than the original one:

image
 
Not it reads as "OrderLine references Burger", "Burger references Country" etc.

brdloush:
Btw it would be also nice if the association lines had cardinalities (`0..1` / `1` / `0..*`) on the "arrow side". But it seems that `md/-get-links` only provides the from-to entity names (such as `["OrderLine" #{"Burger"}]
`), not the cardinality. Or is there any easy way of finding out the cardinality of the relation between entities?

ikitommi:
oh, it was supposed to be aggregation, not inheritance. fixed.

cardinalities, should be doable. Created #469 for it. help welcome.

ikitommi:
malli-fix

PR #471, opened 2021-07-04
malli.instrument

ikitommi:
# TODO

* [x] extract `m/-instrument` from generators
* [x] new `malli.instrument` ns with options to instrument the existing function registry
* [x] list all changes IN CHANGELOG
* [ ] ~~more tests?~~

## what

* `mallli.core/-instrument` for advanced users
* `malli.instrument/instrument!` and `malli.instrument/unstrument!` for normal users
* `malli.instrument/collect!` for peeling schemas from var metadata
* `malli.dev/start!` and `malli.dev/stop!` for the lazy (also emit the [clj-kondo](https://github.com/clj-kondo/clj-kondo) annotations for static type checking)
* `malli.generate/check` and `malli.generate/check!` for function value & `defn` checking

https://user-images.githubusercontent.com/567532/124499366-8afc8480-ddc6-11eb-9c74-37d740edf8d7.mov


pithyless:
I'd like to define whether certain functions are instrumented based on env (e.g. only dev/test or always). Do I understand correctly I can use the `:filters` option to pass a custom function? Would something like this be appropriate?

```
(m/=> kikka [:=> {:custom/check :strict} [:cat :int] [:int {:max 16}]])

(defn filter-tags [tags]
  (fn [_n _s d]
    (contains? tags (:custom/check (m/properties d)))))

(mi/instrument! {:filters [(filter-tags #{:strict})]})
```

If so, what is a good name for `:custom/check`? Otherwise, looking for insight on how to best approach this problem.

ikitommi:
@pithyless , that is one way to do that. The plumatic style defn's use var-level meta `:always-validate`, `:never-validate`.

Just added two extra helpers: `mi/-filter-schema` and `mi/-filter-var`.

Plan is to add more batteries into `malli.instrument`. One would be a function to read all public Vars from a given set of namespaces and extract & register the function schemas from those. There could be multiple extractors:

1. using defined qualified keys, e.g. `:malli/schema`
2. inferring the schemas using `tools-analyzer`, `core-typed` or `:arglists` meta (best-effort), ping @frenchy64 
3. from functions specs

for 1, could look like this:

```clj
(defn kukka
  "kukka"
  {:malli/schema [:=> [:cat :int] [:int {:max 6}]]
   :malli/scope #{:input :output}
   :user/mode #{:dev :prod}} ;;??
  [x] (* x x))
```

about naming / known keys, not sure what would be good. Still evolving. Ideas welcome.

pithyless:
I think your example makes a lot more sense:

```
(defn kukka
  "kukka"
  {:malli/schema [:=> [:cat :int] [:int {:max 6}]]
   :malli/scope #{:input :output}
   :user/mode #{:dev :prod}} ;;??
  [x] (* x x))
```

It's something I'd want anyway, but no need to come up with some custom `defn` syntax, which is always nice for tooling integration. And the `:malli/scope`, `:user/mode` stuff is actually related more to the var, then the schema! 👍 

Now just to have someone come up with a better name for `:user/mode`. I think it may be more appropriate to have `:malli.instrument/scope` (since it's related to the behavior of the instrumentation) and then question is if you would add some custom `app.malli.instrument/mode`?

ikitommi:
poke on `miu/collect!`:

Screenshot 2021-07-06 at 15 12 05

PR #479, opened 2021-07-19
Fix pred-composer

bsless:
Combiner should combine the results of applying composer

### DIFF
--- a/src/malli/impl/util.cljc
+++ b/src/malli/impl/util.cljc
@@ -43,7 +43,7 @@
            f (gensym "f__")
            cases (mapcat (fn [i] [i `(-combine-n ~c ~i ~preds)]) (range 2 (inc n)))
            else `(let [p# (~f (take ~n ~preds)) q# (~f (drop ~n ~preds))]
-                   (fn [x#] (and (p# x#) (q# x#))))]
+                   (fn [x#] (~c (p# x#) (q# x#))))]
        `(fn ~f [~preds]
           (case (count ~preds)
             0 (constantly true)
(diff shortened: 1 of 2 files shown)


ikitommi:
thx!

mikeananev:
I found errors during this cycle:
edn1 -> encode-value -> json -> decode-value -> edn2

In v 0.5.1, edn1 = edn2. All uuids and zoned-date-time encoded/decoded correctly, they are data of the same type.

In v  0.6.0-20210721.071739-2, edn1 != edn2
broken data for uuid and zoned-date-time - they are strings after decoding.

Can't reproduce this error on a simple schema. I've more complicated schemas, but cannot publish them here.
But in my code I do something like that.

`(do (def custom-json-transformer (mt/transformer mt/string-transformer mt/json-transformer))
    (def my-spec1 [:map
                   [:a :zoned-date-time]
                   [:b :string]
                   [:c :uuid]
                   [:d :int]])
    (def my-spec [:map [:s1 my-spec1] [:s2 my-spec1]])
    (def edn-value (mg/generate my-spec))
    (def encoded-value (m/encode my-spec edn-value custom-json-transformer))
    (def json-string-value (json/write-value-as-string encoded-value json/keyword-keys-object-mapper))
    (def encoded-value2 (json/read-value json-string-value json/keyword-keys-object-mapper))
    (def edn-value2 (m/decode my-spec encoded-value2 custom-json-transformer))
    (is (m/validate my-spec edn-value2)))`

bsless:
@mikeananev if you can't publish it publicly I wouldn't mind to take the communication to a private channel

PR #484, opened 2021-07-26
Transform longs bigger than js/Number.MAX_SAFE_INTEGER

nivekuil:
In cljs these are goog.math.Longs, malli transforms should do the same instead of leaving them as strings.

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -1,6 +1,7 @@
 (ns malli.transform
   #?(:cljs (:refer-clojure :exclude [Inst Keyword UUID]))
   (:require #?@(:cljs [[goog.date.UtcDateTime]
+                       [goog.math.Long]
                        [goog.date.Date]])
             [malli.core :as m])
   #?(:clj (:import (java.util Date UUID)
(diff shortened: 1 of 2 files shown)


ikitommi:
Could you add a test for this? For documentation and to avoid regression. Thanks.

nivekuil:
not sure why this is true:
```clojure
#?(:cljs (=  (goog.math.Long/fromString "9007199254740993") (malli.transform/-string->long "9007199254740993")))
```
but ./bin/node fails
```
Testing malli.transform-test

FAIL in (string->long) (/home/nivekuil/code/github/malli/test/malli/transform_test.cljc:34:7)
expected: (= (goog.math.Long/fromString "9007199254740993") (mt/-string->long "9007199254740993"))
  actual: (not (= #object[Long 9007199254740993] #object[Long 9007199254740993]))

FAIL in (string->long) (/home/nivekuil/code/github/malli/test/malli/transform_test.cljc:36:7)
expected: (= (goog.math.Long/fromString "-9007199254740993") (mt/-string->long "-9007199254740993"))
  actual: (not (= #object[Long -9007199254740993] #object[Long -9007199254740993]))

Testing malli.util-test

Ran 121 tests containing 2743 assertions.
2 failures, 0 errors.
```

ikitommi:
If we return `goog.math.Longs`, it will cause issues on js-interop, e.g. can't serialize that to JSON via `JSON.stringify` (writes a map with some internal pointers) etc. This is more right in some sense, but will cause issues elsewhere. Thanks for the PR, not going to merge this now.

If I missed something, comments welcome.

PR #485, opened 2021-07-26
Add arities to -comp and move to util ns

bsless:
The new implementation does not iterate over the functions at call time, preventing the performance fall-off for more than three functions

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -1,6 +1,7 @@
 (ns malli.clj-kondo
   (:require #?(:clj [clojure.java.io :as io])
-            [malli.core :as m]))
+            [malli.core :as m]
+            [malli.impl.util :as miu]))
 
 (declare transform)
 
@@ -68,7 +69,7 @@
 
 (defmethod accept ::m/val [_ _ children _] (first children))
 (defmethod accept :map [_ _ children _]
-  (let [{req true opt false} (->> children (group-by (m/-comp not :optional second)))
+  (let [{req true opt false} (->> children (group-by (miu/-comp not :optional second)))
         opt (apply array-map (mapcat (fn [[k _ s]] [k s]) opt))
         req (apply array-map (mapcat (fn [[k _ s]] [k s]) req))]
     (cond-> {:op :keys}, (seq opt) (assoc :opt opt), (seq req) (assoc :req req))))
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -150,15 +150,6 @@
 
 (defn -boolean-fn [x] (cond (boolean? x) (constantly x) (ifn? x) x :else (constantly false)))
 
-(defn -comp
-  ([] identity)
-  ([f] f)
-  ([f g] (fn [x] (f (g x))))
-  ([f g h] (fn [x] (f (g (h x)))))
-  ([f1 f2 f3 & fs]
-   (let [fs (reverse (list* f1 f2 f3 fs))]
-     #(loop [ret ((first fs) %), fs (next fs)] (if fs (recur ((first fs) ret) (next fs)) ret)))))
-
 (defn -update [m k f] (assoc m k (f (get m k))))
 
 (defn -memoize [f]
@@ -199,7 +190,7 @@
                                              (not (sequential? e)) (if (and naked-keys (-reference? e)) [[e nil e] e] (-fail! ::invalid-ref {:ref e}))
                                              (and (= 1 (count e)) (-reference? (first e))) (if naked-keys [[(first e) nil (first e)] e])
                                              (and (= 2 (count e)) (-reference? (first e)) (map? (last e))) (if naked-keys [(conj e (first e)) e])
-                                             :else [e (->> (-update (vec e) (dec (count e)) (-comp -form #(schema % options))) (keep identity) (vec))])
+                                             :else [e (->> (-update (vec e) (dec (count e)) (miu/-comp -form #(schema % options))) (keep identity) (vec))])
                              [p ?s] (if (or (nil? ?p) (map? ?p)) [?p ?v] [nil ?p])
                              s (cond-> (or ?s (if (-reference? k) f)) lazy-refs (-lazy options))
                              c [k p (schema s options)]]
@@ -217,12 +208,12 @@
 
 (defn -intercepting
   ([interceptor] (-intercepting interceptor nil))
-  ([{:keys [enter leave]} f] (some->> [leave f enter] (keep identity) (seq) (apply -comp))))
+  ([{:keys [enter leave]} f] (some->> [leave f enter] (keep identity) (seq) (apply miu/-comp))))
 
 (defn -parent-children-transformer [parent children transformer method options]
   (let [parent-transformer (-value-transformer transformer parent method options)
         child-transformers (into [] (keep #(-transformer % transformer method options)) children)
-        child-transformer (if (seq child-transformers) (apply -comp (rseq child-transformers)))]
+        child-transformer (if (seq child-transformers) (apply miu/-comp (rseq child-transformers)))]
     (-intercepting parent-transformer child-transformer)))
 
 (defn- -properties-and-children [[x :as xs]]
--- a/src/malli/dot.cljc
+++ b/src/malli/dot.cljc
@@ -1,6 +1,7 @@
 (ns malli.dot
   (:require [malli.core :as m]
             [malli.util :as mu]
+            [malli.impl.util :as miu]
             [malli.registry :as mr]
             [clojure.string :as str]))
 
@@ -21,7 +22,7 @@
     @state))
 
 (defn -schema-name [base path]
-  (->> path (remove #{:malli.core/in}) (map (m/-comp str/capitalize m/-keyword->string)) (into [base]) (str/join "$")))
+  (->> path (remove #{:malli.core/in}) (map (miu/-comp str/capitalize m/-keyword->string)) (into [base]) (str/join "$")))
 
 (defn -normalize [{:keys [registry] :as ctx}]
   (let [registry* (atom registry)]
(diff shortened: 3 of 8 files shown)


bsless:
Will leave usage of new `-comp` for further performance gains for other MRs. Keeping it small.

ikitommi:
Few comments:

1. blowing the stack - at what point the stack is too high so that this approach doesn't work anymore?

2. would love to see a chain of 10-20 big functions being composed with this. At some point, the we go over the JVM inlining budget and the impl could slow down. Compared to looping of small, surely compiled code blocks. I really don't know what happens, would like to learn.

3. cljs-bundle size. Malli has few simplified functions to keep the js-code small. Multi-arity fns generate a lot of code, so cljs-side of this could be simpler. Quick test on the generated js with Klipse:

```clj
cljs.user._comp = (function cljs$user$_comp(var_args){
  var G__33 = arguments.length;
  switch (G__33) {
    case (0):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$0();

      break;
    case (1):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$1((arguments[(0)]));

      break;
    case (2):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$2((arguments[(0)]),(arguments[(1)]));

      break;
    case (3):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$3((arguments[(0)]),(arguments[(1)]),(arguments[(2)]));

      break;
    case (4):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$4((arguments[(0)]),(arguments[(1)]),(arguments[(2)]),(arguments[(3)]));

      break;
    case (5):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$5((arguments[(0)]),(arguments[(1)]),(arguments[(2)]),(arguments[(3)]),(arguments[(4)]));

      break;
    case (6):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$6((arguments[(0)]),(arguments[(1)]),(arguments[(2)]),(arguments[(3)]),(arguments[(4)]),(arguments[(5)]));

      break;
    case (7):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$7((arguments[(0)]),(arguments[(1)]),(arguments[(2)]),(arguments[(3)]),(arguments[(4)]),(arguments[(5)]),(arguments[(6)]));

      break;
    case (8):
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$8((arguments[(0)]),(arguments[(1)]),(arguments[(2)]),(arguments[(3)]),(arguments[(4)]),(arguments[(5)]),(arguments[(6)]),(arguments[(7)]));

      break;
    default:
      var args_arr__25510__auto__ = [];
      var len__25493__auto___35 = arguments.length;
      var i__25494__auto___36 = (0);
      while(true){
        if((i__25494__auto___36 < len__25493__auto___35)){
          args_arr__25510__auto__.push((arguments[i__25494__auto___36]));

          var G__37 = (i__25494__auto___36 + (1));
          i__25494__auto___36 = G__37;
          continue;
        } else {
        }
        break;
      }

      var argseq__25511__auto__ = (new cljs.core.IndexedSeq(args_arr__25510__auto__.slice((8)),(0),null));
      return cljs.user._comp.cljs$core$IFn$_invoke$arity$variadic((arguments[(0)]),(arguments[(1)]),(arguments[(2)]),(arguments[(3)]),(arguments[(4)]),(arguments[(5)]),(arguments[(6)]),(arguments[(7)]),argseq__25511__auto__);

  }
});

(cljs.user._comp.cljs$core$IFn$_invoke$arity$0 = (function (){
  return cljs.core.identity;
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$1 = (function (f){
  return f;
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$2 = (function (f,g){
  return (function (x){
    return f.call(null,g.call(null,x));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$3 = (function (f,g,h){
  return (function (x){
    return f.call(null,g.call(null,h.call(null,x)));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$4 = (function (f1,f2,f3,f4){
  return (function (x){
    return f1.call(null,f2.call(null,f3.call(null,f4.call(null,x))));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$5 = (function (f1,f2,f3,f4,f5){
  return (function (x){
    return f1.call(null,f2.call(null,f3.call(null,f4.call(null,f5.call(null,x)))));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$6 = (function (f1,f2,f3,f4,f5,f6){
  return (function (x){
    return f1.call(null,f2.call(null,f3.call(null,f4.call(null,f5.call(null,f6.call(null,x))))));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$7 = (function (f1,f2,f3,f4,f5,f6,f7){
  return (function (x){
    return f1.call(null,f2.call(null,f3.call(null,f4.call(null,f5.call(null,f6.call(null,f7.call(null,x)))))));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$8 = (function (f1,f2,f3,f4,f5,f6,f7,f8){
  return (function (x){
    return f1.call(null,f2.call(null,f3.call(null,f4.call(null,f5.call(null,f6.call(null,f7.call(null,f8.call(null,x))))))));
  });
}));

(cljs.user._comp.cljs$core$IFn$_invoke$arity$variadic = (function (f1,f2,f3,f4,f5,f6,f7,f8,fs){
  return cljs.user._comp.call(null,cljs.core.apply.call(null,cljs.user._comp,fs),(function (x){
    return f1.call(null,f2.call(null,f3.call(null,f4.call(null,f5.call(null,f6.call(null,f7.call(null,f8.call(null,x))))))));
  }));
}));

/** @this {Function} */
(cljs.user._comp.cljs$lang$applyTo = (function (seq24){
  var G__25 = cljs.core.first.call(null,seq24);
  var seq24__$1 = cljs.core.next.call(null,seq24);
  var G__26 = cljs.core.first.call(null,seq24__$1);
  var seq24__$2 = cljs.core.next.call(null,seq24__$1);
  var G__27 = cljs.core.first.call(null,seq24__$2);
  var seq24__$3 = cljs.core.next.call(null,seq24__$2);
  var G__28 = cljs.core.first.call(null,seq24__$3);
  var seq24__$4 = cljs.core.next.call(null,seq24__$3);
  var G__29 = cljs.core.first.call(null,seq24__$4);
  var seq24__$5 = cljs.core.next.call(null,seq24__$4);
  var G__30 = cljs.core.first.call(null,seq24__$5);
  var seq24__$6 = cljs.core.next.call(null,seq24__$5);
  var G__31 = cljs.core.first.call(null,seq24__$6);
  var seq24__$7 = cljs.core.next.call(null,seq24__$6);
  var G__32 = cljs.core.first.call(null,seq24__$7);
  var seq24__$8 = cljs.core.next.call(null,seq24__$7);
  var self__25484__auto__ = this;
  return self__25484__auto__.cljs$core$IFn$_invoke$arity$variadic(G__25,G__26,G__27,G__28,G__29,G__30,G__31,G__32,seq24__$8);
}));

(cljs.user._comp.cljs$lang$maxFixedArity = (8));
```

bsless:
This sent me on a very interesting exploration.
Regarding stack depth, I tried `-comp`ing 1024 functions and didn't get a stack overflow, and that's without "partitioning" the elements evenly.
I then played around with [JITWatch](https://github.com/AdoptOpenJDK/jitwatch) and analyzed some JIT logs.
I think big functions won't get inlined anyway. The threshold for inlining is pretty low.
I also noticed there's a depth limit for inlining

A small repro:
```clojure
(ns malli.main
  (:require
   [malli.impl.util :refer [-comp]])
  (:gen-class))

(defn -main
  []
  (let [f (apply -comp (map (fn [f] (let [i (rand-int 100000)] (fn [x] (f i x)))) (repeat 1024 +)))]
    (dotimes [_ 1e6] (f 1))
    (println (f 1))
    (dotimes [_ 1e6] (f 1))
    (println (f 1))
    (dotimes [_ 1e6] (f 1))
    (println (f 1))
    ))
```
then
```sh
clojure -X:project/uberjar :main-class malli.main
java -XX:+PrintCompilation -XX:+UnlockDiagnosticVMOptions -XX:+TraceClassLoading  -XX:+LogCompilation  -XX:LogFile=mylogfile.log -XX:+DebugNonSafepoints -XX:+PrintAssembly -Dclojure.compiler.direct-linking=true -XX:+FlightRecorder -XX:StartFlightRecording=duration=60s,filename=myrecording.jfr -jar uber.jar  > /dev/null
```
Creates a log file which can be analyzed and a JFR recording, but I don't have much experience with it
It looks like the JVM can handle it okay and I think the closed over methods get inlined, unless I severely misunderstood the logs.

I can also tweak the maximum width in `-comp` and can spread the inputs evenly to reduce the total depth.

Regarding CLJS I'm not sure regarding the best method. We can default to regular `comp`

What do you think?

bowbahdoe:
@bsless I understand that the command you gave will replicate the results you got, but can you link to any educational resources you used to learn said command?

Specifically, if this library continues to pick up steam "automated performance regression tests" would be sensible, but I don't think its that common to have the ability to.

bsless:
@bowbahdoe The main resource relevant to my results here is JITwatch which is linked and explains the flags I used, but the very short TL;DR is the JVM's JIT compiler is a multi-tiered process which optimizes bytecode to assembly. These flags instruct the JVM to dump all the data relating to that process and I later analyzed it with JITwatch.

I think my findings indicate the compiler, at least on java 15, was able to handle both wide and deep function compositions.

If you know how to analyze JFR logs with VisualVM or CMS I would appreciate instructions on the subject, because I don't and they're pretty opaque.

Performance regression tests are possible, but I would say need to be clearly defined.

How I'd do it:
- define test cases
- implement using JMH
- run before/after
- detect regression/improvement per case
- run per JVM version
- can also run with different GCs

This is a huge test matrix, but if you want to conduct performance regression testing, I don't see another way.

ikitommi:
Thanks for digging into this. Good to know it doesn't blow the stack. Few comments:

1. cljs-size, we can limit it to 3 unrolled args
2. I consider `-comp` being part of the extender api, under `impl` it just internal, why was it moved? there are already similar helpers in `malli.core`. We should move them all or nothing, not few. Let's keep it like this, but will most likely move it back to core before releasing 0.6.0, unless good arguments for the move
3.  `-reverse` is not used?

bsless:
Oh, I would have added another commit to fix this before you merged
Do you want me to clean these parts up? I wouldn't mind
-reverse wasn't really needed, it's vestigial from work I did to use -comp for transformers

ikitommi:
If you are pushing more perf PRs, would be great if you could push those in too. cheers,

bsless:
I'll do it in a separate PR, want to keep it clean

PR #487, opened 2021-07-27
Add tip: values from :deafault-fn transformer

rynkowsg:
This is an example of deafult value derived from a function placed in a schema property `:default-fn`.

More:
https://clojurians.slack.com/archives/CLDK6MFMK/p1626174163375500?thread_ts=1626171933.375100&cid=CLDK6MFMK


ikitommi:
Thanks!!

Issue #489, opened 2021-07-29
Metadata for schema definition?


vemv:
Unrelated to https://github.com/metosin/malli/issues/125

## Problem statement

The current data notation for defining schemas can be considered pretty verbose:

```clj
[:map
 {:closed true} 
 [:y {:optional true} int?]]
```

It's not something I'd enjoy typing or reading much - feels cluttered.

## Proposed solution

Accept metadata for each vector such as `[:map`, `[:my-property` etc

```clj
^:closed
[:map
 ^:optional
 [:y int?]]
```

Interestingly, metadata tends to get its own syntax highlighting in editors, so when reading these schemas it can feel less cluttered (since one can train the eye to fixate less on metadata).

The tradeoff is that the resulting schemas, while still data, would need expansion if consuming them for arbitrary purposes.

```clj
;; Malli would understand this piece of data OOTB
;; however other consumers would need a little helper to expand metadata into data.
;; Malli could expose such as helper as a public API.
(def MySchema ^:closed [:map ...])
```

Maybe Malli could offer its own `def` that would perform expansion OOTB.

WDYT?

Thanks  - V

ikitommi:
Would the metadata be written into properties so this would hold?:

```clj
(m/form ^:closed [:map ^:optional [:y int?]])
; => [:map {:closed true} [:y {:optional true} int?]]
```

the implementation would be few loc, but not sure if having more ways to do the same is better.

opening this for discussion.

ikitommi:
example impl: #492

rynkowsg:
I listed a few reasons why I think we should think carefully before including this feature. 
Feel free to ignore this comment. It is only my personal opinion.

**More complex syntax, even if you don't use it, you have to document it**
It looks more concise for sure, still, personally I prefer the old map. The old way might be a bit more verbosive but it is easier to read, and there is less code to grasp. Less but good approaches might be easier to grasp for anyone new to the library, then a swedisch table of solutions for achiving the same result. More options is not always better.

**Metadatas are not a bread and butter of a regular Clojure dev**
I'm using Clojure for 3 years now, and I'm not using metadata at all (besides `^:private`), and from my perspective it can bring more confusion then a real good.

**Bringing inconsitency with metadata usage**
```clj
(meta ^:closed [:map])           ;; => {:closed true}
(meta (m/form ^:closed [:map]))  ;; => nil
```
[...]

PR #492, opened 2021-07-29
DRAFT: malli.core/extract-meta

ikitommi:
related to #489 

```clj
(m/schema
  ^:closed
  [:map
   ^:optional
   [:y int?]]
  {::m/extract-meta true})
; => [:map {:closed true} [:y {:optional true} int?]]
```


ikitommi:
not a good idea.

PR #497, opened 2021-08-02
Implement -map-transformer with -comp

bsless:
The clj implementation reverses the order of transformers, this makes
the transformer trace test fail without accounting for the change in order

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -298,14 +298,15 @@
                :output output}
               max (assoc :max max)))))
 
+#?(:clj
+   (defn -entry-transformer [[k t]]
+     (fn [^Associative x]
+       (if-let [e ^MapEntry (.entryAt x k)]
+         (.assoc x k (t (.val e)))
+         x))))
+
 (defn -map-transformer [ts]
-  #?(:clj  (let [tl (LinkedList. ^Collection (mapv (fn [[k v]] (MapEntry/create k v)) ts))]
-             (fn [x] (let [i (.iterator ^Iterable tl)]
-                       (loop [x ^Associative x]
-                         (if (.hasNext i)
-                           (let [e ^MapEntry (.next i), k (.key e)]
-                             (recur (if-let [xe (.entryAt x k)] (.assoc x k ((.val e) (.val xe))) x)))
-                           x)))))
+  #?(:clj (apply -comp (map -entry-transformer ts))
      :cljs (fn [x] (reduce-kv
                      (fn reduce-child-transformers [m k t]
                        (if-let [entry (find m k)]
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -298,20 +298,15 @@
                :output output}
               max (assoc :max max)))))
 
-#?(:clj
-   (defn -entry-transformer [[k t]]
-     (fn [^Associative x]
-       (if-let [e ^MapEntry (.entryAt x k)]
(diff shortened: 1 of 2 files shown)


bsless:
Performance comparison:
```clojure
(ns user
  (:require
   [malli.core :as m]
   [criterium.core :as cc]))


(defn ->map
  [n]
  (zipmap (map (comp keyword str) (range n)) (range n)))

(defn ->ts
  [n]
  (zipmap (map (comp keyword str) (range n)) (repeat inc)))

(doseq [n [1 2 4 8 16 32]
        :let [m (->map n)
              ts (->ts n)
              t (m/-map-transformer ts)]]
  (println n (* 1e9 (first (:mean (cc/quick-benchmark (t m) {})))) 'ns))

;;; Before

;;; 1  19.443216437696076 ns
;;; 2  35.95746697334425 ns
;;; 4  73.24730397835111 ns
;;; 8  169.2157678359164 ns
;;; 16 730.6748680815756 ns
;;; 32 1681.840644110276 ns

;;; After

;;; 1  14.743717058461081
;;; 2  31.4470360315853
;;; 4  65.57619225392742
;;; 8  150.60389954672738
;;; 16 675.8431918008785
;;; 32 1587.0258132383883
```

bsless:
Regarding the tests, using reader conditionals feels a bit messy. I wouldn't mind implementing it otherwise if you have a better suggestion

src/malli/core.cljc
```diff
@@ -298,14 +298,15 @@
                :output output}
               max (assoc :max max)))))
 
+#?(:clj
+   (defn -entry-transformer [[k t]]
+     (fn [^Associative x]
+       (if-let [e ^MapEntry (.entryAt x k)]
```
ikitommi, line 304: `.get` would be enough here?

ikitommi: ... with that, could the cljs code use the same code? Is there `Associative` in cljs?

ikitommi: the transformers could be ordered by key, so the order would be consistent over cljs/cljs?

ikitommi (commented):
Looks great!

src/malli/core.cljc
bsless: > the transformers could be ordered by key, so the order would be consistent over cljs/cljs?

The problem with ordering is that composition reverses the order of application. Map keys have no order. Ideally, I would want a way to capture the scope of the current key, not only value, being transformed by the transformer

> `.get` would be enough here?

Don't think so. Your transformer could have a special way of handling `nil`s

> ... with that, could the cljs code use the same code? Is there `Associative` in cljs?

This requires someone with better knowledge of cljs than mine, which is 0.

bsless: Even if we sort the transformers, to maintain consistency between the cljs and clj implementation we'll need to reverse the result in clj anyway.
Another concern, which is lesser but still exists, is the performance of compiling the schema elements. We already saw some pathological cases can make it worse. I wouldn't want to add more without good cause.

ikitommi:
needed to test to see how it was reversed. Now I got it. Should be as fast or faster to create now with fully ordered on both clj & cljs.

ikitommi:
... the original (tested) ordering was based on ill facts, as Maps lose order when PAM is changed into PHM after the size threshold. Now it's a vector of vectors, always in order.

ikitommi:
Thanks!!

PR #503, opened 2021-08-07
Clean up sci and recursive schema wording in README

frenchy64:


frenchy64:
I kept writing until I could get the build to pass, but no luck :)

README.md
```diff
@@ -1140,7 +1140,9 @@ Any (serializable) function can be used for `:dispatch`:
 
 ## Recursive Schemas
 
-[Local Registry](#local-registry) allows an easy way to create recursive schemas. To be recursive, the schema refers to another schema of the local registry using two useful keywords `:ref` and `:schema`. First, the schema is defined in the local registry, one of the value refer to another schema with the `:ref` keyword. The top-level schema is defined in the second argument of the `:schema` definition:
+To create a recursive schema, introduce a [local registry](#local-registry) with the first argument to `:schema`, and then wrap all recursive positions in the registry with `:ref`. Now use `:schema`'s second argument to define a schema in terms of the recursive schema(s) in the registry.
```
ikitommi, line 1143: Thanks! The local registry key can be set to any Schema via schema property `:registry`, not just to `:schema`:

```clj
[:map {:registry {::x :int, ::y :int}} ::x ::y]
```

frenchy64: Thanks, I removed any concrete schema from the first sentence, and then pointed out that the example uses `:schema`. How's that?

ikitommi: that's good.

ikitommi:
thanks!

PR #504, opened 2021-08-07
Fix flaky `:=>` generative tests

frenchy64:
We need to use unchecked operators here to avoid arithmetic overflows,
since test.check is very good at choosing very small and very large
numbers.

To ensure the problem is actually fixed, now all positive :=> generative
tests are run 1000 times.


ikitommi:
Thanks!!

Issue #452, opened 2021-05-18
On fixing Malli's recursive generation


helins:
For validation, current support for recursive data via `:ref` seems to be appropriate and satisfactory. The problem is that this design is not fit for generation. Any example, even a trivial one, becomes so impractical that a size of 2 or 3 has to be used. Quite often, it is already too high (and one cannot go much lower than that). For comparison, default size is 30 and defaut maximum size during tests is 200. Yet, Malli is an excellent library and `test.check` does provide a pretty good solution to what I am about to describe. The hard part is to solve how they can fit together without disturbing (too much) the current API.

In my experience, solving this issue would really push Malli forwards by a lot.

This is what we can call a problem of "exponential explosion". The fact that the size of some recursive data is not controlled as it gets generated. It means that instead of generating recursive data that is somewhat linear to the given size, it is rather somewhat exponential. Let us take this example from the README:

```clojure
[:schema {:registry {::cons [:maybe [:tuple pos-int? [:ref ::cons]]]}}
  ::cons]
```
This example can be generated easily but only because size has no influence on it, the recursive case depends on `:maybe`. More often, the recursive case does depend on the size. Let us add a `:vector`:

```clojure
[:schema {:registry {::cons [:maybe [:vector [:tuple pos-int? [:ref ::cons]]]]}}
  ::cons]
```

Now, with a default size of 30, generated values are often absolutely massive. So you can easily imagine how this is troublesome with any real world example. This is because every time a vector is generated, by chance, any of its element may be another vector generated with the same size as its parent, and so on.

Some cases can be mitigated by #449, by biasing generation towards leaves (non-recursive items) rather than containers (recursive items, collections of leaves). It is an easy fix and it is useful for other use cases anyway. However, this is not enough. It can mitigate our problem, and sometimes recursive generation needs to be biased anyway, but it is not a general solution.

What is missing is a mechanism that generate leaves with the given size but controls and diminishes the given size for containers. Let us take that kind of definition, very common in my experience:

```clojure
[:schema
 {:registry {::data    [:or
                        ::int
                        ::vector]
             ::int     :int
             ::vector  [:vector
                        [:ref ::data]]}}
 ::data]
```

[...]

PR #507, opened 2021-08-08
Proof of concept recursive gen

frenchy64:
Related https://github.com/metosin/malli/issues/452

Proof of concept! Do not merge.

Changes `:ref` generation to use `recursive-gen` when needed. This probably needs an alpha-renaming pass to really work for all cases (to avoid variable-capture), but at least for the schemas I tried it works very well.

Other possible improvements:
- We can also improve the generation of the `container-gen` to not include `scalar-gen`--see the FIXME's in `malli.generator-test/scalar-container-schema-test` for examples.
- to avoid generating bad `:foo` keys, `[:map [:foo never]]` simplifies to `[:map {:closed true}]`. Perhaps a per-key closed syntax could be introduced like `[:map {:closed #{:foo}}]`.
- use `gen/scale` to distribute recursive generators more evenly ([example](https://github.com/clojure/test.check/blob/045615d1634ab399a447e13abcce64d011a5ebc3/src/main/clojure/clojure/test/check/generators.cljc#L1594-L1599))

This experiment shows that generated sizes grows very slowly--I assume just as fast as `recursive-gen`, but I did not check this.

```clojure
  (dotimes [size-pow 7]
    (time
      (prn
        (str "10^" size-pow)
        (into (sorted-map)
              (frequencies
                (map
                  (comp count flatten)
                  (mg/sample
                    [:schema {:registry {::cons [:maybe [:tuple pos-int? [:ref ::cons]]]}}
                     ::cons]
                    {:size (Math/pow 10 size-pow)})))))))
  ; (out) "10^0" {3 1}
  ; (out) "Elapsed time: 2.030981 msecs"
  ; (out) "10^1" {0 6, 2 3, 3 1}
  ; (out) "Elapsed time: 1.525037 msecs"
  ; (out) "10^2" {0 58, 2 28, 3 10, 4 4}
  ; (out) "Elapsed time: 3.935027 msecs"
  ; (out) "10^3" {0 497, 2 275, 3 138, 4 70, 5 17, 6 3}
  ; (out) "Elapsed time: 28.227256 msecs"
  ; (out) "10^4" {0 4968, 2 2789, 3 1338, 4 636, 5 218, 6 40, 7 9, 8 2}
  ; (out) "Elapsed time: 288.117389 msecs"
  ; (out) "10^5" {0 50219, 2 27130, 3 13152, 4 6214, 5 2484, 6 658, 7 121, 8 19, 9 3}
  ; (out) "Elapsed time: 2909.440828 msecs"
  ; (out) "10^6" {0 500465, 2 272651, 3 129733, 4 62264, 5 25107, 6 7724, 7 1759, 8 267, 9 29, 10 1}
  ; (out) "Elapsed time: 28400.642903 msecs"
```


ikitommi:
Looking forward on what comes out of this. The prewalk could be generic? Maybe something like https://github.com/metosin/malli/blob/master/src/malli/util.cljc#L37-L51

frenchy64:
Ah yes, once I know this will work I'll convert it to an interface.

frenchy64:
I totally missed `-walk`, it's what I wanted in the first place I think.

src/malli/util.cljc
```diff
@@ -34,6 +35,128 @@
...
+(declare subst-schema)
+
+(comment
+  (alpha-rename-schema [:schema {:registry {::foo :int}}
+                        ::foo]
+                       {})
+  ((requiring-resolve 'clojure.repl/pst) 100)
+  )
+
+(defn alpha-rename-schema [schema options]
```
frenchy64, line 98: @ikitommi I need some help creating a function that performs alpha renaming on schemas.

I seem to be able to rename the variables just fine, but on `m/deref` something funny is happening--I'm guessing the ref renames aren't being fully propagated.

```clojure
  (is (= :int
         (m/form
           (m/deref-all
             (mu/alpha-rename-schema [:schema {:registry {::foo :int}}
                                      ::foo]
                                     {})))))
; (err) NullPointerException Cannot invoke "clojure.lang.IFn.invoke()" because "this._ref" is null
; (err) 	malli.core/-ref-schema/reify/reify--5822 (core.cljc:1335)
; (err) 	malli.core/deref (core.cljc:1899)
; (err) 	malli.core/deref (core.cljc:1893)
; (err) 	malli.core/deref-all (core.cljc:1906)
; (err) 	malli.core/deref-all (core.cljc:1901)
; (err) 	malli.core/deref-all (core.cljc:1904)
; (err) 	malli.core/deref-all (core.cljc:1901)
; (err) 	malli.util-test/eval18748 (NO_SOURCE_FILE:1181)
; (err) 	malli.util-test/eval18748 (NO_SOURCE_FILE:1179)
; (err) 	clojure.lang.Compiler.eval (Compiler.java:7181)
; (err) 	clojure.lang.Compiler.eval (Compiler.java:7136)
; (err) 	clojure.core/eval (core.clj:3202)
; (err) 	clojure.core/eval (core.clj:3198)
; (err) 	nrepl.middleware.interruptible-eval/evaluate/fn--963/fn--964 (interruptible_eval.clj:87)
; (err) 	clojure.core/apply (core.clj:667)
; (err) 	clojure.core/with-bindings* (core.clj:1977)
; (err) 	clojure.core/with-bindings* (core.clj:1977)
; (err) 	nrepl.middleware.interruptible-eval/evaluate/fn--963 (interruptible_eval.clj:87)
; (err) 	clojure.main/repl/read-eval-print--9110/fn--9113 (main.clj:437)
; (err) 	clojure.main/repl/read-eval-print--9110 (main.clj:437)
; (err) 	clojure.main/repl/fn--9119 (main.clj:458)
; (err) 	clojure.main/repl (main.clj:458)
; (err) 	clojure.main/repl (main.clj:368)
; (err) 	nrepl.middleware.interruptible-eval/evaluate (interruptible_eval.clj:84)
; (err) 	nrepl.middleware.interruptible-eval/evaluate (interruptible_eval.clj:56)
; (err) 	nrepl.middleware.interruptible-eval/interruptible-eval/fn--994/fn--998 (interruptible_eval.clj:152)
; (err) 	nrepl.middleware.session/session-exec/main-loop--1062/fn--1066 (session.clj:202)
; (err) 	nrepl.middleware.session/session-exec/main-loop--1062 (session.clj:201)
; (err) 	java.lang.Thread.run (Thread.java:831)
```

How would you write this function? See `malli.util-test/alpha-rename-schema-test` for what I want to achieve.

frenchy64:
@ikitommi the proof-of-concept is ready to play with and evaluate.

ikitommi:
Wow, this is a big change, will need to find time to read it through in detail.

frenchy64:
Sounds good. If it doesn't make sense I can write something up to help.

FWIW after trying this technique in Typed Clojure, I think it should be possible to localize it entirely in the generators namespace and not have to introduce any new schemas like `:never`. The basic approach would be to define a special generator that represents an unreachable schema, and then simplify generators based on equality to this "unreachable" generator.

You'd lose the `-simplify` and `-unreachable?` methods for use in other contexts other than generation, but I don't have a good use case to propose other than this.

src/malli/util.cljc
ikitommi: I'll check this.

PR #512, opened 2021-08-14
fix :uuid does not produce humanized error, fixes #511

### LINKED ISSUE #511: :uuid schema does not produce correct humanized error
```
(me/humanize (m/explain :uuid "foo"))
=> ["unknown error"]
```

`uuid?` produces this:
```
(me/humanize (m/explain uuid? "foo"))
=> ["should be a uuid"]
```

Looking at `malli.error/default-errors`, it has an entry for `uuid?` but not for `:uuid`

hti-andre:
Fixes issue #511 

`:uuid` schema now produces same error as `uuid?`
```
(me/humanize (m/explain :uuid "foo"))
=> ["should be a uuid"]
```

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -103,6 +103,7 @@
                                  (and min max) (str "should be between " min " and " max " characters")
                                  min (str "should be at least " min " characters")
                                  max (str "should be at most " max " characters"))))}}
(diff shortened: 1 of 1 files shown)


ikitommi:
Thanks!

Issue #513, opened 2021-08-17
Performance of Schema and Worker Creation


ikitommi:
# Problem Statement

Malli is optimized selectively for runtime performance. The performance critical paths include: validation, explanation, transformation and parsing. Good runtime performance has been achieved by pushing all possible work into worker initialization, away from the runtime.

Creating a schema and a validator and applying it, all at once:

```clj
;; 5.7µs
(m/validate [:or :int :string] 42)
; => true
```

Building the schema and schema validator separately, ahead of time:

```clj
;; 3.0µs
(def schema (m/schema [:or :int :string]))

;; 1.7µs
(def validate (m/validator schema))

;; 4ns
(validate 42)
; => true
```

The validation is over 1000x faster! For production apps with low latency requirements, one should always precompute the workers if possible. Tools like [reitit](https://github.com/metosin/reitit) coercion do this already on behalf of the user.

As it stands out now, the Schema & worker creation performance is only optimised for lines-of-implementation-code, which is not the best metric here. The perf is bad, REALLY bad.

Having a 500ms schema worker compilation phase to emit 1000x faster runtime is totally ok on a hi-performance cloud server but can be a show-stopper on targets like on browsers, especially if running on slow mobile phones.

Flamegraphs (less is better) with `m/validate` and `m/validator`:

using `m/validator`:

<img width="752" alt="Screenshot 2021-08-17 at 9 15 08" src="https://user-images.githubusercontent.com/567532/129674172-b613ea8d-eb68-482e-9a3b-481f17d951de.png">

using `m/validate`:

<img width="1238" alt="Screenshot 2021-08-17 at 9 17 23" src="https://user-images.githubusercontent.com/567532/129674191-0d49bac4-528c-45c0-8a28-200fb2755016.png">

Can we do something to make it better?

Definetely.

Schema and worker creation code should be optimised to a reasonable level for performance (both CPU & Memory) and Malli should support lazy initialisation to avoid stop-the-world pre-calculation of all in single-threaded environments

# What can be done

1. Creating Schema instances
2. Transforming Schemas
3. Guidelines and Tools

## 1. Creating Schema instances

1. `m/schema` is just not optimised. We merge registries, eagerly create forms, parse entries etc. Quick wins available.

```clj
;; 400ns
(m/schema :int)
```

2. all schema instances are created separately - any larger domain model can have hundreds or thousands (or hundreds of thousands) of identically behaving leaf schemas for no good reason. With a smart cache, we could re-use once created instances.

[...]

PR #517, opened 2021-08-21
Faster registry

bsless:
Speed up registry creation and usage

Part of #513

### DIFF
--- a/src/malli/registry.cljc
+++ b/src/malli/registry.cljc
@@ -14,9 +14,19 @@
     (-schema [_ type] (get schemas type))
     (-schemas [_] schemas)))
 
-(defn registry [?registry]
-  (cond (satisfies? Registry ?registry) ?registry
-        (map? ?registry) (simple-registry ?registry)))
+(defn registry
+  [?registry]
+  #?(:clj
+     (when ?registry
+       (if (instance? malli.registry.Registry ?registry)
+         ?registry
+         (if (map? ?registry)
+           (simple-registry ?registry)
+           (when (satisfies? Registry ?registry)
+             ?registry))))
+     :cljs
+     (cond (satisfies? Registry ?registry) ?registry
+           (map? ?registry) (simple-registry ?registry))))
 
 ;;
 ;; custom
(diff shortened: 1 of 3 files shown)


bsless:
```clojure
(def regs
  {:nil nil
   :map {}
   :simple (mr/simple-registry {})
   :composite (mr/composite-registry (mr/simple-registry {}))
   :mutable (mr/mutable-registry (atom {}))})

(doseq [[name r] regs]
  (let [mean (first (:mean (cc/quick-benchmark (mr/registry r) {})))
        [scale unit] (cc/scale-time mean)]
    (println name (* mean scale) unit)))

;;; Before
;; :nil 11.314252099384198 ns
;; :map 11.553363773721745 µs
;; :simple 5.894620337443649 ns
;; :composite 5.851375239599097 ns
;; :mutable 5.893465716928456 ns

;;; After
;; :nil 3.259309437187958 ns
;; :map 22.65670884744675 ns
;; :simple 3.267212075708806 ns
;; :composite 3.287585829320794 ns
;; :mutable 3.289205443217265 ns
```

bsless:
Also think I found a mistake in one of the tests

ikitommi:
thanks!

src/malli/registry.cljc
```diff
@@ -11,12 +11,22 @@
...
   (reify
     Registry
-    (-schema [_ type] (get schemas type))
+    (-schema [_ type] (schemas type))
     (-schemas [_] schemas)))
 
-(defn registry [?registry]
-  (cond (satisfies? Registry ?registry) ?registry
-        (map? ?registry) (simple-registry ?registry)))
+(defn registry
```
ikitommi, line 17: simplified this into:

```clj
(defn registry [?registry]
  (cond (nil? ?registry) nil
        #?@(:clj [(instance? malli.registry.Registry ?registry) ?registry])
        (map? ?registry) (simple-registry ?registry)
        (satisfies? Registry ?registry) ?registry))
```

`instance?` -check works with reified classes? woot.

bsless: They do. I dug into why `satisfies?` is fast for reified classes, and voila!
I also found that `nil?` checks are slightly slower than raw `if`

PR #519, opened 2021-08-21
Implement schemas

bsless:
Faster than extending

Part of #513

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -367,7 +367,11 @@
 (defn -simple-schema [?props]
   (let [props* (atom (if (map? ?props) ?props))]
     ^{:type ::into-schema}
-    (reify IntoSchema
+    (reify
+      Schemas
+      (-into-schema? [_] true)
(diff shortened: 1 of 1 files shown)


bsless:
Also, since we're concerned with cold-start performance, too, passing through `extend` has a very high cost, about 4 orders of magnitude higher _for each type_

ikitommi:
proposed #535 removes the `Schemas` protocol completely.

could you tell more about: 

> Also, since we're concerned with cold-start performance, too, passing through extend has a very high cost, about 4 orders of magnitude higher for each type

bsless:
@ikitommi Using `instance?` will be much faster, I don't know regarding `implements?` because that has to do with how protocols are implemented in cljs
What I meant is that in the original implementation with `Schemas`, all the reified schemas pass at least one time through `satisfies?`. For schemas which are instances of the interface it will be very quick, but for schemas which don't the path will be very slow. For cold start we don't want this overhead.
However, checking directly for instantiation is the fastest solution.

PR #521, opened 2021-08-22
Faster schema constructor vector and into-schema

bsless:
Significantly speed up vector path for `m/schema` constructor and improve `into-schema` parsing

Before:
```clojure
(cc/quick-bench (into-schema :int nil nil nil)) ;; 131.999880 ns
(cc/quick-bench (into-schema :int {:min 2} nil nil)) ;; 249.199673 ns
(cc/quick-bench (schema [:int])) ;; 568.556449 ns
(cc/quick-bench (schema [:int {:min 2}])) ;; 707.044799 ns
(cc/quick-bench (schema [:or {:doc "hi"} :int :string])) ;; 1.863531 µs
```
After:
```clojure
(cc/quick-bench (into-schema :int nil nil nil)) ;; 127.841338 ns
(cc/quick-bench (into-schema :int {:min 2} nil nil)) ;; 195.676156 ns
(cc/quick-bench (schema [:int])) ;; 245.934142 ns
(cc/quick-bench (schema [:int {:min 2}])) ;; 334.652885 ns
(cc/quick-bench (schema [:or {:doc "hi"} :int :string])) ;; 1.488505 µs
```
Part of #513

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1563,13 +1563,22 @@
   "Checks if x is a IntoSchema instance"
   [x] (-into-schema? x))
 
+(defn- -some-map
+  [m]
+  (when m
+    (when (pos? #?(:clj (.count ^Counted m) :cljs (count m)))
+      m)))
+
 (defn into-schema
   "Creates a Schema instance out of type, optional properties map and children"
   ([type properties children]
    (into-schema type properties children nil))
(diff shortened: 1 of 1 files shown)


ikitommi:
cherry-picked the changes and cljs'd those int #523. Thanks!

PR #525, opened 2021-08-23
Step towards JMH benchmarks

bsless:


bsless:
I've been talking with the author, looks like we can (should?) do it without building an uberjar first

PR #522, opened 2021-08-23
Reimplement -parse-entries

bsless:
Cut down most of the overhead, portable solution between clj and cljs

```
(cc/quick-bench (-parse-entries [[:x int?] [:y {:optional true} string?]] nil nil))
Evaluation count : 100554 in 6 samples of 16759 calls.
             Execution time mean : 5.964293 µs
    Execution time std-deviation : 37.595767 ns
   Execution time lower quantile : 5.944926 µs ( 2.5%)
   Execution time upper quantile : 6.028715 µs (97.5%)
                   Overhead used : 1.293497 ns

Found 1 outliers in 6 samples (16.6667 %)
	low-severe	 1 (16.6667 %)
 Variance from outliers : 13.8889 % Variance is moderately inflated by outliers


(cc/quick-bench (-parse-entries [[:x int?] [:y {:optional true} string?]] nil nil))
Evaluation count : 220974 in 6 samples of 36829 calls.
             Execution time mean : 2.734451 µs
    Execution time std-deviation : 21.973588 ns
   Execution time lower quantile : 2.722954 µs ( 2.5%)
   Execution time upper quantile : 2.772454 µs (97.5%)
                   Overhead used : 1.266671 ns

Found 2 outliers in 6 samples (33.3333 %)
	low-severe	 1 (16.6667 %)
	low-mild	 1 (16.6667 %)
 Variance from outliers : 13.8889 % Variance is moderately inflated by outliers
```

Part of #513

PR #527, opened 2021-08-24
Add lightweight keyset wrapper

bsless:
Saves on allocation and provides an immutable facade to underlying
regular set

Followup to #522 
Part of #513

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -272,69 +272,69 @@
         (-fail! ::invalid-ref {:ref e})))))
 
 (defn -parse-entries [children {:keys [naked-keys lazy-refs]} options]
-  (letfn [(-arr->vec [^objects arr]
-            #?(:clj (LazilyPersistentVector/createOwning arr)
-               :cljs (vec arr)))
-          (-arange [^objects arr to]
-            #?(:clj (let [-arr (object-array to)] (System/arraycopy arr 0 -arr 0 to) -arr)
-               :cljs (.slice arr 0 to)))
-          #?(:clj
-             (-set-wrapper
-               [^java.util.Set s]
-               (reify
-                 Collection
-                 (size [_] (.size s))
-                 (containsAll [_ c] (.containsAll s c))
-                 clojure.lang.IPersistentSet
-                 (disjoin [_ key] (disj (set s) key))
-                 (contains [_ key] (.contains s key))
-                 (get [_ key] (when (.contains s key) key))
-                 clojure.lang.Counted
-                 (count [_] (.size s))
-                 clojure.lang.IPersistentCollection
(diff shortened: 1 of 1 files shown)


bsless:
I believe the difference is significant enough to consider this given the benefit for larger schemas:
```clojure
(cc/quick-bench
 ;; 1.799482 µs -> 1.694608 µs
 (m/-parse-entries [[:x int?] [:y {:optional true} boolean?] [:z string?]] nil nil))


(cc/quick-bench
 ;; 5.412949 µs -> 4.899880 µs
 (m/-parse-entries [[:a int?]
                    [:b {:optional true} boolean?]
                    [:c string?]
                    [:u int?]
                    [:v int?]
                    [:w int?]
                    [:x int?]
                    [:y int?]
                    [:z int?]]
                   nil nil))
```

bsless:
@ikitommi any interest in this patch? I can rebase it.

ikitommi:
Thanks for your work on this. Could you check the #539 , in which, I swapped the set to be a map with key->index mapping. This allows effective patching with reusing the parse-results.

ikitommi:
... the map generation could be improved I think, just collect the key-index pairs, create a arraymap in the end?

bsless:
The same mechanism I used for the set could be used here - use a mutable java collection and expose it later by an immutable "view"

ikitommi:
maybe just collecting the key + values into object-array and create persistent with `PersistentArrayMap/createWithCheck`? looks fast, haven't tested.

bsless:
Creating an ArrayMap is tricky. You don't want to do it for over 8 keys, and createWithCheck will be worst-case quadratic for it.

ikitommi:
we already do object-arrays for everything else, just one with double size? looks fast and does fast duplicate key check:

```clj
(let [m {:a 1, :b 2, :c 3, :d 4, :e 5}
      a (object-array (* 2 (count m)))]
  (doseq [[i [k v]] (map-indexed vector m)]
    (aset a (* i 2) k)
    (aset a (inc (* i 2)) v))
  ;; 27ns
  (cc/quick-bench
   (clojure.lang.PersistentArrayMap/createWithCheck a)))
```

bsless:
```java
static public PersistentArrayMap createWithCheck(Object[] init){
	for(int i=0;i< init.length;i += 2)
		{
		for(int j=i+2;j<init.length;j += 2)
			{
			if(equalKey(init[i],init[j]))
				throw new IllegalArgumentException("Duplicate key: " + init[i]);
			}
		}
	return new PersistentArrayMap(init);
}
```

It's not slow, but need to be aware it's quadratic, don't know what which size it becomes too heavy

ikitommi:
Not sure if this is relevant anymore, let's close it. Thanks.

PR #531, opened 2021-08-25
Perf part x

ikitommi:
* faster `m/schema` & `m/into-schema`
* lazy ~childs &~ form with `:or` and `:and`
* simplified `m/-simple-schema` 
* type hints from #521 

```clj
;; 1.7µs => 1.1µs
(bench (m/schema [:or :int :string]))

;; 400ns -> 280ns
(bench (m/schema :int))

(def ?schema
  [:map
   [:x boolean?]
   [:y {:optional true} int?]
   [:z [:map
        [:x boolean?]
        [:y {:optional true} int?]]]])

;; 9.4µs -> 8.5µs
(bench (m/schema ?schema))
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1691,9 +1691,10 @@
    (cond
      (schema? ?schema) ?schema
      (into-schema? ?schema) (-into-schema ?schema nil nil options)
-     (vector? ?schema) (let [t (nth ?schema 0)
-                             n (count ?schema)
-                             ?p (when (> n 1) (nth ?schema 1))]
+     (vector? ?schema) (let [v #?(:clj ^IPersistentVector ?schema, :cljs ?schema)
+                             t #?(:clj (.nth v 0), :cljs (nth v 0))
+                             n #?(:clj (.count v), :cljs (count v))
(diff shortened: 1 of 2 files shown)


src/malli/core.cljc
```diff
@@ -433,50 +430,43 @@
...
-                (-properties [_] properties)
-                (-options [_] options)
-                (-children [_] children)
-                (-parent [_] parent)
-                (-form [_] form)
-                LensSchema
-                (-keep [_])
-                (-get [_ _ default] default)
-                (-set [this key _] (-fail! ::non-associative-schema {:schema this, :key key}))))))))))
+(defn -simple-schema [{:keys [type type-properties pred property-pred min max] :or {min 0, max 0} :as ?props}]
```
bsless, line 433: @ikitommi  What about the case where it's used with a function constructor?
I know it's probably a noop but isn't it confusing?

ikitommi: could be `?props` and destructured in a let?

ikitommi: good point, will change that.

ikitommi: fixed in 91773e9

PR #532, opened 2021-08-25
Unroll and inline -reference?

bsless:
Always bugged me seeing `-reference?` in the flame graphs

```clojure
(m/schema :int) ;; 160 ns -> 130 ns
```


ikitommi:
I haven't seen those on the flamegraphs. Do you happen to have a pic before & after?

bsless:
Before, highlighting `reference`
image
After you can't even see it
image

bsless:
Slightly better before graph
image

bsless:
This could be related to the JVM you're running on. Newer compilers might do a better job

ikitommi:
on my M1 Pro and the latest codebase, I see `65ns -> 61ns` with this, which is not that relevant. Thanks anyway!

bsless:
One point to consider here is that `-reference` will probably be called more on application start-up, before the JIT warms up. 
I ran some experiments with different levels of GC.
Besides generally discovering interesting tidbits, such as a performance degradation at tier 3 before reaching full JIT at tier 4, you get different speedups for different types X compilation stages X JVM versions.
Since I don't think we can make assumptions on how users will run this code (they might even still be using JDK 8), and `-reference?` is hit before full JIT + the number of calls to it are proportional to the number of schemas a user defines and their complexity, the estimated overhead is 3-7ns per schema instance referred to, so if I have a `:map` with 10 entries I'll call `-reference` at least 10 times -> 30-70ns overhead per one schema instance.
This is just a guesstimate regarding the costs and you might end up deciding the impact isn't worth it, but I think it's interesting to consider in a bigger context

PR #535, opened 2021-08-29
Even faster

ikitommi:
Small improvements

* fast -check-children!
* optionally don't check children
* **BREAKING**: rremove `Schemas`, use `instance?` and `implements?` instead
* schema recur
* transducers
* parse record
* delayed forms
* fast (clj) registry

### DIFF
--- a/src/malli/registry.cljc
+++ b/src/malli/registry.cljc
@@ -10,7 +10,7 @@
   (-schemas [this] "returns all schemas from a registry"))
 
 (defn fast-registry [m]
-  (let [fm #?(:clj (doto (HashMap.) (.putAll ^Map m)), :cljs m)]
+  (let [fm #?(:clj (doto (HashMap. 1024 0.25) (.putAll ^Map m)), :cljs m)]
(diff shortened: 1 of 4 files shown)


src/malli/registry.cljc
```diff
@@ -8,11 +9,18 @@
   (-schema [this type] "returns the schema from a registry")
   (-schemas [this] "returns all schemas from a registry"))
 
-(defn simple-registry [schemas]
+(defn fast-registry [m]
+  (let [fm #?(:clj (doto (HashMap.) (.putAll ^Map m)), :cljs m)]
```
bsless, line 13: Might want to tune the initial capacity and collision factor for the map. Since you want to minimize hash collisions, I found initial values like 1024 / 0.25 give better results

ikitommi: thanks. Did you find any difference in using the IndentityMap?

bsless: Nothing conclusive, sadly

PR #538, opened 2021-09-01
Support :enum in clj-kondo

DougHamil:
Currently `:enum` is not supported in the clj-kondo generation (it produces `nil` in the kondo config). This adds support for `:enum`:
* If `:enum` has multiple types for its values (i.e. `[:enum "a" 1]`) then use `:any`
* If `:enum` has a single type use the corresponding clj-kondo type or use `:any` if there is no corresponding type.

Let me know if you need more info!


ikitommi:
sorry for taking so long, looks good, thanks!

PR #539, opened 2021-09-10
Perf part4

ikitommi:
* faster `m/deref` & `m/deref-all`
* remove all `satisfies?` calls for perf reasons.
* entry schemas can reuse parse results in transformations
* fixed `:function` lenses
* faster entry parsing

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -256,7 +256,7 @@
               (aset -entries i e)
               (aset -forms i f)
               (unchecked-inc-int i)))
-          (-schema [e] (schema (cond-> e lazy-refs (-lazy options)) options))
+          (-schema [e] (schema (cond-> e (and (-reference? e) lazy-refs) (-lazy options)) options))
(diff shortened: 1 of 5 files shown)


src/malli/core.cljc
```diff
@@ -202,33 +206,56 @@
...
 (defn- -parse-entry [e naked-keys lazy-refs options i ^objects -children ^objects -entries ^objects -forms -keyset]
   (letfn [(-collect [k c e f i]
-            (-keyset k)
             (let [i (int i)]
+              (-keyset k i)
               (aset -children i c)
               (aset -entries i e)
               (aset -forms i f)
               (unchecked-inc-int i)))
-          (-schema [e] (schema (cond-> (or e (when (-reference? e) e)) lazy-refs (-lazy options)) options))
```
ikitommi, line 231: is this correct?

ikitommi: fixd

Issue #501, opened 2021-08-05
Support JSON Schema time formats


bsless:
Required for #60

- [ ] date-time
- [ ] date
- [ ] time
- [ ] duration

According to the specification in:
https://datatracker.ietf.org/doc/html/rfc3339
https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3
https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

While date-time can be parsed to instant, date, time and duration are a bit more general.
Suggested JVM types:
- date-time: either Instant, LocalDateTime or OffsetDateTime. I suggest OffsetDateTime
- date: [LocalDate](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html)
- time: [LocalTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html)
- duration: [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html), even has a parse method compliant with the RFC

By picking the correct classes we can just use their parse methods and not have to define ad-hoc formatters.

wdyt?

bsless:
Initial draft:

```clojure
(ns malli.time
  (:require
   [clojure.string :as str]
   [malli.core :as m])
  (:import
   (java.time Duration LocalDate OffsetDateTime OffsetTime)))

(comment
  (OffsetDateTime/parse "2016-11-01T16:51:35-01:00")
  (OffsetDateTime/parse "2016-11-01T16:51:35Z")
  (LocalDate/parse "2016-11-01")
  (OffsetTime/parse "16:51:35-01:00")
  (OffsetTime/parse "16:51:35Z")
  (Duration/parse "PT3H"))

(defn ->parser
  [f]
  (fn [x]
    (if (string? x)
      (try
        (f x)
        (catch Exception _ x))
      x)))

(defn -string->duration [x]
  (Duration/parse x))

(defn -string->offset-date-time [x]
  (OffsetDateTime/parse x))

(defn -string->local-date [x]
  (LocalDate/parse x))

(defn -string->offset-time [x]
  (OffsetTime/parse x))

(def default-date-types
  {:date {:class LocalDate :parser -string->local-date}
   :time {:class OffsetTime :parser -string->offset-time}
   :date-time {:class OffsetDateTime :parser -string->offset-date-time}
   :duration {:class Duration :parser -string->duration}})

(defn class-base-name
  [^Class c]
  (last (str/split (.getName c) #"\.")))

[...]

Issue #49, opened 2019-08-26
Better support for dates


ikitommi:
Maybe `:date-time` and `:local-date` mapping to Java8 Time & Google dates? Or should we lean on libs like [tick](https://github.com/juxt/tick)?

nilern:
`tick` pulls in `js-joda`, preferably `malli` would avoid such bloat and opinion.

ikitommi:
Agree. Looking at JSON Schema [dates](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times) and [java.time classes](https://dzone.com/articles/java-8-apis-javautiltime), maybe we should have these implemented:

* `:date` -> `java.util.Date`, `:date-time` in JSON Schema
* `:date-time` -> `java.time.Instant`, `:date-time` in JSON Schema
* `:local-time` -> `java.time.LocalTime`, `:time` in JSON Schema
* `:local-date` -> `java.time.LocalDate`, `:local-date` in JSON Schema

not sure what the js classes would be, maybe `Date` for `:date` and the corresponding versions of `goog.date` for the rest?

rschmukler:
Just linking this here for consideration: https://github.com/henryw374/cljc.java-time

I've done no research, just wanted to make sure it was known to exist for when this is implemented.

rschmukler:
Also worth checking out http://juxt.pro/tick/docs/index.html

ikitommi:
There is a discussion about tick, cljc-java-time and the future of js dates in [clojureverse](https://clojureverse.org/t/progress-update-on-the-clojure-script-date-time-libraries/5344/5). It seems that there isn't a lightweight clj/s option available right now. Suggestion: malli will start with own `malli.time` ns, which has just the needed mappings to make java8 dates available that work with js too. For js, it would be all `js/Date`s for now. 

Is this a bad idea?

## suggestion

| malli           | java                    | javascript | json schma | example string             | info
|-----------------|-------------------------|------------|------------|----------------------------|-----------------------
| :date           | java.util.Date          | js/Date    | "date-time"| "2013-07-16T19:20:51Z"     | java/clojure legacy
| :instant        | java.time.Instant       | js/Date    | "date-time"| "2013-07-16T19:20:51Z"     | UTC, system-time
| :date-time      | java.time.OffsetDateTime| js/Date    | "date-time"| "2013-07-16T19:23:51+03:00"| with offset, user-time
[...]

PR #545, opened 2021-09-23
Add java.time schemas

bsless:
Add support for mostly-compliant ISO-8601 time formats in clj

Enables future support of JSON Schema

Related to #49 and #501

### DIFF
--- a/src/malli/time.clj
+++ b/src/malli/time.clj
@@ -1,0 +1,69 @@
+(ns malli.time
+  (:require
+   [clojure.string :as str]
+   [malli.core :as m])
+  (:import
+   (java.time Duration LocalDate OffsetDateTime OffsetTime)))
+
+(defn ->safe-parser
+  [f]
+  (fn -parse [x]
+    (if (string? x)
+      (try
+        (f x)
+        (catch Exception _ x))
+      x)))
+
+(defn ->error-reporter
+  [parser message]
+  (fn -report [value]
+    (if (string? value)
+      (try
+        (parser value)
+        (catch Exception e (ex-message e)))
+      message)))
+
+(defn -string->duration [x]
+  (Duration/parse x))
+
+(defn -string->offset-date-time [x]
+  (OffsetDateTime/parse x))
+
+(defn -string->local-date [x]
+  (LocalDate/parse x))
+
+(defn -string->offset-time [x]
+  (OffsetTime/parse x))
+
+(def default-date-types
+  {:local-date {:class LocalDate :parser -string->local-date :json-schema/type :date}
+   :offset-time {:class OffsetTime :parser -string->offset-time :json-schema/type :time}
+   :offset-date-time {:class OffsetDateTime :parser -string->offset-date-time :json-schema/type :date-time}
+   :duration {:class Duration :parser -string->duration}})
+
+(defn class-base-name
+  [^Class c]
+  (peek (str/split (.getName c) #"\.")))
+
+(defn time-schema
+  [type]
+  (when-let [props (get default-date-types type)]
+    (let [{klass :class parser :parser} props
+          pred #(instance? klass %)
+          safe-parser (->safe-parser parser)
+          -name (name type)
+          message (str "Should be " -name " or " (class-base-name klass))
+          error-fn (->error-reporter parser message)]
+      (m/-simple-schema
+       {:type type
+        :type-properties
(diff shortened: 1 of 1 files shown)


ikitommi:
looks good, with basic tests could merge this and rename to `malli.experimental.time` so it's usable, despite not complete yet. Or we should add implementations to all concern (generation etc), in case, should we:

* extract core protocols into `malli.protocol`
* keep these in `malli.time`
* malli.core would require `malli.time` so it's all available out-of-the-box

moving protocols is a breaking change, but just not basic users. And, break for the better (malli.core is way too big already)

bsless:
As I was working on it I was afraid protocols would have to be extracted to their own NS for full acceptance.

How about a different PR to extract the protocols then pick it up from there? If it's going to be required by malli.core I can add generators and errors in the relevant namespaces

ikitommi:
Closed #546 without merging.

src/malli/time.clj
```diff
@@ -0,0 +1,69 @@
...
+  (:import
+   (java.time Duration LocalDate OffsetDateTime OffsetTime)))
+
+(defn ->safe-parser
+  [f]
+  (fn -parse [x]
+    (if (string? x)
+      (try
+        (f x)
+        (catch Exception _ x))
```
piotr-yuxuan, line 14: Shall we try `Throwable` to catch really everything and make `(f x)` totally safe?

danieldroit:
the reitit swagger works if I define the `default-date-types` and `time-schema` like this:
```
(def default-date-types
  {:local-date       {:class LocalDate :parser -string->local-date :json-schema/type :string :json-schema/format "date"}
   :zoned-date-time  {:class ZonedDateTime :parser -string->zoned-date-time :json-schema/type :string :json-schema/format "date-time"}
   :offset-time      {:class OffsetTime :parser -string->offset-time :json-schema/type :string}
   :offset-date-time {:class OffsetDateTime :parser -string->offset-date-time :json-schema/type :string :json-schema/format "date-time"}
   :duration         {:class Duration :parser -string->duration}})
```

```
(defn time-schema
  [type]
  (when-let [props (get default-date-types type)]
    (let [{klass :class parser :parser} props
          pred        #(instance? klass %)
          safe-parser (->safe-parser parser)
          -name       (name type)
          message     (str "Should be " -name " or " (class-base-name klass))
          error-fn    (->error-reporter parser message)]
      (m/-simple-schema
        {:type type
         :type-properties
               (cond-> {:error/fn         error-fn
                        :decode/json      {:enter safe-parser}
                        :encode/json      {:enter #(prn %)}
                        :json-schema/type (:json-schema/type props -name)}
                 (:json-schema/format props -name) (assoc :json-schema/format (:json-schema/format props -name)))
         :pred pred}))))
```

src/malli/time.clj
danieldroit: Isn't the `->safe-parser` silently swallowing data replacing it with `nil` when it can't parse it?

src/malli/time.clj
```diff
@@ -0,0 +1,69 @@
...
+      x)))
+
+(defn ->error-reporter
+  [parser message]
+  (fn -report [value]
```
rlovtangen, line 19: Should there be a second param here?
If I specify a property as `:local-date` and try to send a non-parsable value, e.g. "2022-01-aa", I get:
_Error: Wrong number of args (2) passed to: slakt-service.malli.time/->error-reporter/-report--28426_ 

If I change this line to 
`(fn -report [value _]`
I get the error message "Should be local-date or LocalDate" as expected

rlovtangen:
Maybe also add a default json-schema/example?

`:local-date {:class LocalDate :parser -string->local-date :json-schema/type :local-date :json-schema/example "2022-02-23"}`
...
`:json-schema/example (:json-schema/example props -name)`

danieldroit:
What do I need to add/change so it works with edn, not just json?

frenchy64:
I think `#(instance? klass %)` compiles to slower code than `#(instance? LocalDate %)`, you might want to hardcode each pred in `default-date-types`.

Sam-3344:
Can I use :local-date in malli now? Or should I define it on my own. I tried using local-date. It failed. 
So I tried this : :re #"\d{4}-\d{2}-\d{2}"
It stills fails as such :
 (m/validate s2 {:a 1 :b "2022-07-28"}
=> false

Issue #236, opened 2020-07-25
Investigate a Protocol-based top-level API


ikitommi:
## currently

Options is an optional parameter to all/most public apis. Would be cleaner if there was a `SchemaContext` Record to hold all configuration and state for working with Schemas. Should have at least `:registry` and `:evaluator` fields. 

It would implement Protocols that would cover the current public APIs and would be passed as FIRST argument to apis. Would make top-level configuration explicit:

```clj
;; using the default context
(m/schema [:and int?])

;; using explicit context
(m/schema ctx [:and int?])
 
;; explicit contexts
(def ctx (m/default-context))

(def ctx (-> (m/default-context) (assoc :evaluator (malli.sci/evaluator))))

(def ctx (m/context {:registry ..., :evaluator ...}))
```

## why?

* Protocols yield coherent public apis, where the (optional) context is always the first argument
* Records & Protocols give type errors (cosmetic)
* Easy to extend
* Clear separation between one-time/use-case options and shared/context options:

```clj
;; ctx has things like :registry, options something for this invocation
(mg/generate ctx [:maybe int?] {:size 10, :seed 10})
```

## draft core Protocol

```clj
(defprotocol SchemaContext
  (-schema [this ?schema])
  (-into-schema [this type properties children])
  (-registry [this])
  (-evaluator [this]))
```

ikitommi:
relates to #235.

opqdonut:
Closing this for now. Not in scope.

PR #546, opened 2021-09-26
Move core protocols to new ns

bsless:
Open questions:
What to do with instance check predicates? Some are in protocols ns, some in core.

Possible solution:
Move regex protocol extension back to core.

Prerequisite for #545 and many more(?)

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1,71 +1,42 @@
 (ns malli.core
   (:refer-clojure :exclude [eval type -deref deref -lookup -key])
   #?(:cljs (:require-macros malli.core))
-  (:require [malli.sci :as ms]
-            [malli.impl.util :as miu]
-            [malli.impl.regex :as re]
-            [malli.registry :as mr])
-  #?(:clj (:import (java.util.regex Pattern)
-                   (clojure.lang Associative IPersistentCollection MapEntry IPersistentVector LazilyPersistentVector PersistentArrayMap)
-                   (malli.impl.util SchemaError)
-                   (java.util.concurrent.atomic AtomicReference)
-                   (java.util Collection LinkedList))))
+  (:require
+   [malli.protocols
+    :as p
+    :refer [-type -type-properties -properties-schema -children-schema
+    -into-schema -validator -explainer -parser -unparser -transformer
+    -walk -properties -options -children -parent -form -entries -keep
+    -get -set -ref -deref -regex-op? -regex-validator -regex-explainer
+    -regex-unparser -regex-parser -regex-transformer -regex-min-max
+    -accept -inner -outer -value-transformer]]
+   [malli.sci :as ms]
+   [malli.impl.util :as miu]
+   [malli.impl.regex :as re]
+   [malli.registry :as mr])
+  #?(:clj
+     (:import
+      (java.util.regex Pattern)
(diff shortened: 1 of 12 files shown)


bsless:
Moved extend to core ns

ikitommi:
Hmm. Does this actually solve the problem of splitting schemas into separate ns's while still being part of the default api? e.g. `malli.time` still needs functions from the current core (`m/-simple-schema`,  `m/schema` or others) to work.

I think the root problem is that `mali.core` has both the impl-part (protocols, IntoSchema impls) and the actual stateful default-registry, which collects everything together.

So, let's state the problem(s) we are trying to solve.

1. Easy way to create extensions to the library, e.g. "time-schemas", "util-schemas"
2. Easy way to use them.

for 1, it's easy to do already, just depend on the `mallil.core` and you have all the things available.

for 2, not so, as you can't both depend on `malli.core´ and inject you schemas as default into the default registry. But, what if we added easy way to register modules.

e.g. without needing JVM/Compiler options:

```clj
(require '[malli.core :as m])
(require '[malli.mutable :as mm])

;; swap the default registry, without tears
(r/set-default-registry! (mm/registry (atom {}))

;; explicitly register schemas from a module
(mm/register! (malli.time/schemas))

;; register a single schema
(mm/register! ::user [:map [:name :string] [:birthday :local-date])

(m/validate ::user {:name "tiina", :birthday (java.time.LocalDate/of 1980 12 12)})
; => true
``` 

Will open a discussion on this.

bsless:
Besides the need to figure out what to do with the registry, I imagined it would only be a first step in refactoring
After protocols are moved, the schema implementations can move as well, leaving in `malli.core` only the public APIs, then you have inverted dependencies and any other module can depend on `malli.impl.core` (example name) and instantiate their own registry / schemas

ikitommi:
Thing is, I'm not anymore that this is the way to go: currently, the schema implementations also depend on the public api: many impls call `m/schema`, `m/from-ast` etc directly - and for good reason - the children could be anything and the logic in `m/schema` is just meant for that use case (parse all forms). Also, now everything is in `malli.core`, it would add some mental overhead for library developers and extenders to remember what should come from impl, what should come from protocols (protocol api is quite large atm). Not a big thing, but still not speaking for the namespace refactoring.

We **could** work around this by
1. moving the public api also behind a protocol => all impls just depend on the protocols (or the impl)
2. inject the public functions from the public functions, e.g. `(assoc options ::schema schema)` => slow, easy to mess up (new public functions NOT injecting -> can fail late at runtime)

both add complexity. If there would be really a good reason for the 1, it would justify making the move. I have tried to find a good reason from the #236, but it also would be a big breaking change and not sure if that's worth it at the moment. Might be the last thing before 1.0.0. Or a 2.0.0. Or not going to happen ever.

Today, I think that the first version of `malli.time` could be standalone ns, depending directly malli.core and thus, not part of the default registry. now as it's easy to swap the default registry, using those in the project is quite straightforward (like `malli.util` schemas).

I think malli will in the end evolve to have the protocols and impls as separate ns'es, but not happy of doing things that add complexity without really good reason. I don't see that yet: it might come, it might be there already or it might not ever come.

Comments welcome.

ikitommi:
As discussed, moving the protocols would be beneficial only if a much bigger reordering would be applied. It's too big at the moment, so closing this. Thanks for making the PR anyway!

PR #549, opened 2021-10-06
Min version required for PlantUML

brettrowberry:
I was on 0.5.1 and then had a class path error when calling `(require '[malli.plantuml :as plantuml])`. Then, I figured out it was added in 0.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2086,6 +2086,8 @@ Visualized with [Graphviz](https://graphviz.org/):
 
 ### PlantUML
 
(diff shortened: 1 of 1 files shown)


ikitommi:
Thanks, but idea of README is to cover the most recent version, history is described in the CHANGELOG. Addition of PlantUML-support: https://github.com/metosin/malli/blob/master/CHANGELOG.md#060-2021-08-08

PR #570, opened 2021-11-10
Fix #569: Correctly handle Pattern/RegExp class

### LINKED ISSUE #569: Pattern/RegExp class schema doesn't actually check if value is RegExp instance
```
(m/validate #?(:clj java.util.regex.Pattern, :cljs js/RegExp) #"foo")
;; :malli.core/child-error {:type :re, :properties nil, :children nil, :min 1, :max 1}
```

I'm quite sure the implementation should check if the value is Regexp instance here.

We don't seem to have any test cases for this.

We could also consider if there should be an easy cross-platform way to create this schema?

Deraen:


Deraen:
Might be better to just replace Pattern/RegExp schema with new `-instance-schema` or such, instead of trying to make `-re-schema` do two different things.

PR #582, opened 2021-11-23
Optional argument `default-fn` to `mt/default-value-transformer`

piotr-yuxuan:
See an example usage here: https://github.com/piotr-yuxuan/malli-cli/blob/main/src/piotr_yuxuan/malli_cli.clj#L220-L222

This doesn't change existing behaviour but only add `default-fn` optional argument. When set, this function is applied on the default value found under key `key`. In the example above this optional, additional indirection allows a seamless way to define `:env-var` defaults.


piotr-yuxuan:
I didn't expect the tests to fail when adding one optional argument atop upstream commit – but I guess it's what tests are for 😛 If there were any interest in this, happy to make tests pass, work out some real-life examples, and add it to the README.

piotr-yuxuan:
Just amended the commit, which makes the update quite straightforward I guess.

ikitommi:
Looks good, could you add a test for this to avoid future regression.

ikitommi:
changed this to 2-arity, passing in the schema too as first argument:

```clj
(m/decode
 [:map
  [:os [:string {:property "os.name"}]]
  [:timezone [:string {:property "user.timezone"}]]]
 {} 
 (mt/default-value-transformer 
  {:key :property
   :default-fn (fn [_ x] (System/getProperty x))}))
; => {:os "Mac OS X", :timezone "Europe/Helsinki"}
```

ikitommi:
#644

piotr-yuxuan:
Sorry for the broken tests, I should have been more engaged and proactively merge `master` into this branch. Thanks very much for the merge!

PR #583, opened 2021-11-27
allow swapping the default registry without tears

ikitommi:
```clj
(require '[malli.core :as m]
         '[malli.util :as mu]
         '[malli.registry :as mr]
         '[malli.generator :as mg])

;; look ma, just works
(mr/set-default-registry!
  (mr/composite-registry
    (m/default-schemas)
    (mu/schemas)))

(mg/generate
  [:merge
   [:map [:x :int]]
   [:map [:y :int]]])
; => {:x 0, :y 92}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1990,8 +1990,6 @@ With this, you can register just what you need and rest are DCE'd. The previous 
 
 Malli supports multiple type of registries.
 
-Since 0.7.0, one can set up the default re
-
 ### Immutable registry
 
 Just just a `Map`.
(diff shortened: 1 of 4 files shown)


ikitommi:
related discussion: https://github.com/metosin/malli/discussions/565

README.md
```diff
@@ -1963,23 +1939,66 @@ See also [Recursive Schemas](#recursive-schemas).
...
+```
+
+With this, you can register just what you need and rest are DCE'd. The previous example results in just a 3KB gzip bundle.
+
+## Registry implemenations
+
+Malli supports multiple type of registries.
 
-Malli supports multiple types of registries.
+Since 0.7.0, one can set up the default re
```
Deraen, line 1993: Unfinished sentence

ikitommi: good catch, fixed.

Issue #585, opened 2021-12-03
Function with Sequential return value cannot define as function schema


MokkeMeguru:
Problem Overview

```clojure
(defn gen-one [num-times]
  (repeat num-times 1))
(m/validate [:* [:= 1]] (gen-one 10)) ;; true

(m/=> gen-one [:=> [:cat int?] [:* [:= 1]]]) 
;; will cause
;; Execution error (ExceptionInfo) at malli.core/-fail! (core.cljc:119).
;; :malli.core/invalid-schema {:schema nil}
```

and Also

```clojure
(defn wrapped-gen-one [num-times]
  {:result (repeat num-times 1)})

(m/=> wrapped-gen-one [:=> [:cat int?] [:map [:result [:* [:= 1]]]]])
;; will cause
;; Execution error (ExceptionInfo) at malli.core/-fail! (core.cljc:119).
;; :malli.core/invalid-schema {:schema nil}
```

Question 

How to define the function spec with sequential return value?

MokkeMeguru:
I found when this problem caused, I enabled malli.clj-kondo, I will throw the smallest project soon.

ikitommi:
fixed in `[metosin/malli "0.7.2"]`

PR #586, opened 2021-12-04
refine clj-kondo's sequential spec

MokkeMeguru:
related with : https://github.com/metosin/malli/issues/585

Now, `malli.clj-kondo` and also `malli.dev` will cause error when we wrote the sequential spec have map / tuple spec.

e.g.
 
```clojure
(def Item [:map [:name :string] [:price :int]])
(def ItemList [:+ Item])
(clj-kondo/transform ItemList) ;; will cause error
```

so that, I propose the temporary fix and also, some tests to fix them.

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -114,15 +114,9 @@
 (defmethod accept :schema [_ schema _ options] (transform (m/deref schema) options))
 (defmethod accept ::m/schema [_ schema _ options] (transform (m/deref schema) options))
 
-(defn rest-transform [child]
-  (cond
-    (vector? child) :any
-    (map? child) :any
-    :else (transform child)))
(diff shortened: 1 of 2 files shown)


ikitommi:
Hi. As the `child` is already transformed into clj-kondo format, wouldn't:

```clj
(defmethod accept :+ [_ _ [child] _] {:op :rest, :spec child})
```

be the correct fix?

with that, the example would be:

```clj
(def Item [:map [:name :string] [:price :int]])
(def ItemList [:+ Item])
(clj-kondo/transform ItemList)
;{:op :rest
; :spec {:op :keys
;        :req {:name :string
;              :price :int}}}
```

MokkeMeguru:
@ikitommi thanks your advice! It seems to works correctly.

PR #587, opened 2021-12-05
Add string schema properties

bsless:
Currently only implemented for JVM
What do you think?

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -571,6 +571,75 @@
     (fn [x] (= (namespace x) ns-name))))
 
 ;;
+;; string schema helpers
+;;
+
+#?(:clj
+   (defn -charset-predicate
+     [o]
+     (case o
+       :digit #(Character/isDigit ^char %)
+       :letter #(Character/isLetter ^char %)
+       :letter-or-digit #(Character/isLetterOrDigit ^char %)
+       :alphanumeric #(Character/isLetterOrDigit ^char %)
+       :alphabetic #(Character/isAlphabetic (int %))
+       (cond
+         (set? o) (miu/-some-pred (map -charset-predicate o))
+         (char? o) #(= ^char o %)
+         (fn? o) o
+         (nil? o) nil
+         :else (throw (ex-info "Invalid string predicate" {:pred o}))))))
+
+#?(:clj
+   (defn string-char-predicate
+     [p]
+     (fn charset-pred ^Boolean [^String s]
+       (let [n (.length s)]
+         (loop [i 0]
+           (if (= i n)
+             true
+             (if (p (.charAt s (unchecked-int i)))
+               (recur (unchecked-inc i))
+               false)))))))
+
+#?(:clj
+   (defn find-blank-method
+     []
+     (try
+       (.getMethod String "isBlank" (into-array Class []))
+       #(.isBlank ^String %)
+       (catch Exception _
+         (require 'clojure.string)
+         clojure.string/blank?))))
+
+#?(:clj
+   (def blank? (find-blank-method)))
+
+#?(:clj
+   (defn -string-predicates
+     ([{:keys [charset pattern non-blank]}]
+      (let [pattern
+            (when pattern
+              (let [pattern (re-pattern pattern)]
+                #(.find (.matcher ^Pattern pattern ^String %))))
+            charset
+            (when charset
+              (let [p (-charset-predicate charset)]
+                (string-char-predicate p)))
+            non-blank (when non-blank #(not (blank? %)))]
+        (-> non-blank
+            (miu/-maybe-and charset)
+            (miu/-maybe-and pattern))))))
+
+#?(:clj
+   (defn -string-property-pred
+     []
+     (fn [properties]
+       (miu/-maybe-and
+        ((-min-max-pred #(.length ^String %)) properties)
+        (-string-predicates properties)))))
+
+;;
 ;; Schemas
 ;;
 
@@ -625,7 +694,7 @@
 
 (defn -nil-schema [] (-simple-schema {:type :nil, :pred nil?}))
 (defn -any-schema [] (-simple-schema {:type :any, :pred any?}))
-(defn -string-schema [] (-simple-schema {:type :string, :pred string?, :property-pred (-min-max-pred count)}))
+(defn -string-schema [] (-simple-schema {:type :string, :pred string?, :property-pred #?(:clj (-string-property-pred) :cljs (-min-max-pred count))}))
 (defn -int-schema [] (-simple-schema {:type :int, :pred int?, :property-pred (-min-max-pred nil)}))
 (defn -double-schema [] (-simple-schema {:type :double, :pred double?, :property-pred (-min-max-pred nil)}))
 (defn -boolean-schema [] (-simple-schema {:type :boolean, :pred boolean?}))
--- a/src/malli/impl/util.cljc
+++ b/src/malli/impl/util.cljc
@@ -65,3 +65,10 @@
 (def ^{:arglists '([[& preds]])} -some-pred
   #?(:clj (-pred-composer or 16)
      :cljs (fn [preds] (fn [x] (boolean (some #(% x) preds))))))
+
+(defn -maybe-and
+  [f g]
+  (cond
+    (and f g) #(and (f %) (g %))
+    f f
+    g g))
--- a/test/malli/core_test.cljc
+++ b/test/malli/core_test.cljc
@@ -2622,3 +2622,45 @@
   (is (= ["1"] (m/-vmap str (subvec [1 2] 0 1))))
   (is (= ["1"] (m/-vmap str (lazy-seq [1]))))
   (is (= ["1" "2"] (m/-vmap str [1 2]))))
+
+#?(:clj
+   (deftest string
+     (testing "pattern"
+       (let [s (m/schema [:string {:pattern "foo"}])]
+         (is (true? (m/validate s "foo")))
+         (is (true? (m/validate s "afoo")))
+         (is (true? (m/validate s "fooa")))
+         (is (false? (m/validate s "foao"))))
+       (let [s (m/schema [:string {:pattern "^foo"}])]
+         (is (true? (m/validate s "foo")))
+         (is (false? (m/validate s "afoo")))
+         (is (true? (m/validate s "fooa")))
+         (is (false? (m/validate s "foao")))))
+     (testing "charset"
+       (let [s (m/schema [:string {:charset :alphabetic}])]
+         (is (true? (m/validate s "foo")))
+         (is (false? (m/validate s "fo1o"))))
+       (let [s (m/schema [:string {:charset :letter}])]
+         (is (true? (m/validate s "foo")))
+         (is (false? (m/validate s "fo1o"))))
+       (let [s (m/schema [:string {:charset :letter-or-digit}])]
+         (is (true? (m/validate s "foo")))
+         (is (true? (m/validate s "fo0")))
+         (is (false? (m/validate s "f-1o"))))
+       (let [s (m/schema [:string {:charset #{\- :letter-or-digit}}])]
+         (is (true? (m/validate s "foo")))
+         (is (true? (m/validate s "fo0")))
+         (is (true? (m/validate s "f-1o")))
+         (is (false? (m/validate s "f?1o")))))
+     (testing "non blank"
+       (let [s (m/schema [:string {:non-blank true}])]
+         (is (true? (m/validate s "foo")))
+         (is (false? (m/validate s "")))
+         (is (false? (m/validate s "  ")))))
+     (testing "Combined"
+       (let [s (m/schema [:string {:non-blank true :pattern "foo" :charset :letter-or-digit}])]
+         (is (true? (m/validate s "foo")))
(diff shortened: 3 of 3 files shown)


src/malli/core.cljc
```diff
@@ -570,6 +570,82 @@
...
+         (when pattern
+           (let [pattern (re-pattern pattern)]
+             #?(:clj #(.find (.matcher ^Pattern pattern ^String %))
+                :cljs #(boolean (re-find pattern %)))))
+         charset
+         (when charset
+           (let [p (-charset-predicate charset)]
+             (string-char-predicate p)))
+         non-blank (when non-blank #(not (blank? %)))]
+     (-> non-blank
```
ikitommi, line 635: is the `:non-blank` needed? much shorter to use `:min`:

```clj
[:string {:non-blank true}]
[:string {:min 1}]
```

ikitommi (commented):
Looks Great! few comments attached. Also, to be complete, we need also:

1. JSON Schema mappings
2. Changes to generator

src/malli/core.cljc
```diff
@@ -570,6 +570,82 @@
...
+  (case o
+    :digit #?(:clj #(Character/isDigit ^char %) :cljs -numeric-char?)
+    :letter #?(:clj #(Character/isLetter ^char %) :cljs -letter?)
+    :letter-or-digit #?(:clj #(Character/isLetterOrDigit ^char %) :cljs -alphanumeric?)
+    :alphanumeric #?(:clj #(Character/isLetterOrDigit ^char %) :cljs -alphanumeric?)
+    :alphabetic #?(:clj #(Character/isAlphabetic (int %)) :cljs -letter?)
+    (cond
+      (set? o) (miu/-some-pred (mapv -charset-predicate o))
+      (char? o) #?(:clj #(= ^char o %) :cljs (let [i (.charCodeAt o 0)] #(= i %)))
+      (fn? o) o
```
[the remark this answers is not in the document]
bsless: Idea was to keep open to extension and let users specify their own function as predicate. Will use Eval

src/malli/core.cljc
```diff
@@ -570,6 +570,82 @@
...
+      (nil? o) nil
```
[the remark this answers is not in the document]
bsless: Combinators will either throw it away (maybe-and) and if everything is nil the simple schema constructor will throw it away

src/malli/core.cljc
```diff
@@ -570,6 +570,82 @@
...
+      (loop [i 0]
+        (if (= i n)
+          true
+          (if (p #?(:clj  (.charAt s (unchecked-int i))
+                    :cljs (.charCodeAt s (unchecked-int i))))
+            (recur (unchecked-inc i))
+            false))))))
+
+#?(:clj
+   (defn find-blank-method
```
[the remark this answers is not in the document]
bsless: That's a huge breaking change for users. Sadly, there's still plenty of Java 8 in the world and we must accommodate

src/malli/core.cljc
bsless: valid min 1 string: `" "`, but it's invalid for `non-blank`

bsless: This could be solved by transformers, but users can certainly want to specify they want a string with minimum length which is not blank. Blank in this case is a superset of empty, which I almost missed in the beginning

src/malli/core.cljc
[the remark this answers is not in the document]
Deraen: @ikitommi Definitely over half of our work projects are still on Java 8

Deraen:
- Charset name here is quite strange? It doesn't refer to charset really.
- Looks like after additional changes this works fully on Cljs now? I would be really careful of adding JVM only features, especially if there is no schema creation time error about using non-supported properties on some env.

Deraen:
Perhaps just `:characters` (or `:chars`)?

Charset (or even character set) means completely different thing in the same context, so it is not an option to use that: https://en.wikipedia.org/wiki/Character_encoding

Quick test to check this is faster than just using equivalent patterns:

  (bench/quick-bench (validate [:string {:charset :digit}] "123"))
  ;; 470ns

  (bench/quick-bench (validate [:string {:pattern "^\\d*$"}] "123"))
  ;; 765ns

src/malli/core.cljc
```diff
@@ -570,6 +570,79 @@
...
+#?(:cljs (defn -numeric-char? [c] (and (< 47 c) (< c 58))))
+#?(:cljs (defn -upper-alpha-char? [c] (and (< 64 c) (< c 91))))
+#?(:cljs (defn -lower-alpha-char? [c] (and (< 96 c) (< c 123))))
+#?(:cljs (defn -letter? [c] (or (-lower-alpha-char? c) (-upper-alpha-char? c))))
+#?(:cljs (defn -alphanumeric? [c] (or (-letter? c) (-numeric-char? c))))
+
+(defn -charset-predicate
+  [o]
+  (case o
+    :digit #?(:clj #(Character/isDigit ^char %) :cljs -numeric-char?)
```
Deraen, line 586: Btw. there is both char and int versions of the predicates in Java:

https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isLetter(char)
https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isLetter(int)

The int version supports unicode characters.

Deraen: Okay, this is probably fine. Char supports 0-65535 so it is enough to support unicode range 0x0000 - 0xFFFF.

Clojure characters don't support supplementary char ranges either. Though strings support:

0x2F81A:

\冬
=> Unsupported character: \冬

(.codePointAt "冬" 0)
=> 194586

(char (.codePointAt "冬" 0))
=> Value out of range for char: 194586

(Character/isLetter (int 0x2F81A))
=> true

(int (.charAt "冬" 0))
=> 55422 in this case .charAt only returns first two bytes.

Deraen: JS charCodeAt works the same as JVM charAt.

src/malli/json_schema.cljc
```diff
@@ -136,7 +136,18 @@
 (defmethod accept :nil [_ _ _ _] {:type "null"})
 
 (defmethod accept :string [_ schema _ _]
-  (merge {:type "string"} (-> schema m/properties (select-keys [:min :max]) (set/rename-keys {:min :minLength, :max :maxLength}))))
+  (let [props (-> schema m/properties)
+        pattern (case (:charset props)
+                  :digit "^[0-9]*$"
```
Deraen, line 141: These patterns don't match the Character predicates, the predicates allow other ranges, e.g.:

'\u0030' through '\u0039', ISO-LATIN-1 digits ('0' through '9')
'\u0660' through '\u0669', Arabic-Indic digits
'\u06F0' through '\u06F9', Extended Arabic-Indic digits
'\u0966' through '\u096F', Devanagari digits
'\uFF10' through '\uFF19', Fullwidth digits

There are some classes in at least JVM Pattern which might match the predicates, but not sure if there are equivelents to all, and what does JS support. Listing all the ranges might work for some cases.

src/malli/core.cljc
bsless: Problem is I'm using `charAt` which returns char, not int

src/malli/json_schema.cljc
bsless: I can use unicode ranges, this works:
```
(re-find #"[\u0061-\u00F6]" "a")
```

Deraen:
It is possible to rewrite the string loop and char checks to work with unicode values, using codePointAt:

```
diff --git a/src/malli/core.cljc b/src/malli/core.cljc
index e8ab01b..2878c12 100644
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -583,13 +583,14 @@
 (defn -charset-predicate
   [o]
   (case o
-    :digit #?(:clj #(Character/isDigit ^char %) :cljs -numeric-char?)
-    :letter #?(:clj #(Character/isLetter ^char %) :cljs -letter?)
-    (:alphanumeric :letter-or-digit) #?(:clj #(Character/isLetterOrDigit ^char %) :cljs -alphanumeric?)
+    :digit #?(:clj #(Character/isDigit ^int %) :cljs -numeric-char?)
+    :letter #?(:clj #(Character/isLetter ^int %)
+               :cljs -letter?)
+    (:alphanumeric :letter-or-digit) #?(:clj #(Character/isLetterOrDigit ^int %) :cljs -alphanumeric?)
     :alphabetic #?(:clj #(Character/isAlphabetic (int %)) :cljs -letter?)
     (cond
       (set? o) (miu/-some-pred (mapv -charset-predicate o))
-      (char? o) #?(:clj #(= ^char o %) :cljs (let [i (.charCodeAt o 0)] #(= i %)))
+      (char? o) #?(:clj #(= ^int o %) :cljs (let [i (.charCodeAt o 0)] #(= i %)))
       :else (eval o))))

 (defn string-char-predicate
@@ -599,10 +600,13 @@
       (loop [i 0]
         (if (= i n)
           true
-          (if (p #?(:clj  (.charAt s (unchecked-int i))
-                    :cljs (.charCodeAt s (unchecked-int i))))
-            (recur (unchecked-inc i))
-            false))))))
+          (let [cur-char #?(:clj  (.codePointAt s (unchecked-int i))
+                            :cljs (.codePointAt s (unchecked-int i)))]
+            (if (p cur-char)
+              (recur (unchecked-add i #?(:clj (Character/charCount cur-char)
+                                         ;; TODO: Is there better way?
+                                         :cljs (.length (String/fromCodePoint cur-char)))))
+              false)))))))
```

It doesn't seem to hurt the performance either:

  ;; 470ns
  (bench/quick-bench (validate [:string {:charset :letter}] "冬"))
  ;; 840ns
  (bench/quick-bench (validate [:string {:pattern "^\\p{L}*$"}] "冬"))

src/malli/core.cljc
```diff
@@ -570,6 +570,79 @@
   (when-let [ns-name (some-> properties :namespace name)]
     (fn [x] (= (namespace x) ns-name))))
 
+;;
+;; string schema helpers
+;;
+
+#?(:cljs (defn -numeric-char? [c] (and (< 47 c) (< c 58))))
```
Deraen, line 577: These checks work differently than the JVM versions.

bsless:
> Perhaps just `:characters` (or `:chars`)?

I think characters is good, I'll make the appropriate change

Deraen:
What is the use case for this feature?

The performance is a bit better (~2x) than Patterns but I don't think that is a huge difference usually. Patterns already support unicode quite well, and both JVM and JS.

This is surprisingly complex feature due to character encodings, and what letter are included in the character classes.

`[A-Za-z]` and just checking for `64 < c < 91` and `96 < c < 123` break e.g. with Finnish and Scandinavian languages. I don't think providing features that work only in specific cases is that useful. And now the JVM predicates work considerably differently than Cljs predicates. At least we should be very careful to ensure these work similarly in both platforms.

bsless:
Not sure why mix the constructors here:
```
(require '[criterium.core :as cc])
(require '[malli.core :as m])

(def v0 (m/validator [:string {:charset :digit}]))
(def v1 (m/validator [:string {:pattern #"^\d*$"}]))

(cc/quick-bench (v0 "123")) ;; 16.756693 ns
(cc/quick-bench (v0 "123456789")) ;; 48.106319 ns
(cc/quick-bench (v0 "123456789a")) ;; 53 ns

(cc/quick-bench (v1 "123")) ;; 71.737728 ns
(cc/quick-bench (v1 "123456789")) ;; 109.460351 ns
(cc/quick-bench (v1 "123456789a")) ;; 128 ns
```

bsless:
The use case for this feature is twofold:
- regular expressions are slower
- expressive power - being able to say "a string is alphanumeric" is more readable than `^[A-Za-z0-9]*$`
- Especially when mixed with other constraints, like "it should have at least two elements"
- Composition / working with data: specify a set of options instead of a regex. One of the options can be a characters set, too, or an ad-hoc validation function

I've seen too many examples at work for such use cases and I don't really want a regex on my hot path

bsless:
Doing some more thinking about this, how about two properties, `include-chars` and `exclude-chars`?

src/malli/core.cljc
evg-tso: 0x2F81A (which is \⾁) is supported under the 12161 unicode: ([link](https://unicodemap.org/details/0x2F81/index.html))
```Clojure
(int \⾁)
=> 12161
```

Same with all other unicode characters:
```Clojure
(char 8809)
=> \≩
(char 508)
=> \Ǽ
(int \Ǽ)
=> 508
(int \≩)
=> 8809
(int \Θ)
=> 920
(char 33071)
=> \脯
```

ikitommi:
Looking at the old PRs, I think the work by @frenchy64 is another way of doing this, but didn't see PR just for string. Would be great to have at least a minimum part of this in malli core, e.g. regexs as string for both Clojure and ClojureScript. Ideas on how to follow this up? Split is up? new impl by @frenchy64?

opqdonut:
Just a note: we also have `:re` schemas.

bsless:
@opqdonut the main idea behind this PR was avoiding `:re` schemas and using native string methods directly. 
As you can see there's a pretty large performance difference and they are also semantically clearer imo than a regex (now I have two problems)

Issue #554, opened 2021-10-26
me/-resolve-root-error does not respect :error/path


tomasd:
If using validation functions and specifying `:error/path`, this is not taken into consideration when using `me/-resolve-root-error`:

```clojure
(-> [:and [:map
           [:a :int]]
     [:fn {:error/path [:a]}
      (constantly false)]]
    (m/explain {:a 1})
    (me/humanize {:resolve me/-resolve-root-error}))
=> [nil]
```

It works with `me/-resolve-direct-error`:

```clojure
(-> [:and [:map
           [:a :int]]
     [:fn {:error/path [:a]}
      (constantly false)]]
    (m/explain {:a 1})
    (me/humanize {:resolve me/-resolve-direct-error}))
=> {:a ["unknown error"]}
```

ikitommi:
`me/-resolve-root-error` is a non-documented feature, but it's still a bug.

PR #589, opened 2021-12-06
Support :fn with :error/path for me/-resolve-root-error

tomasd:
This PR fixes https://github.com/metosin/malli/issues/554


ikitommi:
thanks!

Issue #482, opened 2021-07-22
malli.instrument should work with cljs


ikitommi:
clj-kondo file emitting & var-lookups work just now with clj.

jeaye:
Is there any estimate for when this would be implemented? I'm currently looking to port a CLJS project from spec to malli to take advantage of some of its features and we're using instrumentation/Orchestra quite heavily.

ikitommi:
I'm not at least working on this, looking forward to someone's PR :)

ikitommi:
merged in master 🥳

PR #604, opened 2021-12-27
Add instrumentation and clj-kondo support for ClojureScript

dvingo:
Addresses https://github.com/metosin/malli/issues/482 

Implements function instrumentation for ClojureScript code, attempting
to match as closely as possible to the Clojure implementation.

Adds separate namespaces to implement instrumentation support in cljs
because macros are required to dynamically change vars in cljs
(symbol literals are required at runtime).

malli.dev.cljs/start! does not feature a watch on the function schemas
atom because the var names need to be known at compile time in order
for a macro to produce valid cljs code. In cljs, hot code reloading
is the norm so this isn'ta major deficiency.

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -1,4 +1,5 @@
 (ns malli.clj-kondo
+  #?(:cljs (:require-macros [malli.clj-kondo]))
   (:require #?(:clj [clojure.java.io :as io])
             [fipp.edn :as fipp]
             [malli.core :as m]))
@@ -175,15 +176,18 @@
   ([] (collect nil))
   ([ns]
    (let [-collect (fn [k] (or (nil? ns) (= k (symbol (str ns)))))]
-     (->> (for [[k vs] (m/function-schemas) :when (-collect k) [_ v] vs v (from v)] v)))))
+     (for [[k vs] (m/function-schemas) :when (-collect k) [_ v] vs v (from v)] v))))
 
 (defn linter-config [xs]
   (reduce
     (fn [acc {:keys [ns name arity] :as data}]
       (assoc-in
-        acc [:linters :type-mismatch :namespaces (symbol (str ns)) name :arities arity]
+        acc [:linters :type-mismatch :namespaces ns name :arities arity]
         (select-keys data [:args :ret :min-arity])))
     {:linters {:unresolved-symbol {:exclude ['(malli.core/=>)]}}} xs))
 
 #?(:clj
    (defn emit! [] (-> (collect) (linter-config) (save!)) nil))
+
+#?(:clj
+   (defmacro emit-cljs! [] (emit!)))
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2387,6 +2387,9 @@
      (let [name' `'~(symbol (str name))
            ns' `'~(symbol (str *ns*))
            sym `'~(symbol (str *ns*) (str name))]
+       ;; in cljs we need to register the schema in clojure (the cljs compiler) so it is visible in the -function-schemas* map at macroexpansion time.
+       (when (some? (:ns &env))
+         (-register-function-schema! (symbol (str *ns*)) name value (meta name)))
        `(do (-register-function-schema! ~ns' ~name' ~value ~(meta name)) ~sym))))
 
 (defn -instrument
--- a/src/malli/dev/cljs.clj
+++ b/src/malli/dev/cljs.clj
@@ -1,0 +1,27 @@
+(ns malli.dev.cljs
+  (:require
+    [malli.clj-kondo :as clj-kondo]
+    [malli.instrument.cljs :as mi]))
+
+(defmacro stop!
+  "Stops instrumentation for all functions vars and removes clj-kondo type annotations."
+  []
+  `(do
+     ~(mi/unstrument* nil)
+     ~(do (clj-kondo/save! {}) nil)))
+
+(defmacro start!
+  "Collects defn schemas from all loaded namespaces and starts instrumentation for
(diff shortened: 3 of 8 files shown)


src/malli/clj_kondo.cljc
```diff
@@ -175,15 +176,18 @@
...
       (assoc-in
-        acc [:linters :type-mismatch :namespaces (symbol (str ns)) name :arities arity]
+        acc [:linters :type-mismatch :namespaces ns name :arities arity]
         (select-keys data [:args :ret :min-arity])))
     {:linters {:unresolved-symbol {:exclude ['(malli.core/=>)]}}} xs))
 
 #?(:clj
    (defn emit! [] (-> (collect) (linter-config) (save!)) nil))
+
+#?(:clj
```
ikitommi, line 192: why is this needed?

src/malli/dev/cljs.clj
```diff
@@ -0,0 +1,27 @@
...
+
+(defmacro start!
+  "Collects defn schemas from all loaded namespaces and starts instrumentation for
+   a filtered set of function Vars (e.g. `defn`s). See [[malli.core/-instrument]] for possible options.
+   Also emits clj-kondo type annotations.
+   This does NOT re-instrument functions if the function schemas change - use hot reloading to get a similar effect."
+  [options]
+  `(do
+     ~(mi/unstrument* nil)
+     ~(do (clj-kondo/save! {}) nil)
```
ikitommi, line 21: why ensure a`nil` return here?

ikitommi (approved):
Looks great! Don't fully understand what happens at `malli.instrument.cljs`, but big thanks for making this!

src/malli/dev/cljs.clj
dvingo: No reason in particular - I can just have it return whatever `(clj-kondo/save! {})` returns.

src/malli/clj_kondo.cljc
dvingo: This will allow you to emit the kondo config from a ClojureScript REPL because CLJS will need to call the Clojure code in this namespace from a macro.

ikitommi:
Awesome, thanks!!

src/malli/core.cljc
```diff
@@ -2387,6 +2387,9 @@
      (let [name' `'~(symbol (str name))
            ns' `'~(symbol (str *ns*))
            sym `'~(symbol (str *ns*) (str name))]
+       ;; in cljs we need to register the schema in clojure (the cljs compiler) so it is visible in the -function-schemas* map at macroexpansion time.
+       (when (some? (:ns &env))
+         (-register-function-schema! (symbol (str *ns*)) name value (meta name)))
```
ikitommi, line 2392: Because of this change, this doesn't work now:

```clj
(def small-int [:int {:max 6}])

(defn plus [x] (inc x))
(m/=> plus [:=> [:cat :int] small-int])
```

... as the `small-int` is not resolved at macro-expansion time and the `small-int` is not resolved yet.

any ideas?

src/malli/instrument/cljs.clj
```diff
@@ -0,0 +1,156 @@
...
+
+;;
+;; instrument
+;;
+
+(defn emit-instrument-fn [{:keys [gen filters] :as instrument-opts} {:keys [schema] :as schema-map} ns-sym fn-sym]
+  ;; gen is a function
+  (let [schema-map (assoc (select-keys schema-map [:gen :scope :report])
+                     ;; At macroexpansion time the schema will be a JVM object, this converts it to a JS object.
+                     :schema `(m/function-schema ~(m/form schema)))
```
ikitommi, line 15: these too happen at macroexpansion time, so will fail with local variables.

src/malli/instrument/cljs.clj
```diff
@@ -0,0 +1,156 @@
...
+  ([opts] (instrument* opts)))
+
+;;
+;; unstrument
+;;
+
+(defn emit-unstrument-fn [{:keys [schema filters] :as opts} ns-sym fn-sym]
+  (let [opts (assoc (select-keys opts [:gen :scope :report])
+               ;; At macroexpansion time the schema will be a JVM object, this converts it to a JS object.
+               :schema `(m/function-schema ~(m/form schema)))
```
ikitommi, line 59: here too.

ikitommi: calling `m/form` also removes the possibility to use non-registered Schema Instances:

```clj
(def MyInt (m/-simple-schema {:type 'MyInt, :pred int?}))

(m/validate MyInt 2) ; => true

(m/form MyInt) ; => MyInt

;; there is no way to construct the local instance from form.
(m/schema (m/form MyInt))
; =throws=> :malli.core/invalid-schema {:schema MyInt}
```

I'm testing if we can just use `schema` in the three places.

ikitommi: with my skills, we can't.

dvingo: ahh that's unfortunate! I will use these examples to see if I can get it working. I _think_ moving everything to happen at runtime (in cljs) should be possible. I will report back when I have an update.

ikitommi: awesome, looking forward to it!

dvingo:
@kovasap  you're totally correct - this wasn't fully complete! 

I just pushed a PR to fix it up https://github.com/metosin/malli/pull/655

thanks for reporting that.

kovasap:
Thanks!  And with this it seems like the documentation at https://github.com/metosin/malli/blob/master/docs/function-schemas.md#development-instrumentation is sufficient.

dvingo:
I was just looking at that - the only piece that doesn't apply is
> and starts watching the function registry for changes. 

but the doc string for `start!` in dev.cljs calls this out.

PR #606, opened 2021-12-29
malli.destructure

ikitommi:
# What

1. Support for Clojure Binding syntax (destructuring)
2. Support for Plumatic-style schematized binding syntax (destructuring)

## Public API

```clj
(defn parse
  "Takes a destructuring bindings vector (arglist)
   and returns a map with keys:

   | key            | description |
   | ---------------|-------------|
   | `:raw-arglist` | the original arglist (can have type-hints)
   | `:arglist`     | simplified clojure arglist (no type-hints)
   | `:schema`      | extracted malli schema
   | `:parsed`      | full parse results

   Parsing can be configured using the following options:

   | key                    | description |
   | -----------------------|-------------|
   | `::md/inline-schemas`  | support plumatic-style inline schemas (true)
   | `::md/sequential-maps` | support sequential maps in non-rest position (true)
   | `::md/references`      | qualified schema references used (true)
   | `::md/required-keys`   | destructured keys are required (false)
   | `::md/closed-maps`     | destructured maps are closed (false)

   Examples:

      (require '[malli.destructure :as md])

      (-> '[a b & cs] (md/parse) :schema)
      ; => [:cat :any :any [:* :any]]

      (-> '[a :- :string, b & cs :- [:* :int]] (md/parse) :schema)
      ; => [:cat :string :any [:* :int]]"
  ...)
```

## 1) Clojure Binding syntax

```clj
(require '[malli.destructure :as md])

(-> '[a [b c & rest :as bc]
      & {:keys [d e]
         :demo/keys [f]
         g :demo/g
         :or {d 0}
         :as opts}]
    (md/parse)
    (select-keys [:arglist :schema]))
;{:arglist [a [b c & rest :as bc]
;           & {:keys [d e]
;              :demo/keys [f]
;              g :demo/g
;              :or {d 0}
;              :as opts}],
; :schema [:cat
;          :any
;          [:maybe [:cat [:? :any] [:? :any] [:* :any]]]
;          [:altn
;           [:map
;            [:map
;             [:d {:optional true} :any]
;             [:e {:optional true} :any]
;             [:demo/f {:optional true}]
;             [:demo/g {:optional true}]]]
;           [:args
;            [:*
;             [:alt
;              [:cat [:= :d] :any]
;              [:cat [:= :e] :any]
;              [:cat [:= :demo/f] :demo/f]
;              [:cat [:= :demo/g] :demo/g]
;              [:cat :any :any]]]]]]}
```

## 2) Plumatic-style schematized binding syntax

```clj
(-> '[a :- :int,
      [b :- :boolean
       c
       & rest :- [:* :string]
       :as bc]
      & {:keys [d e]
         :demo/keys [f]
         g :demo/g
         :or {d 0}
         :as opts}]
    (md/parse)
    (select-keys [:arglist :schema]))
;{:arglist '[a [b c & rest :as bc]
;            & {:keys [d e]
;               :demo/keys [f]
;               g :demo/g
;               :or {d 0}
;               :as opts}],
; :schema [:cat
;          :int
;          [:maybe [:cat [:? :boolean] [:? :any] [:* :string]]]
;          [:altn
;           [:map
;            [:map
;             [:d {:optional true} :any]
;             [:e {:optional true} :any]
;             [:demo/f {:optional true}]
;             [:demo/g {:optional true}]]]
;           [:args
;            [:*
;             [:alt
;              [:cat [:= :d] :any]
;              [:cat [:= :e] :any]
;              [:cat [:= :demo/f] :demo/f]
;              [:cat [:= :demo/g] :demo/g]
;              [:cat :any :any]]]]]]}
```

### DIFF
--- a/test/malli/destructure_test.cljc
+++ b/test/malli/destructure_test.cljc
@@ -37,9 +37,12 @@
     :bind '[a {:keys [b]
                :strs [c]
                :syms [d]
-               :demo/keys [e]
-               :demo/syms [f]
-               :or {b 0, d 0, f 0} :as map}]
+               :demo/syms [e]
+               :demo/keys [f]
+               g :demo/g
+               h 123
+               :or {b 0, d 0, f 0}
+               :as map}]
     :schema [:cat
              :any
              [:altn
@@ -47,16 +50,20 @@
                      [:b {:optional true} :any]
                      ["c" {:optional true} :any]
                      ['d {:optional true} :any]
-                     [:demo/e {:optional true} :any]
-                     ['demo/f {:optional true} :any]]]
+                     ['demo/e {:optional true} :any]
+                     [:demo/f {:optional true}]
+                     [:demo/g {:optional true}]
+                     [123 {:optional true} :any]]]
               [:args [:schema
                       [:*
                        [:alt
                         [:cat [:= :b] :any]
                         [:cat [:= "c"] :any]
                         [:cat [:= 'd] :any]
-                        [:cat [:= :demo/e] :any]
-                        [:cat [:= 'demo/f] :any]
+                        [:cat [:= 'demo/e] :any]
+                        [:cat [:= :demo/f] :demo/f]
+                        [:cat [:= :demo/g] :demo/g]
+                        [:cat [:= 123] :any]
                         [:cat :any :any]]]]]]]
     :errors '[[{::keysz [z]}]
               [{:kikka/keyz [z]}]]}
@@ -67,32 +74,17 @@
              [:altn
               [:map [:map
                      [:a :any]
-                     [:demo/b :any]
-                     [:demo/c :any]]]
+                     :demo/b
+                     :demo/c]]
               [:args [:schema [:* [:alt
                                    [:cat [:= :a] :any]
-                                   [:cat [:= :demo/b] :any]
-                                   [:cat [:= :demo/c] :any]
+                                   [:cat [:= :demo/b] :demo/b]
+                                   [:cat [:= :demo/c] :demo/c]
                                    [:cat :any :any]]]]]]]}
    {:name "map destructuring with required-keys and closed-maps"
     :bind '[{:keys [a :demo/b] :demo/keys [c]}]
     :options {::md/required-keys true
               ::md/closed-maps true}
-    :schema [:cat
-             [:altn
-              [:map [:map {:closed true}
-                     [:a :any]
-                     [:demo/b :any]
-                     [:demo/c :any]]]
-              [:args [:schema [:* [:alt
-                                   [:cat [:= :a] :any]
-                                   [:cat [:= :demo/b] :any]
-                                   [:cat [:= :demo/c] :any]]]]]]]}
-   {:name "map destructuring with required-keys, closed-maps and references"
-    :bind '[{:keys [a :demo/b] :demo/keys [c]}]
-    :options {::md/required-keys true
-              ::md/closed-maps true
-              ::md/references true}
     :schema [:cat
              [:altn
               [:map [:map {:closed true}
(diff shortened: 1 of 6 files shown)


ikitommi:
Mostly DONE, could add an option to pull the namespaced keys from registry instead of attaching those to `:any`.

Deraen:
Map keys in the schema refer to symbols in the binding forms, maybe other parts should also refer to the names? Like use `:catn` instead of `:cat`?

In this case `:args` could be `:opts` but not sure if that makes it harder to parse the result, and what to do if the binding form doesn't use `:as`...

ikitommi:
Good point. I think using `:catn` would yield superior error messages, and better docs.

test/malli/destructure_test.cljc
```diff
@@ -0,0 +1,316 @@
...
+             :any
+             [:altn
+              [:map [:map
+                     [:b {:optional true} :any]
+                     ["c" {:optional true} :any]
+                     ['d {:optional true} :any]
+                     ['demo/e {:optional true} :any]
+                     [:demo/f {:optional true}]
+                     [:demo/g {:optional true}]
+                     [123 {:optional true} :any]]]
```
Deraen, line 56: Hmm, okay. The key names are the keys from the map instead of the symbols they are being bound into. Makes sense now that I think about this, as the schema should describe the value.

ikitommi: yes, the new `MapLike` schema groups the keys now by behaviour, so can't group by key anymore. 

Related, one can bind same symbol many times, making the binding... weird:

```clj
(let [{[a a] :a
       [b a] :b
       [c a] :c
       [d a] :d
       [e a] :d
       [f a] :d
       [g a] :d
       [h a] :d}
      {:a [1 2]
       :b [3 4]
       :c [5 6]
       :d [7 8]
       :e [9 10]
       :f [11 12]
       :g [13 14]
       :h [15 16]}] a)
; => ???
```

NoahTheDuke: If you meant to use the same key `:d` for all of the right hand side, then it's 8. If you meant to use each of the keys, then it's 16. They're assigned in the order written because destructuring happens at macro expansion time.

ikitommi: oh, there was a bug in the code, too many `:d`s. Here's a example that highlights that this looks like undefined behavior (the map changes the impl, where we lose the key order:

```clj
(let [{[a a] :a
       [b a] :b
       [c a] :c
       [d a] :d
       [e a] :e
       [f a] :f
       [g a] :g
       [h a] :h
       [i a] :i}
      {:a [1 2]
       :b [3 4]
       :c [5 6]
       :d [7 8]
       :e [9 10]
       :f [11 12]
       :g [13 14]
       :h [15 16]
       :i [16 17]}] a)
; => 8


(let [{[a a] :a
       [b a] :b
       [c a] :c
       [d a] :d
       [e a] :e
       [f a] :f
       [g a] :g
       [h a] :h
       #_#_[i a] :i}
      {:a [1 2]
       :b [3 4]
       :c [5 6]
       :d [7 8]
       :e [9 10]
       :f [11 12]
       :g [13 14]
       :h [15 16]
       :i [16 17]}] a)
; => 16

(class '{[a a] :a
         [b a] :b
         [c a] :c
         [d a] :d
         [e a] :e
         [f a] :f
         [g a] :g
         [h a] :h
         #_#_[i a] :i})
; => clojure.lang.PersistentArrayMap

(class '{[a a] :a
         [b a] :b
         [c a] :c
         [d a] :d
         [e a] :e
         [f a] :f
         [g a] :g
         [h a] :h
         [i a] :i})
; => clojure.lang.PersistentHashMap
```

ikitommi:
I think this is a good alpha-quality namespace, with lot of tests, so merging it in.

ikitommi:
post-comments still most welcome, will not release this now as I had to change how `m/-property-registry` works, which is a **BREAKING** change, so need to ship minor out of this. 

It's a fix, but theoretically someone could rely on the property registry containing forms, not schemas.

PR #611, opened 2022-01-08
Construct schema objects at runtime when instrumenting cljs fns

dvingo:
Addresses the errors that came up in the prior cljs instrumentation PR:
https://github.com/metosin/malli/pull/604

This solution involves adding a new atom to contain the cljs vars that contain function schemas. 

I think we could have one atom for both clj + cljs if they don't invoke `m/function-schema` when registering the function schema,
and then in clojure, when instrumenting here: (https://github.com/metosin/malli/blob/80775e74eaf191f3916649ccbeae095dcfad62e9/src/malli/instrument.clj#L26)
updating the schema to check that it is a function schema at that point.

for example:
```clojure
(m/-instrument (update dgen :schema m/function-schema) original-fn)
```

what do you think?

*update*: I think the intention of throwing on invalid fn schema would be lost though for clojure usage as the user would only find out when/if they instrument the function.
I think this dual setup of one atom for cljs and one for clj may be the most straightforward, but maybe you have some ideas.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2370,7 +2370,9 @@
 ;;
 
 (defonce ^:private -function-schemas* (atom {}))
+(defonce ^:private -function-schemas-cljs* (atom {}))
 (defn function-schemas [] @-function-schemas*)
+(defn function-schemas-cljs [] @-function-schemas-cljs*)
 
 (defn function-schema
   ([?schema]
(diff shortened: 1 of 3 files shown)


ikitommi:
oh, this is beautiful. Don't have an opinion yet on the number of atoms, but this solves the issue, merging it. Big thanks for resolving this quickly!

ikitommi:
One thing I noticed, the tests do not print the "..unstrumented" into the logs despite it clearly unstruments the functions. Idea's why is that not happening?

ikitommi:
and, I moved this to use single atom with `:clj` or `:cljs` as top-level keys to differentiate those, also did some re-organising the code in `malli.instrument.cljs` in the plumatic branch, merging it back to master soon.

dvingo:
Very nice!! 

that's weird about the print statements not showing up for unstrument. I'll take a look.

dvingo:
@ikitommi found the issue - classic macro problem, I had a typo here, and forgot a quote:
https://github.com/metosin/malli/blob/4f38a081dc25285881f49aba750a36b1b225333a/src/malli/instrument/cljs.clj#L76
this line should be:
```clojure
`(when (some #(% '~ns-sym (var ~fn-sym) ~opts) ~filters)
```
The unstrumentation didn't actually happen in the tests. 

I didn't catch this before because I only tried plain unstrument during dev with no filters..

If you reverse the testing blocks in `malli.instrument.cljs-test` so that unstrumentation happens after instrumentation the tests will fail.

I figure it will be faster for you to make the change, so I won't make a PR to fix it.

PR #614, opened 2022-01-10
Allow custom options for `mx/defn` and custom type emitting

pfeodrippe:
I was trying to use custom types in `mx/defn` and was not able to, so
here I'm just passing `:malli/options` metadata which has the
custom options (with some `:registry`). Also, I'm checking if there is
some `prepost` instead of outputting `nil`.

`m/=>` was not working to namespaced symbols (maybe this was
intentional?), so this PR allows it to pass it, altough it has to be
fully qualified.

When emitting the configuration to `clj-kondo` for some custom
type, it was returning `:any`, but now it's outputting a type which the
user-defined in the type properties at the `:clj-kondo/type` key.

Let me know if it makes sense o/

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -133,7 +133,9 @@
 (defmethod accept :union [_ schema _ options] (transform (m/deref schema) options))
 (defmethod accept :select-keys [_ schema _ options] (transform (m/deref schema) options))
 
-(defn- -walk [schema _ children options] (accept (m/type schema) schema children options))
+(defn- -walk [schema _ children options]
+  (or (:clj-kondo/type (m/type-properties schema))
+      (accept (m/type schema) schema children options)))
 
 (defn -transform [?schema options] (m/walk ?schema -walk options))
 
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2396,17 +2396,18 @@
 (defn -register-function-schema!
   ([ns name ?schema data] (-register-function-schema! ns name ?schema data :clj function-schema))
   ([ns name ?schema data key f]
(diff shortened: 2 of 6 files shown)


ikitommi:
Hi. This PR has many parts:

> I was trying to use custom types in mx/defn and was not able to, so
here I'm just passing :malli/options metadata which has the
custom options (with some :registry). Also, I'm checking if there is
some prepost instead of outputting nil.

You can use schemas as vars or inject the schemas into global registry, e.g.

## Vars

```clj
(def UserId :int)
(def User [:map
           [::user-id UserId]
           [:name :string]
           [:age :int]])

(mx/defn get-user :- User [id :- UserId]
  (assoc (mg/generate ::user) ::user-id id))
```

## Global

```clj
(mr/set-default-registry!
 (mr/composite-registry
  (m/default-schemas)
  {::user-id :int
   ::user [:map
           ::user-id
           [:name :string]
           [:age :int]]}))

(mx/defn get-user :- ::user [id :- ::user-id]
  (assoc (mg/generate ::user) ::user-id id))
```

not happy yet adding support for custom registries for `mx/defn` at this time.

> m/=> was not working to namespaced symbols (maybe this was
intentional?), so this PR allows it to pass it, altough it has to be
fully qualified.

Could you show a code sample where it doesn't work?

> When emitting the configuration to clj-kondo for some custom
type, it was returning :any, but now it's outputting a type which the
user-defined in the type properties at the :clj-kondo/type key.

Indeed! I Would appreciate a PR with just this. It should look from both `m/type-properties` and `m/properties` like JSON Schema does, e.g.:

```clj
 (let [p (merge (m/type-properties schema) (m/properties schema))]
     ....
```.

pfeodrippe:
Got it, @ikitommi. I will try something here and will open another PR if the case. Thanks!

PR #617, opened 2022-01-12
Fix quote in unstrument macro for cljs

dvingo:
fixes unstrument filtering for cljs, as mentioned here:

https://github.com/metosin/malli/pull/611


ikitommi:
thanks!

Issue #622, opened 2022-01-21
Add support for malli.experimental/defn- 


FieryCod:
Looking at the source code I can spot the `malli.experimental` namespace supports definitions of public functions. 
Would be cool to have a `defn-` variant to support definitions of private functions.

ikitommi:
wouldn't this be sufficient:

```clj
(mx/defn ^:private kakka :- :int
  "inline schemas (plumatic-style)"
  [x :- :int] (inc x))
```

Some consider `defn-` as bad practice.

FieryCod:
`defn-` is already in core. I see no reason why it shouldn't be supported. I do also use it extensively, but I understand some people may not like it. I will leave the decision to you.

ikitommi:
After thinging about this, to keep the public api smaller, not going to add this. There will be schematized `def` too, and no `def-`.

PR #623, opened 2022-01-21
Add defn- in malli.experimental

FieryCod:
Related to issue: https://github.com/metosin/malli/issues/622


ikitommi:
After thinking about this, not going to add this to keep the public API smaller.  Thanks anyway!

```clj
(mx/defn ^:private kakka :- :int
  "inline schemas (plumatic-style)"
  [x :- :int] (inc x))
```

PR #630, opened 2022-01-26
fix: The malli pretty printer (virhe) is failing when it tries to print a datomic dbwhen it tries to pr

### LINKED ISSUE #629: The malli pretty printer (virhe) is failing when it tries to print a datomic db.
Starting `malli.dev` with
```
(malli.dev/start! {:report (pretty/reporter)})
```

Take for example this function:
```
(defn fetch-user
  "Fetch a user by its entity `id`. If no entity is found, nil is returned."

  {:malli/schema
   [:=> [:cat :int :any]
    [:maybe :map]]}

  [id db]

  ...)
```

The `db` (second) parameter is defined as `:any` - I'm not even interested in checking that it is a valid db right now, I just want to ensure the id is of valid type.

When calling the function with a valid `id` (i.e. a `:int`) and `db`: 
```
(fetch-user 96757023244461 db)
```
it works as expected.  However when calling it with a string or anything else in the first position (id), it results in this:
```
(fetch-user "96757023244461" (db/current))

Execution error at malli.dev.virhe.EdnPrinter/visit_set (virhe.cljc:80).
Unable to convert: class datomic.core.btset.BTSet to Object[]
```

i.e. it is not printing the malli error as it should, but instead it fails with the above.

Passing anything else in the second argument, and still the (incorrect type) string in first argument, prints the malli error as expected:
```
-- Schema Error ---------------------------------------------------

Invalid function arguments:

  ["96757023244461" nil]

Input Schema:

  [:cat :int :any]

Errors:

  {:in [0],
   :message "should be an integer",
   :path [0],
   :schema :int,
   :type nil,
   :value "96757023244461"}
```

I have attached the stacktrace for when the error can not be pretty printed (i.e. when an actual datomic db is passed).

Presumably it is traversing the db object's fields and fails when it gets to the `datomic.core.btset.BTSet` and tries to convert it to an array.

It's not clear to me why datomic's `BTSet` is failing at that point, and it is probably not a defect in `virhe`.

Wrapping the argument `x` in a `seq` in `EdnPrinter`'s `visit-set`, resolves the issue:

```
(visit-set [this x]
    (let [xs (sort-by identity (fn [a b] (arrangement.core/rank a b)) ___(seq x)___)] 
      (fipp.edn/pretty-coll this "#{" xs :line "}" fipp.visit/visit)))      
```

I will raise a pull-request for this change, if you could please consider it. 

[stacktrace.txt](https://github.com/metosin/malli/files/7942459/stacktrace.txt)

hti-andre:
Fixes #629

### DIFF
--- a/src/malli/dev/virhe.cljc
+++ b/src/malli/dev/virhe.cljc
@@ -75,7 +75,7 @@
                               [:span (fipp.visit/visit printer k) " " (fipp.visit/visit printer v)]))))
 
   (visit-set [this x]
-    (let [xs (sort-by identity (fn [a b] (arrangement.core/rank a b)) x)]
+    (let [xs (sort-by identity (fn [a b] (arrangement.core/rank a b)) (seq x))]
       (fipp.edn/pretty-coll this "#{" xs :line "}" fipp.visit/visit)))
 
   (visit-tagged [this {:keys [tag form]}]


ikitommi:
thanks!

PR #634, opened 2022-02-04
Replace one children :and with properties with :fn

Deraen:


ikitommi:
`[:fn string?]` is not a real schema, it does not transform, convert to JSON Schema etc. These would be more idiomatic, in the order of idiomacy:

```clj
string?
-> [:schema {} string?]
-> [string? {}]

:string
-> [:string {}]
-> [:schema {} :string]
```

Deraen:
Argh. The difference is really confusing.

Deraen:
`[string? {}]` is also bad example because it takes different properties to `[:string ...]`

ikitommi:
yes, the `string?`s should go, sugar leading to a dead-end.

PR #637, opened 2022-02-07
Accept UUIDs up to version 8 when decoding from a string

dvingo:
Hi Tommi, I am using sequential UUIDs on a project (using this library https://github.com/yetanalytics/colossal-squuid) and noticed they were not being coerced to UUIDs in ClojureScript. This PR updates the regex that accepts a valid UUID string on cljs to work up to version 8.


ikitommi:
Good to know, thanks!

Issue #642, opened 2022-02-15
qualified-keyword with namespace doesn't generate the ns


Invertisment:
Let's say I have this schema:

```
[:map {}
  [:id/house [:qualified-keyword {:namespace :house}]]
  [:house/address string?]]
```

If I generate the value I get this output:
```
{:id/house :a.../IJ9-7RW
 :house/address "EHL2302t6Afp79tX79A"}
```

But if I try to validate it I get a `false`.
But if I change the generated value into this then it passes:

```
{:id/house :house/IJ9-7RW
 :house/address "EHL2302t6Afp79tX79A"}
```

I think that the generated value should be the same as the fixed value.
i.e. the namespace of the generated keyword should be `house`.

PR #643, opened 2022-02-15
Qualified keyword with preconfigured namespace

Invertisment:
Closes https://github.com/metosin/malli/issues/642

Got some help from `p-himik`: https://clojurians.slack.com/archives/C03S1KBA2/p1644919467267179
and `borkdude`.

In this PR I also did update the generator for symbols but it doesn't validate the namespace and I didn't touch the validator of symbols.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -14,6 +14,10 @@
 (defprotocol Generator
   (-generator [this options] "returns generator for schema"))
 
+;;
+;; generators
+;;
+
 (defn- -random [seed] (if seed (random/make-random seed) (random/make-random)))
 
 (defn -recur [schema {::keys [recursion recursion-limit] :or {recursion-limit 4} :as options}]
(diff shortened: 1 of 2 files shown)


src/malli/generator.cljc
```diff
@@ -158,6 +158,18 @@
 ;; generators
 ;;
 
+(defn- qualified-keyword-generator [schema]
```
ikitommi, line 161: could this be `-qualified-ident-gen` with +3 args? The two functions are basically the same.

Invertisment: Sure. After I created this PR I found that you have some code before `generators` comment. I was confused initially what that meant to I've put my generator below it. Because it... marks the generators. 
I'll do the change.
I first wanted to make sure this change makes sense.

Invertisment: I did some code reuse commits but I don't like where it went. I'm not sure if it's more readable than before.
I prefer the version from commit [817652f](https://github.com/metosin/malli/pull/643/commits/817652f5d4ce588af38e46fab4b969587fe46f1f).

ikitommi:
thanks for this, merging now :)

ikitommi:
https://github.com/metosin/malli/commit/38d899f8a2d3b56d25b26feea437bb79f601c142 + `[metosin/malli "0.8.3"]`

PR #655, opened 2022-02-27
Fix malli.dev.cljs implementation

dvingo:
The dev code wasn't actually available in cljs as reported here: https://github.com/metosin/malli/pull/604#issuecomment-1052544869

This PR fixes that.

There was also an annoyance when instrumented vars only in some cljs builds - like in tests and in a main application build,
I added a check to prevent compilation warnings about unused vars.

- Adds a guard to check var existence before emitting instrumentation cljs code
- Make malli.dev.cljs available to cljs runtimes.


ikitommi:
Thanks!!!

Issue #659, opened 2022-03-06
Interpreting CLJS invalid schema errors


kovasap:
I got this error today, and I'm very confused about what it means or how to resolve it:

![image](https://user-images.githubusercontent.com/8763010/156904219-f8d1e95e-77c7-4876-953a-8045c377ab23.png)

Any ideas?

kovasap:
(I have no `nil` valued schemas in my project)

kovasap:
This error can be reproduced with https://github.com/kovasap/malli-errors-repro

kovasap:
This seemed to be fixed by using this code: https://github.com/metosin/malli/issues/654#issuecomment-1060861408, but it was very difficult to find out why.

PR #660, opened 2022-03-08
Fully qualify symbols used in cljs metadata schemas

dvingo:
This PR ensures all vars used in cljs schemas are fully qualified.  This is necessary because the namespace where `collect!` is invoked will be the one where the generated code from the macro is produced, which likely won't have those vars available.

This should fix the issue https://github.com/metosin/malli/issues/659


ikitommi:
Thank You.

kovasap:
One potential issue with this PR (not sure if it's intended or not) is that predicate functions like `float?` and `int?` can't be used in schemas - when I try, I get an error that `my-namespace/float?` doesn't exist.

PR #662, opened 2022-03-09
Remove clj-kondo config output for cljs schemas

dvingo:
I have been trying this out and unfortunately realized that kondo support will not be so simple.

If the cljs schemas use vars that are only available during the cljs runtime
we do not have access to them during macroexpansion and thus will cause exceptions.

This line https://github.com/metosin/malli/blob/303c6062f02a2beed62e432c1a1a1046cfa1e288/src/malli/clj_kondo.cljc#L161
will attempt to construct a malli schema and if there any symbols that represent clojurescript variables then there is no way to know their values in clojure.

For now I think removing the kondo support makes sense to allow the instrumentation to work.

As an alternative, we could potentially add clojurescript code that will print out the kondo config to the javascript console when all the vars are available and the user can copy it to the condo config file. I know it's not a great solution but at least provides some value.

_update_: I added a printer helper for cljs.

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -189,16 +189,16 @@
 #?(:clj
    (defn emit! [] (-> (collect) (linter-config) (save!)) nil))
 
-#?(:clj
-   (defn collect-cljs
-     ([] (collect-cljs nil))
-     ([ns]
-      (let [-collect (fn [k] (or (nil? ns) (= k (symbol (str ns)))))]
-        (for [[k vs] (m/function-schemas :cljs) :when (-collect k) [_ v] vs v (from v)] v)))))
+(defn collect-cljs
+  ([] (collect-cljs nil))
+  ([ns]
+   (let [-collect (fn [k] (or (nil? ns) (= k (symbol (str ns)))))]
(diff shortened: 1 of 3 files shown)


kovasap:
Could you update the documentation to describe how this copy-pasting would work?

Also, is there anything I can do to help get this merged?

ikitommi:
Bummer, I really would love the get the clj-kondo to work. Merging in.

ikitommi:
Could we uses `delay` or similar? just register something that will be later available. when `dev/start!` is called, everything should be ready, right?

dvingo:
I'm not sure how to make this happen. The fundamental problem is that the kondo config is emitted at compile (macro-expansion) time, but the values needed to construct that config are only available at runtime. 
For JS running in a browser, there is no access to the filesystem, so some other ideas would be:
- add a browser js function that will trigger a file download dialog containing the kondo config, a user would run this from a REPL.
- add a function that copies the kondo config to your clipboard (just saves the step a user would have to do with the current console.log helper)
- implement kondo file emission using node.js - but I'm guessing most users will be targeting browsers so this may cause its own problems when browser APIs aren't available in node environments.

riotrah:
Hi, I'm fairly new to cljs, but are there docs for this manual copying thing?

PR #665, opened 2022-03-17
JSON examples in README.md

opqdonut:
I thought end-to-end value->json->value examples using malli would be
useful. I was especially confused by the fact the encode and decode
can let through invalid data, so I thought I'd document that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -841,6 +841,61 @@ Going crazy:
 ; => {:x 24}
 ```
 
+## To and from JSON
+
+The `m/encode` and `m/decode` functions work on clojure data. To go
+from clojure data to JSON, you need a JSON library like
+[jsonista](https://github.com/metosin/jsonista). Additionally, since
+`m/decode` doesn't check the schema, you need to run `m/validate` (or
+`m/explain`) if you want to make sure your data conforms to your
+schema.
+
+To JSON:
+
+```clj
+(def Tags
+  [:map
+   {:closed true}
+   [:tags [:set :keyword]]])
+(jsonista.core/write-value-as-string
+ (m/encode Tags
+           {:tags #{:bar :quux}}
+           mt/json-transformer))
+; => "{\"tags\":[\"bar\",\"quux\"]}"
+```
+
+From JSON without validation:
+
+```clj
+(m/decode Tags
+          (jsonista.core/read-value "{\"tags\":[\"bar\",[\"quux\"]]}"
+                                    jsonista.core/keyword-keys-object-mapper)
+          mt/json-transformer)
+; => {:foo #{:bar ["quux"]}}
+```
+
+From JSON with validation:
+
+```clj
+(m/explain Tags
+           (m/decode Tags
+                     (jsonista.core/read-value "{\"tags\":[\"bar\",[\"quux\"]]}"
+                                               jsonista.core/keyword-keys-object-mapper)
+                     mt/json-transformer))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -841,6 +841,61 @@ Going crazy:
...
+[jsonista](https://github.com/metosin/jsonista). Additionally, since
+`m/decode` doesn't check the schema, you need to run `m/validate` (or
+`m/explain`) if you want to make sure your data conforms to your
+schema.
+
+To JSON:
+
+```clj
+(def Tags
+  [:map
```
ikitommi, line 857: this could be wrapped into `m/schema` (as a good practice) to enable caching of validators & explainers.

README.md
```diff
@@ -841,6 +841,61 @@ Going crazy:
...
+                     (jsonista.core/read-value "{\"tags\":[\"bar\",[\"quux\"]]}"
+                                               jsonista.core/keyword-keys-object-mapper)
+                     mt/json-transformer))
+; => {:schema [:map {:closed true} [:tags [:set :keyword]]],
+;     :value {:tags #{:bar ["quux"]}},
+;     :errors ({:path [:tags 0], :in [:tags ["quux"]], :schema :keyword, :value ["quux"]})}
+```
+
+```clj
+(m/validate Tags
```
ikitommi, line 891: I would use threading macro here, short lines are easier to read (on mobile)

README.md
```diff
@@ -841,6 +841,61 @@ Going crazy:
...
 ```
 
+## To and from JSON
+
+The `m/encode` and `m/decode` functions work on clojure data. To go
+from clojure data to JSON, you need a JSON library like
+[jsonista](https://github.com/metosin/jsonista). Additionally, since
+`m/decode` doesn't check the schema, you need to run `m/validate` (or
+`m/explain`) if you want to make sure your data conforms to your
+schema.
```
ikitommi, line 851: there is a fast `coercer` function described in https://www.metosin.fi/blog/high-performance-schemas-in-clojurescript-with-malli-1-2/#transformation, would it make sense to built that? pure transformers created ahead-of-time. up to you.

ikitommi (approved):
with comments.

README.md
```diff
@@ -841,6 +841,61 @@ Going crazy:
...
+```
+
+From JSON without validation:
+
+```clj
+(m/decode Tags
+          (jsonista.core/read-value "{\"tags\":[\"bar\",[\"quux\"]]}"
+                                    jsonista.core/keyword-keys-object-mapper)
+          mt/json-transformer)
+; => {:foo #{:bar ["quux"]}}
```
viesti, line 874: Here `:foo` but in goes `"tags"`?

README.md
opqdonut: I'd like to use `->` or `->>` as well, but `m/decode` wants the value as argument 2 out of 3.

README.md
opqdonut: done!

README.md
opqdonut: thanks, I switched from a nonsemantic to a semantic example at some point and forgot to update this

README.md
opqdonut: I'll add something about that.

opqdonut:
I'm waiting for more comments, but intend to merge later today

README.md
ikitommi: `as->` for the win :)

PR #671, opened 2022-03-25
 Add static type checking instructions

frenchy64:
Malli users might be interested in this Typed Clojure feature.


ikitommi:
This is good, thanks!!

PR #674, opened 2022-03-27
Speed up strip-extra-keys-transformer

bsless:
reduce-kv is faster than reduce-ing over a map's keys

Opportunistically found when profiling


ikitommi:
Thanks!!

Issue #675, opened 2022-03-27
`(malli.dev.cljs/start!)` produces very spaced out log messages instead of an error in the browser log


kovasap:
I get this "error":

![image](https://user-images.githubusercontent.com/8763010/160295241-75cac370-3314-418b-ab63-70dad6e8b69a.png)

When running this code:

```
(ns app.core
  (:require
   [reagent.core :as r]
   [reagent.dom :as d]
   [malli.core :as m]
   [malli.instrument.cljs]
   [malli.dev.cljs]
   [malli.dev.pretty :as pretty]))


(def Hiccup [:vector :any])

(defn list-to-hiccup
  "Converts a list of string to hiccup."
  {:malli/schema [:=> [:cat [:sequential :string]] Hiccup]}
  [list]
  5)
  ; [:ul
  ;  (for [item list]
  ;    [:li {:key item} item])])

; (meta #'list-to-hiccup)


(def data ["first" "second" "third"])

(defn home-page []
  (fn []
    [:div [:h2 "My App"]
     [:div "hello world!"]
     (list-to-hiccup data)]))


;; -------------------------
;; Initialize app

(defn mount-root []
  (d/render [home-page] (.getElementById js/document "app")))

(defn ^:dev/after-load refresh []
  (prn "Hot code Remount")
  ; Check all malli function "specs"
  (malli.dev.cljs/start!) ; {:report (pretty/reporter)})
  ; (malli.dev.cljs/collect-all!)
  ; (malli.instrument.cljs/instrument!)
  (mount-root))

(defn ^:export init! []
  ; Check all malli function "specs"
  (malli.dev.cljs/start!) ; {:report (pretty/reporter)})
  ; (malli.dev.cljs/collect-all!)
  ; (malli.instrument.cljs/instrument!)
  (mount-root))
```

Using the 

```
(malli.dev.cljs/collect-all!)
(malli.instrument.cljs/instrument!)
```

lines instead gives me a single self contained error message.

The `{:report (pretty/reporter)}` arg to `start!` seems to have no effect.

kovasap:
@dvingo this might be due to your recent changes?  Interestingly the errors I get via the `collect-all!` and `instrument!` calls are better than those I was getting in https://github.com/metosin/malli/issues/666, so that issue may not be always present...

kovasap:
This behavior appears for me on:

```
metosin/malli {:git/url "https://github.com/metosin/malli.git"
[...]

Issue #666, opened 2022-03-18
Clojurescript schema checking errors are difficult to map to the specific lines of failing code


kovasap:
See https://github.com/metosin/malli/issues/659#issue-1160480472 and https://github.com/metosin/malli/issues/654#issuecomment-1066001501 for an example.  In the Chrome console there is no direct reference to the lines of cljs code that are triggering the errors.  Inlined example:

![image](https://user-images.githubusercontent.com/8763010/158041085-7d2f1e56-09ef-4fb0-a211-66eeca59586b.png)

kovasap:
For people who know malli better, what exactly is different about these errors when compared to those you would get when running in plain Clojure (not CLJS)?  Or are these the same errors?

I'd like to dig into this issue a big more, but am not quite sure what the expected behavior is or otherwise where to start.

dvingo:
@kovasap when you get a chance, check out the latest code that was merged here: 
https://github.com/metosin/malli/pull/681

kovasap:
That PR solves this problem for me.  The current state with that PR is really really good actually, nice job and thank you so much!!

I think part of my problem was that I [didn't enable custom formatters in `cljs-devtools`](https://github.com/binaryage/cljs-devtools/blob/master/docs/faq.md#why-some-custom-formatters-were-not-rendered).

PR #678, opened 2022-03-30
Add a more useful default error reporter for cljs instrumentation

dvingo:
Fipp, the pretty printer, is not designed with the browser console in
mind and the output when developing in cljs is not usable.
(Every print call invokes console.log which will result in a new output line per invocation, 
instead of all on one line like in a terminal).
This commit allows navigating directly to the callsite where the error occurred via a JavaScript stacktrace.

This should address https://github.com/metosin/malli/issues/675 and https://github.com/metosin/malli/issues/666

I believe we would have to create another pretty printer namespace for ClojureScript that mirrors the `malli.dev.pretty` namespace but with the browser console in mind to get useful output.

### DIFF
--- a/src/malli/dev/cljs.cljc
+++ b/src/malli/dev/cljs.cljc
@@ -2,7 +2,6 @@
   #?(:cljs (:require-macros [malli.dev.cljs]))
   #?(:cljs (:require [malli.instrument.cljs]))
   #?(:clj (:require [malli.clj-kondo :as clj-kondo]
-                    [malli.dev.pretty :as pretty]
                     [malli.instrument.cljs :as mi])))
 
 #?(:clj (defmacro stop!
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks! just now read the issues. I think we could wrap the fipp-printing easily.

# just printing

https://github.com/metosin/malli/blob/master/src/malli/dev/pretty.cljc#L75

=> just wrap the results with `with-out-str` and print that using single `js/console.log`

# throwing

https://github.com/metosin/malli/blob/master/src/malli/dev/pretty.cljc#L83

... it seems that the exception don't get pretty printed in console. Maybe just print the message there too (with cljs)

dvingo:
oh right, that's brilliant! 

I am going to close this PR in favor of this one: https://github.com/metosin/malli/pull/681

PR #680, opened 2022-03-31
Pass function metadata to malli.core/-instrument

meditans:
This change extends the third arity of malli.core/-instrument so that it
optionally accepts the metadata of the old function. That metadata is
passed downstream to the `report` function.

This new modality is called in malli.instrument so that when we render
we can know the name (and the source file location) of the offending
function.

### DIFF
--- a/src/malli/instrument.clj
+++ b/src/malli/instrument.clj
@@ -23,7 +23,7 @@
                                           (true? (:gen d)) (dissoc $ :gen)
                                           :else $))]
                          (alter-meta! v assoc ::original-fn original-fn)
-                         (alter-var-root v (constantly (m/-instrument dgen original-fn)))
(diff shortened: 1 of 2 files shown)


src/malli/instrument.clj
```diff
@@ -23,7 +23,7 @@
                                           (true? (:gen d)) (dissoc $ :gen)
                                           :else $))]
                          (alter-meta! v assoc ::original-fn original-fn)
-                         (alter-var-root v (constantly (m/-instrument dgen original-fn)))
+                         (alter-var-root v (constantly (m/-instrument dgen original-fn nil :original-fn-meta (meta v))))
```
opqdonut, line 26: we discussed this briefly with @ikitommi and we'd prefer if you passed in `:meta (meta v)` as part of `dgen`, the first argument of `-instrument`

opqdonut (changes requested):
please also add a test!

PR #681, opened 2022-04-01
Fix fn instrumentation errors for ClojureScript in a browser

dvingo:
- Print all output from the pretty printer in one string
- Add documentation for working on function instrumentation in the malli codebase
- Add documentation for using malli.dev.cljs as a user of malli


ikitommi:
💯!!

kovasap:
This is great!  Thanks so much!

Issue #684, opened 2022-04-04
cljs check should not require the user to require `malli.generator`


dvingo:
This was mentioned in slack, I'm just recording it here as a reminder:

- have the macro call a function in the instrument.cljs runtime namespace, instead of invoking malli.generator directly: https://github.com/metosin/malli/blob/005dfd2868518918b3b1b0ce83f51fbb6b1e4b37/src/malli/instrument/cljs.clj#L128
 so that it is available without the user needing to require it.

PR #685, opened 2022-04-05
cljs instrumentation: ensure malli.generator is available at runtime

dvingo:
To resolve https://github.com/metosin/malli/issues/684

This ensures the client code does not need to `(require [malli.generator])` - the library code will require it instead so the compiler makes it available.


ikitommi:
Thanks!!

PR #690, opened 2022-04-12
Fix cat{n} unparser with invalid value

frenchy64:
The reduce-kv calls need to short-circuit on ::m/invalid.

Discovered by Typed Clojure.


ikitommi:
thanks!

Issue #87, opened 2019-10-14
Warning under advanced compilation


ikitommi:
```
References to the global RegExp object prevents optimization of regular expressions.
malli.core.class_registry = cljs.core.PersistentArrayMap.createAsIfByAssoc([RegExp,malli.core._re_schema(true)]);
                                                                            ^^^^^^
```

tekacs:
It looks like an issue similar to this was fixed in https://github.com/plumatic/schema/issues/106 -- posting here in case I or someone else would like to look at this later

ikitommi:
should be fixed in master

PR #692, opened 2022-04-15
Use JS constructor property to access RegExp

dvingo:
This removes the warning under advanced compilation about accessing the global `RegExp` function from goog closure.

 Addresses: https://github.com/metosin/malli/issues/87

see here for more info:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2262,7 +2262,9 @@
          (reduce -register-var {}))))
 
 (defn class-schemas []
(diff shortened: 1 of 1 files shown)


ikitommi:
Asked about this is many places, thought it's broken by design. Seems not, big thanks!!

dvingo:
I don't fully understand the google closure warning (what exactly cannot be optimized) and wasn't sure this change would work, but I'm glad it did!

Issue #693, opened 2022-04-18
Instrument on ClojureScript?


mauricioszabo:
I'm sorry if I'm misreading something, but so far I was not able to make `instrument!` works on ClojureScript projects. Is it not supported yet, will it be supported, etc? Is there any documentation showing how to do it?

Thanks!

mauricioszabo:
Ok, I _kinda_ figured out that I need to upgrade to a newer SHA to make this work. But even then, I can't make it right.

So, this is an example: I added a schema:

```clojure
(def EvtSchema [:cat any? string? fn?])
(mx/defn register-provider! [{:keys [db]}
                             a :- EvtSchema])
```

And, in another namespace, I have the `(malli.dev.cljs/start!)` running when my code hot-reloads. This doesn't really works - the error I'm seeing is `"Compile Warning": Use of undeclared Var my.namepace/EvtSchema` - maybe because `EvtSchema` is on another namespace than the one I'm running `start!`?

Also, I can't seem to make the code above work. What I want is to instrument that `a` must be a vector containing anything, a string, and a function. If I do run `start!` on the same namespace, Malli will complain that I have a wrong number of arguments when I try to run `register-provider!` with two arguments

lsenjov:
I'm playing around with this too atm

In model.query
```clj
(def Query [...])
(mx/defn get-fields :- [:set :string] [query :- Query] ...)
```

In dev.cljs, when macroexpanding (this is a snippet in a big tree of `mi/instrument!`
```clj
(set!
    model.query/get-fields
    (malli.core/-instrument
     {:schema (malli.core/function-schema [:=> [:cat Query] [:set :string]])}
     model.query/get-fields))
```

So it's not registering the function schema at compile in the same namespace, but bringing it over directly.

lsenjov:
With a bit more of a look, the same issue also happens if you use `m/=>` instead of `mx/defn`. It does appear to be worked around by fully qualifying each var with a full namespace (use the full `:- model.query/Query` instead of just `:- Query`), but I'm still testing this.

dvingo:
hi! there are some notes here on usage:
https://github.com/metosin/malli/blob/master/docs/clojurescript-function-instrumentation.md

And yes, this has been getting fixes recently so you'll likely need to use a recent git coordinate instead of a maven release version.
Specifically, this is the PR that fixed lots of the functionality:
https://github.com/metosin/malli/pull/681

[...]

PR #694, opened 2022-04-18
Fix breaking instrumenting => when using schemas in other namespaces

lsenjov:
Adds a fix and failing tests for https://github.com/metosin/malli/issues/693

At `m/=>` compile time, walks the schema for symbols and namespaces them.

Bug:
Very specifically: the issue happens when you `mi/instrument!` in namespace A, with `mx/defn` or `m/=>` a function in another namespace B, that uses a return type/arg type stored in another var in a third namespace C.
They also fail if the vars are in namespace B, and aren't `refer`red in namespace A.
It does appear to work correctly if the return/arg var are fully namespaced from B.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [eval type -deref deref -lookup -key])
   #?(:cljs (:require-macros malli.core))
   (:require [clojure.core :as c]
+            [clojure.walk :as walk]
             [malli.impl.regex :as re]
             [malli.impl.util :as miu]
             [malli.registry :as mr]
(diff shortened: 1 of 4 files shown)


dvingo:
We don't want to qualify all symbols - for example things that refer to cljs.core vars. 

You can extract this https://github.com/metosin/malli/blob/6751b1d767f6189cde56f39c9e79e950b7264aca/src/malli/instrument/cljs.clj#L14
fn to be used in both places (=> and collect!)

test/malli/instrument/cljs_test/indirect_fn.cljs
```diff
@@ -0,0 +1,28 @@
...
+  (:require [malli.core :as m]
+            [malli.experimental :as mx]
+            [malli.instrument.cljs-test.defs
+             :as defs
+             :refer [defs-small-int defs-int-arg]]))
+
+;; There's a matrix of things to test here
+;; Calling either arg or return type from another namespace
+;;  vs whether doing it as a refer or as a namespaced var
+;; This hits all 4
```
dvingo, line 13: For completeness, can you add the scenario where the schema is a core var like `int?`

src/malli/core.cljc
```diff
@@ -2399,18 +2400,37 @@
...
+           x))
+       d)))
+
 #?(:clj
    (defmacro => [name value]
      (let [name' `'~(symbol (str name))
            ns' `'~(symbol (str *ns*))
-           sym `'~(symbol (str *ns*) (str name))]
+           sym `'~(symbol (str *ns*) (str name))
+           value' (fix-syms &env value)]
```
dvingo, line 2426: Because we're in the core ns here it's probably a good idea not create more publicly visible vars, how about wrapping the `defmacro` with a let?

some code golf, `(:ns &env)` is an idiom often used to determine if you're targeting a cljs macro:
```clojure
#?(:clj 
  (let [resolve' (if (:ns &env) (ns-resolve 'cljs.analyzer.api 'resolve) resolve)
      resolve-symbols
      (fn [env d]
        (walk/postwalk
         (fn [x]       
            (cond->> x (symbol? x)
             (or (:name (resolve' env x)))))
         d)))

value' (resolve-symbols &env value))
  (defmacro => ...
```

dvingo: I'm curious because you added the `resolve` for clojure schemas as well - do clojure schemas have this issue as well if you `instrument!` from another namespace?

test/malli/instrument/cljs_test.cljs
```diff
@@ -47,6 +48,23 @@
 
     (is (= 4 (power "2")))
     (is (= 36 (power 6))))
+  (testing "without instrumentation, from another ns"
```
dvingo, line 51: Can you move the unstrument after the instrument in the test, this will actually test if the fns are unstrumented. (There was a bug with  the filters in the past so it's a good idea to put them in that order.)

src/malli/core.cljc
lsenjov: I'm... not actually sure. That's probably a good question to ask, it would simplify this whole mess a lot of I only need to call this in cljs

lsenjov:
I'm not enjoying pulling out `-qualify-sym`.

- malli.instrument.cljs requires malli.core
- malli.core/=> will need to use -qualify-sym
- it _only_ needs to use it in a cljs env (so far it looks to be working fine in clj)

Options are:
- Refer to malli.instrument.cljs/-qualify-sym directly in malli.core
- Create another ns under malli.instrument.cljs and move it there

It's a little more spaghetti, but I'm leaning towards the latter atm

dvingo:
Good news - I was confused and pulling out `qualify-sym` isn't needed. 
What you have looks good, but I left some comments here about it:
https://github.com/metosin/malli/pull/694#discussion_r852408623

dvingo:
@lsenjov hey - do you have time today to make the adjustments? I can submit a PR with the changes if not.

dvingo:
@lsenjov hey - do you have time today to make the adjustments? I can submit a PR with the changes if not.

lsenjov:
@dvingo not till the weekend. Feel free to run with it if you want to get it out sooner

Issue #695, opened 2022-04-18
Hot reload on Shadow-CLJS bugs


mauricioszabo:
I decided to add Malli and instrumentation on my Shadow-CLJS project (Node.JS script). I'm not being able to make instrumentation + hot-reload work.

For example, I tried to add `(when js/goog.DEBUG (prn :instrumenting (malli.dev.cljs/start!)))` on my `core.cljs` file. Then, on another file, I added (as an example):

```clojure
(ns saturn.rings.services
  (:require [malli.core :as m]))

(defn example
  {:malli/schema [:=> [:cat string? int?] keyword?]}
  [a b]

  (prn :RAN)
  :ok)
```

And I also included this file on my `core.cljs` file. Now, when I save this file, it does indeed gets instrumented - but then things get weird. If I save this file, then `(example 10 20)` fails as expected. But if I change the code to:

```clojure
(defn example
  {:malli/schema [:=> [:cat int? int?] keyword?]}
  [a b]

  (prn :RAN)
  :nok)
```

And save *only once*, then the `(example "a" 20)` works, and it returns `:nok`, but the schema changed to `int?`, so this should fail... if I save a second time, then it works correctly. I also tried to run `(malli.dev.cljs/stop!)` before the start!, and the same result happens.

dvingo:
Hi, there are some notes here: https://github.com/metosin/malli/blob/master/docs/clojurescript-function-instrumentation.md

because this works via macros you'll need to ensure all your code is loaded _and then_ apply the dev/start! (use shadow after-load hook)

mauricioszabo:
Well, I _am using_ the after-load hook:

```clojure
(defn ^:dev/after-load main []
  (when js/goog.DEBUG (dev/start!))
  ...more init code...
```

dvingo:
It's hard to diagnose without a sample repo. Most of the testing has been targeting the browser so there could be something different needed for node.js. A sample repository would help a lot to see what's going on.

mauricioszabo:
Sure, I am making my experiments on Saturn, branch "services", but I generated a new one (`malli-bug-example`) to test this bug.

I am starting the instrument here: https://gitlab.com/clj-editors/saturn/-/blob/malli-bug-example/src/saturn/view.cljs#L173
And I am testing on this namespace: https://gitlab.com/clj-editors/saturn/-/blob/malli-bug-example/src/saturn/rings/services.cljs#L8

[...]

PR #696, opened 2022-04-20
Only call functions from a cljs macro, not other macros

dvingo:
It seems like `-collect-all-ns` expanding to `(collect! {:ns ~(ana-api/all-ns)})` was causing issues with hot code reloading.
This PR switches to have the `dev/cljs/start!` macro use functions as helpers instead of other cljs macros.
Addresses:
https://github.com/metosin/malli/issues/695


ikitommi:
Thanks!!

PR #697, opened 2022-04-21
Resolve symbols used in function schemas during macroexpansion

dvingo:
Picking up on the work by @lsenjov in PR:
https://github.com/metosin/malli/pull/694

If a function schema uses a symbol alias either an ns alias or a refer'ed var the macro would expand as the alias literal. By resolving all symbols used in function schemas during macroexpansion time this problem is solved.


ikitommi:
Thanks2!!

PR #700, opened 2022-05-01
In => only resolve symbols for cljs

dvingo:
Some minor cleanup for recent cljs instrumentation updates.

- `(unchecked-get #"" "constructor")` => this is the implementation of `type` in cljs https://github.com/clojure/clojurescript/blob/40358fc4c53090676da82e3f4c52920003678902/src/main/cljs/cljs/core.cljs#L317
- Only resolve symbols in cljs. This isn't required for clojure because the schema will be evaluated by `function-schema` here https://github.com/metosin/malli/blob/a3db330eae029863e9be443cda3a2cafd8f61a33/src/malli/core.cljc#L2399

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2265,7 +2265,7 @@
 (defn class-schemas []
   {#?(:clj Pattern,
       ;; closure will complain if you reference the global RegExp object.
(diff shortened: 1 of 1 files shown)


ikitommi:
🥇

PR #701, opened 2022-05-02
Map `:re` schema to clj-kondo `:string`

ferdinand-beyer:
`:re` schemas are currently mapped to `:regex` in the clj-kondo type generator. This causes schemas that require a string matching a certain pattern to be considered invalid by clj-kondo, as the `:regex` type expects a regular expression object, e.g. a `java.util.regex.Pattern` object.

See [this Slack thread](https://clojurians.slack.com/archives/CLDK6MFMK/p1651066428745799) as well.


ikitommi:
Thanks!

PR #702, opened 2022-05-05
Add experimental inline instrumented defn macro

dvingo:
This came out of the discussion over here:
https://github.com/metosin/malli/issues/695

When working in cljs having the code output by macros in the same namespace you're working in will enable repl-driven development without needing to think about refreshing the namespace where `instrument!` was called. 

This PR adds a macro, inspired by Guardrails/Ghostwheel `>defn` macro, which will output an instrumented function.
A parallel `noop` ns can be added to allow controlling this via shadow-cljs config (see https://github.com/fulcrologic/guardrails#dead-code-elimination)

I needed to add the `{:registry (m/default-schemas)}` lines because the schemas used for the parsing code will not always be available if the user is using a custom registry.

### DIFF
--- a/src/malli/experimental.cljc
+++ b/src/malli/experimental.cljc
@@ -1,60 +1,94 @@
 (ns malli.experimental
-  (:refer-clojure :exclude [defn])
-  #?(:cljs (:require-macros malli.experimental))
-  (:require [clojure.core :as c]
-            [malli.core :as m]
-            [malli.destructure :as md]))
+  #?@(:cljs
+      [(:require-macros malli.experimental)
+       (:require
+         [malli.core :as m]
+         [malli.dev.pretty])])
+  #?@(:clj
+      [(:refer-clojure :exclude [defn])
+       (:require
+         [clojure.core :as c]
+         [malli.core :as m]
+         [malli.dev.pretty]
+         [malli.destructure :as md])]))
 
-(c/defn -schema [inline-schemas]
-  (m/schema
-   [:schema
-    {:registry {"Schema" any?
-                "Separator" (if inline-schemas [:= :-] md/Never)
-                "Args" [:vector :any]
-                "PrePost" [:map
-                           [:pre {:optional true} [:sequential any?]]
-                           [:post {:optional true} [:sequential any?]]]
-                "Arity" [:catn
-                         [:args "Args"]
-                         [:prepost [:? "PrePost"]]
-                         [:body [:* :any]]]
-                "Params" [:catn
-                          [:name symbol?]
-                          [:return [:? [:catn
-                                        [:- "Separator"]
-                                        [:schema "Schema"]]]]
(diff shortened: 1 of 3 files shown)


mauricioszabo:
So, it works but there are some issues with metadata:

image

`:arglists` is empty, and I tried to add a docstring but it also gets removed from metadata...

dvingo:
thanks for trying it out (and checking for the meta)! - I will refactor that to support it. I think the `>fn` should be straightforward too, I'm interested to see that in action

mauricioszabo:
Sure, please let me know if I can help in any way.

I tried some combinations of `with-meta` and others, and found a sad thing: seems that it's not possible to override the `:arglists` element using metadata. I looked at Prismatic Schema's implementation, and indeed it uses `defn` under the hood, so maybe it's the only way to make it work...

src/malli/experimental.cljc
```diff
@@ -1,60 +1,94 @@
...
+           parse        (fn [{:keys [args] :as parsed}] (merge (md/parse args) parsed))
+           ->schema     (fn [{:keys [schema]}] [:=> schema (:schema return :any)])
+           single       (= :single (key arities))
+           parglists    (if single (->> arities val parse vector) (->> arities val :arities (map parse)))
+           raw-arglists (map :raw-arglist parglists)
+           schema       (as-> (map ->schema parglists) $ (if single (first $) (into [:function] $)))]
+       `(def ~name (m/-instrument {:schema ~schema :report ~report}
+                     (c/fn ~name
+                       ~@(map (fn [{:keys [arglist prepost body]}] `(~arglist ~prepost ~@body)) parglists)
+                       ~@(when-not single (some->> arities val :meta vector))))))))
```
mauricioszabo, line 79: I found a way to keep the meta with minimum effort, if you want to use it:
```suggestion
           schema       (as-> (map ->schema parglists) $ (if single (first $) (into [:function] $)))
           meta-to-defn (assoc meta :doc doc)]
       `(do
          (c/defn ~name ~meta-to-defn ~(map :arglist parglists))
          (set! ~name
            (m/-instrument {:schema ~schema :report ~report}
                           (c/fn ~name
                                 ~@(map (fn [{:keys [arglist prepost body]}] `(~arglist ~prepost ~@body)) parglists)
                                 ~@(when-not single (some->> arities val :meta vector)))))))))

```

But this code does not work with Clojure (because you can't `set!` root bindings). I'm supposing `->defn` is considered an implementation detail, so we can send an additional parameter (like `is-clojurescript?`) to handle the Clojure case?

dvingo: Thanks for looking into this! based on that I went with a strategy to produce an actual defn that just delegates to the instrumented version.

dvingo:
This was implemented in https://github.com/metosin/malli/pull/825

PR #704, opened 2022-05-10
fix: ClojureScript podcast renamed to ClojureStream podcast

BTowersCoding:
This PR just fixes a broken link since the podcast was renamed.


ikitommi:
thanks!

PR #711, opened 2022-06-13
Show function input arguments on invalid return value

ingesolvoll:
When an instrumented function's return value fails schema validation, it's useful to see the input args. This PR adds the input args to the pretty printer output.

[Slack thread with some details](https://clojurians.slack.com/archives/CLDK6MFMK/p1655107719370529)


ikitommi:
Looks good, thanks!

PR #712, opened 2022-06-20
Don't lose properties when updating a key

### LINKED ISSUE #645: When using mu/update-in on a conditional key, it shows itself as optional but it is actually mandatory
(ns my-project.malli-bug
  (:require [malli.core :as m]
            [malli.dev.pretty :as pretty]
            [malli.error :as me]
            [malli.util :as mu]))

(def my-schema
  (m/schema [:map
             [:y {:optional true} :int]]))

(def my-other-schema (mu/update-in my-schema [:y] (fn [_] true?)))
(m/validate my-other-schema {}) ; bug here
(m/explain my-other-schema {}) ; bug here
(m/validate (mu/optional-keys my-other-schema [:y]) {})

From the repl:

(m/validate my-other-schema {})
=> false
(m/explain my-other-schema {})
=>
{:schema [:map [:y {:optional true} true?]],
 :value {},
 :errors ({:path [:y], :in [:y], :schema [:map [:y {:optional true} true?]], :value nil, :type :malli.core/missing-key})}

dmcgillen:
Fixes #645

Updating a key in a schema that had `{:optional true}` (or any properties) would lose those properties, and a subsequent attempt to validate against that schema without that key would fail. It appears to me (not looked at this codebase before) that this was simply because the new child was being calculated before calculating the properties.

One thing I admittedly don't understand is that even before my fix the output to update looked ok:

```
(malli.util/update
  [:map
   [:a {:optional true} int?]
   [:b {:optional true} string?]]
  :a
  identity)
;; => [:map [:a {:optional true} int?] [:b {:optional true} string?]]
```

This is why the existing tests pass, because they just check this output (I've added a test to run `validate`).

But if you, say, called `malli.util/to-map-syntax` on it, you could see the properties had been lost:

```
{:type :map, :children [[:a nil {:type int?}] [:b {:optional true} {:type string?}]]}
```

With this PR that is no longer the case.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -374,8 +374,8 @@
       ;; remove
       (letfn [(cut [v] (into (subvec v 0 i) (subvec v (inc i))))]
         (-simple-entry-parser (dissoc keyset k) (cut children) (cut forms)))
-      (let [c [k p s]
-            p (if i (if override p (nth (children i) 1)) p)
+      (let [p (if i (if override p (nth (children i) 1)) p)
+            c [k p s]
             f (if (seq p) [k p (-form s)] [k (-form s)])]
         (if i
           ;; update
--- a/test/malli/util_test.cljc
+++ b/test/malli/util_test.cljc
@@ -466,7 +466,13 @@
                     [:a int?]
                     [:b {:optional true} int?]
                     [:c string?]
-                    [:d boolean?]]))))
+                    [:d boolean?]]))
+
+    (testing "Optional property is maintained when optional key is updated"
+      (is (true? (m/validate
+                  (mu/update schema :b identity)
+                  {:a 1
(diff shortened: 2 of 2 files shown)


src/malli/core.cljc
```diff
@@ -374,8 +374,8 @@
       ;; remove
       (letfn [(cut [v] (into (subvec v 0 i) (subvec v (inc i))))]
         (-simple-entry-parser (dissoc keyset k) (cut children) (cut forms)))
-      (let [c [k p s]
-            p (if i (if override p (nth (children i) 1)) p)
+      (let [p (if i (if override p (nth (children i) 1)) p)
+            c [k p s]
```
dmcgillen, line 378: No change here, just switched the two lines around

ikitommi:
A really good catch! thanks!

PR #715, opened 2022-06-26
:maybe map schema can use :nilable prop in clj-kondo :op spec

### LINKED ISSUE #714: `[:maybe [:map ...]]` produces non-nilable clj-kondo config
If I define a function schema like 
```clojure 
[:maybe [:map [:a-number :int]]]
```

The generated clj-kondo config will be 

```clojure
{:op :keys
 :req {:a-number :int}}
```

This is not correct, and triggers unwanted linter warnings when passing `nil` to the function. I have [opened an issue](https://github.com/clj-kondo/clj-kondo/issues/1735) in clj-kondo to support nilable map configs, but until then maybe the generated clj-kondo config should be `:nilable/map` in this case?

ingesolvoll:
Fixes #714 

The accompanying clj-kondo [PR](https://github.com/clj-kondo/clj-kondo/pull/1736) has been merged, but not released yet.

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -95,7 +95,11 @@
           (symbol? child) :symbol
           :else :any)))))
 
-(defmethod accept :maybe [_ _ [child] _] (if (keyword? child) (keyword "nilable" (name child)) child))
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks!

PR #718, opened 2022-06-30
Fix #302: babashka compatibility

### LINKED ISSUE #302: Not working with babashka
```clojure
user=> (require '[malli.core :as ma])
Could not resolve symbol: deftype [at borkdude/dynaload.cljc:4:2]
```

I made several quick fixes here to make the `require` work: https://github.com/lucywang000/malli/commit/1f46840

But it still can't run the simplest validation:

```clojure
user=> (require '[malli.core :as ma])
nil
user=> (ma/validate int? "1")
java.lang.NullPointerException [at malli/core.cljc:1003:7]
```

Looks like something wrong with babashka's defprotocol implementation.

borkdude:
This PR relies on bb master which adds compatibility with malli. Fixes #302.

Notes:
- To make malli work well with bb, the `:bb` reader conditionals rely less on internal details of Clojure and use core functions instead, similar to what happens in `:cljs` branches. As discussed with @puredanger, I didn't add `LazilyPersistentVector` to bb as that was too far off from what is considered a public API.
- Upgrade `borkdude/dynaload` which was made compatible with bb
- Add bb test runner to ensure no breakage happens with future changes to malli
- The evaluation part is implemented by just using `load-string` in bb, since SCI itself isn't exposed yet (might happen in the future). SCI also has `eval-form` which lets you skip the string parsing bit, that might be more performant, but we can address that in a future PR.

@ikitommi I'm not sure what the workflow is with `metosin/malli` but I usually squash+merge the commits through the Github  option (dropdown in merge button). If you want I can squash this PR manually as well.

If you want to try:

```
bash <(curl -s https://raw.githubusercontent.com/babashka/babashka/master/install) --version 0.8.157-SNAPSHOT --dir /tmp
```

for installing bb master in /tmp.
Then in a script:

``` clojure
(ns malli-test
  (:require [babashka.deps :as deps]))

(deps/add-deps '{:deps {metosin/malli
                        {:git/url "https://github.com/borkdude/malli"
                         :git/sha "1ec8582e3a39c8d170cf5f8a7482386bf66370fe"}}})

(require '[malli.core :as malli])

(prn (malli/validate [:map [:a [:int]]] {:a 1}))
(prn (malli/explain [:map [:a [:int]]] {:a "foo"}))
```

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -59,3 +59,25 @@ jobs:
         run: npm ci
       - name: Run tests
         run: bin/node
+
+  build-bb:
+    name: Babashka
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup Java 11
+        uses: actions/setup-java@v3.4.0
+        with:
+          distribution: zulu
+          java-version: 11
+          cache: maven
+      - name: Setup Clojure
+        uses: DeLaGuardo/setup-clojure@master
+        with:
+          cli: latest
+          # bb: latest
+      - name: Download bb master
+        run: bash <(curl -s https://raw.githubusercontent.com/babashka/babashka/master/install) --version 0.8.157-SNAPSHOT
+      - name: Run tests
+        run: bb test-bb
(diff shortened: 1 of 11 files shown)


borkdude:
Sorry, had to download bb master instead of using the setup-clojure one.

borkdude:
@ikitommi Addressed feedback: squashed commit and updated README.

You can see how the README looks here:

https://github.com/borkdude/malli/tree/babashka#babashka

If you will release 0.8.9 then I will release a new version of babashka.

ikitommi:
💯

PR #719, opened 2022-06-30
Set additionalProperties if map is closed

### LINKED ISSUE #716: json-schema generation, specifying :additionalProperties
Hey friends,

I have this schema:
```clojure
(def Item
  (m/schema
    [:map {:closed true}
     [:description [:maybe :string]]
     [:id :int]
     [:name :string]
     [:type :string]]))
```
Calling `malli.json-schema/transform` on it returns:
```clojure
{:type "object",
 :properties
 {:description {:oneOf [{:type "string"} {:type "null"}]},
  :id {:type "integer"},
  :name {:type "string"},
  :type {:type "string"}},
 :required [:description :id :name :type]}
```
I would like it if `{:closed true}` would set `:additionalProperties false`, which would align it with the [spec](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties).

Thanks so much.

NoahTheDuke:
Decided to take a swing at this.

Closes #716

### DIFF
--- a/src/malli/json_schema.cljc
+++ b/src/malli/json_schema.cljc
@@ -80,11 +80,14 @@
 (defmethod accept :orn [_ _ children _] {:anyOf (map last children)})
 
 (defmethod accept ::m/val [_ _ children _] (first children))
-(defmethod accept :map [_ _ children _]
+(defmethod accept :map [_ schema children _]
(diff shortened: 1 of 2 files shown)


ikitommi:
thanks!

PR #720, opened 2022-07-04
switch Value and Errors in the pretty output order

valyagolev:
It's common for the error to go on top of the error report, as the most pertinent part of the information. Otherwise, if the value is too big, the output might become too unwieldy and hide the error. Even now it's not the first thing that catches the eye.

Perhaps there are other ways to address this rather than by changing the default. Please let me know... But I think it might be a reasonable change in this case.

Screenshot 2022-07-05 at 01 01 30

### DIFF
--- a/src/malli/dev/pretty.cljc
+++ b/src/malli/dev/pretty.cljc
@@ -34,8 +34,8 @@
   (let [{:keys [schema value]} explanation]
     {:body
      [:group
-      (-block "Value:" (v/-visit value printer) printer) :break :break
       (-block "Errors:" (v/-visit (me/humanize explanation) printer) printer) :break :break
+      (-block "Value:" (v/-visit value printer) printer) :break :break
       (-block "Schema:" (v/-visit schema printer) printer) :break :break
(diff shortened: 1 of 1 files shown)


ikitommi:
Thanks. Another way to resolve this is not to print valid values, like expound does. Did a spike on that earlier, looks like this:

```clojure
;; given a schema
(-> [:map
     [:id :string]
     [:tags [:set :keyword]]
     [:address [:map
                [:street :string]
                [:city :string]
                [:zip :int]
                [:lonlat [:tuple :double :double]]]]]
    ;; explain a value against it
    (m/explain
     {:id "Lillan"
      :tags #{:artesan "coffee" :garden "ground"}
      :address {:street "Ahlmanintie 29"
                :zip 33100
                :lonlat [61.4858322, "23.7832851,17"]}})
    (me/error-value
     {::me/mask-valid-values '...}))
;{:id ...
; :tags #{"coffee" "ground" ...}
; :address {:street ...
;           :zip ...
;           :lonlat [... "23.7832851,17"]}}
```

will finish that at a later time and merge this now.

valyagolev:
oh yes that would be very helpful, thank you. and thanks for the merge

Issue #734, opened 2022-08-08
:tuple requires vector


NoahTheDuke:
Hey there! I just ran into an issue where using `mt/strip-extra-keys-transformer` wasn't working because I had `[:tuple OtherSchema]` and was providing a seq: `(OtherSchema)`. I'm not using any Malli validation yet (planned but currently relying on json schema validation) which may have caught this, but it seems worth considering changing. I'd prefer to not have to call `(m/encode Example (vec obj) mt/strip-extra-keys-transformer)` just to be sure that I'm providing a vector, as sometimes I'm transforming straight from a database call and the output is a seq instead of a vector.

Thanks so much!

vharmain:
Hi!

I guess tuples should be vectors by design. README says:

> A :tuple describes a fixed length Clojure vector of heterogeneous elements

I'm not sure what is the reason for this but maybe @ikitommi can chime in. :)

However, one option is to use [sequence schema](https://github.com/metosin/malli#sequence-schemas) instead.

ikitommi:
perf, of course :) you can easily add a custom transformer in user space that does the transforming for tuples, relevant helpers exists already, see: https://github.com/metosin/malli/blob/master/src/malli/transform.cljc#L416-L422.

PR to add `:tuple` to `mt/-collection-transformer` most welcome. You can easily compose that with the strip-keys-thing.

NoahTheDuke:
> However, one option is to use sequence schema instead.

Shortly after I posted, I learned that the collection could be empty, so a sequence schema is better for my use case lol. Thanks for mentioning it anyway. Still glad I posted!

> PR to add :tuple to mt/-collection-transformer most welcome.

Sure thing! Thanks for pointing that out. I'll write it up tonight.

ikitommi:
PR merged, thanks! For ease of use, I think it would be good to have a new built-in transformer that would have this as built-in (e.g. string + collections). But, it's another issue. Closing this one.

PR #735, opened 2022-08-09
Add seq to tuple transformation

NoahTheDuke:
Implementation of #734. I chose `-sequential->vector` instead of the more generous `-sequential-or-set->vector` as order matters in tuples.


ikitommi:
good call on not coercing from sets 💯

ikitommi:
Thanks!

PR #743, opened 2022-08-22
Update README docs

danielcompton:
I was reading through the README and found a few small things that might make it clearer.


ikitommi:
Thanks!

PR #746, opened 2022-08-26
Attach fn-name metadata to instrumented CLJS functions

ingesolvoll:
Our app uses function values for looking up their original name/symbol from a registry. It is not something we can or want to change, and using malli instrumentation for CLJS breaks this. The reason is that after instrumentation, the anonymous wrapper has a different identity than the original function.

If malli could attach the original symbol as metadata, we are able to use this as an alternative way of obtaining the same thing.

I used the key `:fn-name` as that was used elsewhere in the same code. Possible alternatives:
- `:instrumented-symbol`
- `:instrumented-var`
- `:instrumented-fn`

### DIFF
--- a/src/malli/instrument/cljs.cljs
+++ b/src/malli/instrument/cljs.cljs
@@ -7,5 +7,14 @@
 (defn -filter-var [f] (fn [_ s _] (f s)))
 (defn -filter-ns [& ns] (fn [n _ _] ((set ns) n)))
 
+(defn meta-fn
+  ;; Taken from https://clojure.atlassian.net/browse/CLJS-3018
+  ;; Because the current MetaFn implementation can cause quirky errors in CLJS
+  [f m]
+  (let [new-f (goog/bind f #js{})]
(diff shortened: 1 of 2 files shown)


src/malli/instrument/cljs.cljs
```diff
@@ -7,5 +7,14 @@
 (defn -filter-var [f] (fn [_ s _] (f s)))
 (defn -filter-ns [& ns] (fn [n _ _] ((set ns) n)))
 
+(defn meta-fn
```
ingesolvoll, line 10: Stolen from https://github.com/CrypticButter/snoop/blob/main/src/crypticbutter/snoop/impl/cljs.cljs

ikitommi:
As it's a symbol, I think `:fn-name` is misleading. I like the `:instrumented-symbol` (it's not a Var either, or not a name).

ingesolvoll:
@ikitommi Thanks, I updated the name.

ikitommi:
Awesome, thanks!

PR #747, opened 2022-08-26
Group instrumentation logging so it doesn't spam the browser console

ingesolvoll:
When instrumenting tens or hundreds of CLJS functions, getting all those `.. instrumented my.ns` logged in the browser console is quite noisy. 

This makes it slightly better by grouping them collapsed.

A logging config would probably be ideal, but this does the trick for us.


ikitommi:
yes, this is much better, thanks!

PR #748, opened 2022-08-30
Improve generator distribution for sized schemas

frenchy64:
We want the same distribution for `gen/vector` for both
these cases:
1. only :min is provided
2. no min-max opts are passed.

This fixes generation for 1 by using + to shift the length
distribution instead of squashing it via clojure.core/max.

Same approach as clojure.test.check.generators/coll-distinct-by:
  https://github.com/clojure/test.check/blob/b6a318af92e836f1f20960f38a5944168b88d60d/src/main/clojure/clojure/test/check/generators.cljc#L799


ikitommi:
Thanks!

PR #754, opened 2022-09-19
Use shadow-cljs build hooks to optimize cljs instrumentation

dvingo:
First pass at optimizing the cljs instrumentation so that every namespace in an application is not blindly re-instrumented after editing any namespace. Only the dependent namespaces along the require path in the graph of namespaces that are affected by the updated namespaces are re-instrumented.

Adding `refresh!` was my solution to the problem of having multiple browsers open in app during development. on-load we instrument every namespace (using `start!`), but on subsequent builds we only instrument those that were edited.

I am not sure if there is a simpler way of collecting the edited namespaces in shadow-cljs before compilation runs again. A simpler implementation would somehow grab the affected namespaces in a `:compile-prepare` hook. I am currently not aware of a way to get at that information, so this solution is what I landed on.


thheller:
Can you give a brief overview about what instrumentation actually does? Like a bit of pseudo-code ish or so. How does a var look before instrumentation and after? The hook you have currently is definitely not ideal, but I'm not sure what you actually need. I'm not sure you need a hook at all.

dvingo:
Hey @thheller sure thing.

Instrumentation (`malli.dev.cljs/start!`) loops over all namespaces (via `cljs.analyzer.api all-ns`) and each public var in them. If the var has a [`:malli/schema`](https://github.com/metosin/malli/blob/56533336ceb742e63a84d5c1c13095c0ef5383ef/src/malli/instrument/cljs.clj#L13) [function schema](https://github.com/metosin/malli/blob/master/docs/function-schemas.md) in its metadata then we add it to an atom of collected vars (keyed by namespace).
After this we loop over all of those collected (registered) vars and for each one output a line like this in a macro:
```clojure
`(set! ~fn-sym (malli.core/-instrument ~schema-map ~fn-sym))
```
We place this macro in our application init function and mark the entry namespace to always be recompiled with `{:dev/always true}`

On first load of the application we instrument every function in the app that was registered.

As a user makes edits to the application they may remove schemas, update them, and add them. This will happen in one hot-reload loop. The purpose/idea of this PR (the `malli.dev.cljs/refresh!` fn) is that all of the existing namespaces that were instrumented will remain instrumented because their code in the JS runtime is the same as the prior compile+load, only the edited namespaces will have to be re-instrumented. So the goal is somehow communicating to the macro which namespaces to loop over in order to emit the minimal amount of code per hot-reload cycle, and this must be done before the macro is recompiled.

thheller:
So, I think the mistake here is moving everything to the CLJ side. Only because one part needs to be done in CLJ doesn't mean all of it has to happen there.

My suggestion would be to only do the collection part in CLJ. Have that macro supply data to the client.

```
(defmacro collect! []
  (->> all-vars-with-metadata
       (map (fn [var] `(some.helper/add-meta ...)
       ...)
```

You end up with a bunch of `(some.helper/add-meta "my.app/foo" {:some-data ...})`. What this looks like is entirely up to you and what malli needs. The goal should be to make this trivial so it can just always run for everything.

That `add-meta` can either immediately or later (on some instrument! call) apply the actual instrumentation. It has all the var into it needs and can look up actual runtime things easily. It can leave marker values on the actual thing to check if something is already instrumented, which would catch hot-reloaded things since they'll no longer have that marker.

I think this gives you much more flexibility and gets you completely out of the macro madness. `cljs.test` also made that mistake and moved far too much code into macros that should never have been macros. So, `shadow.test` tries to address this by collecting all metadata via a macro (eg. for [:node-test](https://github.com/thheller/shadow-cljs/blob/master/src/main/shadow/test/node.cljs#L16-L18)) and then the [running tests](https://github.com/thheller/shadow-cljs/blob/master/src/main/shadow/test.cljs#L134) is done by actual functions, not this macro madness in `cljs.test`.

dvingo:
Makes a lot of sense, thanks @thheller! I think I was slowly coming to this realization as well - we have all the control at runtime, just need to communicate the info from the compiler. I'll try an implementation using that strategy.

dvingo:
I suppose to clarify - this is already what the macro is doing - instead of your `add-meta` call, it is inline code, but it outputs logic for whether or not to instrument for _every var_. My main concern was generating lots of runtime code for large codebases which this PR addresses. The generated code on hot-reload is only for the edited namespaces. The only change to the logic you described above vs what the instrumentation does now is storing a flag indicating whether instrumentation was applied for a var already and if so skipping doing so again.

I'm not sure there is much macro madness going on - the `set!` calls will have to be generated because `set!` only supports a symbol literal. Were you thinking of some other solution to replace a function's implementation at runtime?

thheller:
Yes, don't limit yourself to CLJS ops (eg. `set!`). Things may require doing direct JS calls (eg. via `goog.object`). I only glanced at https://github.com/metosin/malli/blob/master/src/malli/instrument/cljs.clj and it seems to me that 90% of this shouldn't be done by a macro.

dvingo:
That makes a lot more sense about using JS calls - I was thinking we were limited by use of `set!` only. The current implementation is heavily inspired from how instrumentation happens in cljs.spec.alpha, but I see now that there are advantages to not following that strategy. Thanks for the info!

PR #755, opened 2022-09-22
Refactor cljs fn instrumentation to happen at runtime

dvingo:
Based on feedback from Thomas Heller (in https://github.com/metosin/malli/pull/754) this PR moves function instrumentation to occur in the JS runtime instead of via macros which emit code to replace the JS functions. The only part that happens at compile time is schema collection. This results in a much closer match to the Clojure implementation.
One downside of this approach is that instrumentation will not work under advanced compilation, but since its intended use is during development this tradeoff makes sense.

I excluded the instrumentation tests from running under advanced compilation.

I left in the namespace `malli.instrument.cljs` for backwards compatibility but it should likely be removed in the future after some grace period.

### DIFF
--- a/app/malli/helpers.cljs
+++ b/app/malli/helpers.cljs
@@ -4,3 +4,8 @@
   [:int {:max 6}])
 
 (def int-schema :int)
+
+(defn x+y
+  {:malli/schema [:=> [:cat float? float?] :double]}
+  [x y]
+  (+ x y))
--- a/app/malli/instrument_app.cljs
+++ b/app/malli/instrument_app.cljs
@@ -1,17 +1,20 @@
 (ns malli.instrument-app
   (:require
-    [malli.clj-kondo :as mari]
+    [malli.instrument :as mi-new]
+    ;[malli.clj-kondo :as mari]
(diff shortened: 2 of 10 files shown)


ikitommi:
Not 100% sure I understood what is happening here, but understand the idea, looks great. 🙇

dvingo:
Awesome! I'm happy to add some markdown notes in the `docs` directory with what is happening if you want.

ikitommi:
All docs most welcome, but no need to do it for me. I'll re-read the current instrumentation code for cljs with a good cup of coffee at some point.

knubie:
@dvingo Does the advice in https://github.com/metosin/malli/blob/master/docs/clojurescript-function-instrumentation.md still apply after this change? Specifically this line:

> In your application's entry namespace you need to tell the compiler to always reload this namespace so that the macro will rerun when you change schemas and function definitions in other namespaces while developing.
> 
> We do this with the `{:dev/always true}` metadata on the namespace:

dvingo:
@knubie I think it's still a good idea, the collection is still a macro so you'll want that to re-compile after every change. If not, the collected schemas could be stale.

PR #756, opened 2022-09-22
Improve regex driver cache performance

bsless:
Unroll values loop avoids thunk allocation and keeps primitives on stack

Calling Array/newInstance avoids allocating a Number when allocating an object-array

Improve hashing performance - inline all hash functions and call hashCode directly on function object

### DIFF
--- a/src/malli/impl/regex.cljc
+++ b/src/malli/impl/regex.cljc
@@ -35,7 +35,10 @@
 
   (:refer-clojure :exclude [+ * repeat cat])
   (:require [malli.impl.util :as miu])
-  #?(:clj (:import [java.util ArrayDeque])))
+  #?(:bb (:import [java.util ArrayDeque])
+     :clj (:import [java.util ArrayDeque]
+                   [clojure.lang Util Murmur3]
+                   [java.lang.reflect Array])))
 
 ;;;; # Driver Protocols
 
@@ -445,6 +448,8 @@
 
 (deftype ^:private CacheEntry [^long hash f ^long pos regs])
 
+#?(:clj (set! *unchecked-math* true))
+
 ;; Custom hash set so that Cljs Malli users can have decent perf without having to to set up Closure ES6 Set polyfill.
 ;; Uses quadratic probing with power-of-two sizes and triangular numbers, what a nice trick!
 (deftype Cache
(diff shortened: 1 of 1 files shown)


bsless:
I'll fix the babashka tests :)

ikitommi:
Looks great! Could you run a perf test before and after?

bsless:
Tested with the malli datalog parser, before is malli 0.8.9, after is using my branch as a git dep.

```clojure
(ns bsless.datalog-parser.malli.perf.core
  (:require
   [criterium.core :as cc]
   [bsless.datalog-parser.malli :refer [parse]]))

(def q1 '[:find ?e
          :where [?e :age 42]])

(cc/quick-bench (parse q1))
;; Execution time mean : 7.907128 µs -> 6.496485 µs

(def q2
  '[:find ?track-name ?minutes
    :in $ ?artist-name
    :where [?artist :artist/name ?artist-name]
    [?track :track/artists ?artist]
    [?track :track/duration ?millis]
    [(quot ?millis 60000) ?minutes]
    [?track :track/name ?track-name]])

(cc/quick-bench (parse q2))
;; Execution time mean : 31.873938 µs -> 25.940028 µs

(def q3
  '[:find (count ?artist) .
    :where [?artist :artist/name]
    (not-join [?artist]
              [?release :release/artists ?artist]
              [?release :release/year 1970])])

(cc/quick-bench (parse q3))
;; Execution time mean : 26.251216 µs -> 21.549259 µs

(def q4 '[:find ?e
          :keys foo
          :in $ ?fname ?lname
          :where [?e :user/firstName ?fname]
          [?e :user/lastName ?lname]])

(cc/quick-bench (parse q4))
;; Execution time mean : 19.228749 µs -> 14.364110 µs
```

bsless:
@ikitommi any other comparisons you'd like to see?

src/malli/impl/regex.cljc
```diff
@@ -455,22 +460,35 @@
...
-                   (loop [i* (bit-and (.-hash v) max-index), collisions 0]
-                     (if (aget values* i*)
-                       (let [collisions (unchecked-inc collisions)]
-                         (recur (bit-and (unchecked-add i* collisions) max-index) ; i* = (i* + collisions) % capacity*
-                                collisions))
-                       (aset values* i* v)))))
+
+        (let [len (alength values)]
+          (loop [i 0]
+            (if (< i len)
```
ikitommi, line 470: `when` would be better here? e.g. removed `do` + `nil` return

ikitommi:
thanks for the numbers, nice gains!

ikitommi:
https://github.com/metosin/malli/commit/6f0b87e37b443c0c5b0eb10fee29b6b45be2f0f1

PR #759, opened 2022-10-06
Fix #758: generate alphanumeric strings with :min/:max

### LINKED ISSUE #758: string type with min property - Generator generates Invalid UTF-8 strings
**Problem**
When adding the `:min` property to a string type, malli.generator/generate starts generating non utf-8 strings. 

**Reproducing the error** 
`(malli.generator/generate [:string {:min 10}])`

**Version**
0.8.9

frenchy64:
The generator for `:string` seems to be intended as a variant on `gen/string-alphanumeric`. If that's the case, we need to use `gen/char-alphanumeric` instead of `gen/char`.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -91,10 +91,10 @@
 (defn- -string-gen [schema options]
   (let [{:keys [min max]} (-min-max schema options)]
     (cond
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks!

PR #766, opened 2022-10-16
Fix instrument! in dev.cljs/start! - pass cljs function schemas

dvingo:
I totally missed this oversight when updating the code to happen at runtime. 
Nothing was being instrumented in a cljs app I'm working on because `instrument!` is now shared with clojure which defaults to `:clj` schemas.
This fix passes the `:cljs` schemas to `instrument!`


ikitommi:
thanks, will create a new release with this.

ikitommi:
https://github.com/metosin/malli/blob/master/CHANGELOG.md#091-2022-10-17

PR #769, opened 2022-10-18
Add tests for malli.dev and malli.dev.cljs

dvingo:
Adding tests for `malli.dev/start!` and `malli.dev.cljs/start!`.

This resulted in cleaning up some some of the cljs code to ensure the functions exist before they are instrumented and using `resolve` instead of `var` when passing vars to the filter functions, which will return `nil` instead of throwing if a var is missing.

This was discovered from having multiple shadow-cljs builds running, the collection happens at compile time so if you collect all namespaces their vars may not be available in all builds.

In `malli.dev.cljs/start!` the `:ns` argument wasn't properly supported, this PR fixes that.


ikitommi:
Thanks^2! The `:ns` args, I think it's useful to be able to limit what gets loaded, but having two ways to filter namespaces is kinda weird.

PR #772, opened 2022-10-26
deref-all in utils?

ikitommi:


ikitommi:
Thanks but no thanks.

PR #774, opened 2022-10-27
fix: _Key must be integer_ when `assoc` called with :kw on a vec

piotr-yuxuan:
At some point I stumbled upon this unexpected behaviour:

``` clojure
(def explanation
  {:value {:my-field ["id" "credated-at"]} ; Come from an http request, decoded with malli.
   :errors (list {:path [:my-field 0]
                  :in [:my-field :credated-at]
                  :schema [:enum :created-at :id]
                  :value :credated-at})})

(me/humanize explanation)
; => (IllegalArgumentException. "Key must be integer")

(alter-var-root
  (var me/-push)
  (constantly (fn [x k v fill]
                (let [i (when (int? k) k)
                      x' (cond-> x (and i (sequential? x) (> k (count x))) (me/-fill k fill))]
                  (cond (nil? x') (assoc x' k v)
                        (and (associative? x') i) (assoc x' i v)
                        (associative? x') (assoc x' (count x') v)
                        (set? x') (conj x' v)
                        :else (apply list (assoc (vec x') k v)))))))

(me/humanize explanation)
; => {0 [["should be either :created-at or :id"]]}
```

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -146,8 +146,11 @@
 (defn -fill [x i fill] (-concat x (repeat (- i (count x)) fill)))
 
 (defn -push [x k v fill]
-  (let [x' (cond-> x (and (int? k) (sequential? x) (> k (count x))) (-fill k fill))]
-    (cond (or (nil? x') (associative? x')) (assoc x' k v)
+  (let [i (when (int? k) k)
+        x' (cond-> x (and i (sequential? x) (> k (count x))) (-fill k fill))]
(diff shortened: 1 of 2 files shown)


piotr-yuxuan:
Please feel free to guide further explanation on why this behaviour would actually be intended, or to advise on how to better fix this bug.

ikitommi:
closing this as an old draft. thanks.

piotr-yuxuan:
Sure, no worries @ikitommi and thanks again for this liberating piece of software!

PR #776, opened 2022-10-28
Quick poke on schematized def

ikitommi:
... should this fail always or just under instrumentation? Plumatic fails always.

TODO:

- [ ] support `:malli/always` meta to fail early
- [ ] collect schematized defs also to a registry (new registry or the current with a hint, it's not fn schema?)
- [ ] tests

Screenshot 2022-10-28 at 16 55 31


escherize:
mx/defn only fails from isntrumentation. I think using :malli/always is the way to go here.

ikitommi:
`:malli/always` is a good idea!

ikitommi:
there is no way to collect non-function schemas atm. need to add support for it before this can be merged.

ikitommi:
This would be a valuable addition, but should be finished. Help welcome!

opqdonut:
Closing as inactive. Feel free to open a new PR if someone wants to pick this up!

PR #777, opened 2022-10-28
cljs instrument, add function name in error when schema creation fails

dvingo:
If the user creates an invalid schema on a function this will help point to the source of the problem. Currently you are only informed which schema failed to be created, but not which function name.


ikitommi:
Thanks!!

Issue #264, opened 2020-09-25
Derive and declare effective types for schemas


ikitommi:
We should be able to both declare and derive/infer effective types for wrapper schemas like `:and`, `:multi` and `:enum`.

forked from #246. Replaces #39.

### This is clearly also `:map`

```clj
[:multi {:dispatch :type}
 [:sized [:map 
          [:type [:= :sized]] 
          [:size int?]]]
 [:human [:map 
          [:type [:= :human]] 
          [:name string?] 
          [:address [:map [:country keyword?]]]]]]
```

### this is an `:int`

```clj
[:fn {:type :int} '(fn [x] (< 10 x 100))]
```

### and this is a `:keyword`

```clj
[:enum :kikka :kakka :kukka]
```

### this too

```clj
[:and keyword? keyword?]
```

ikitommi:
working on this.

opqdonut:
There have been a number of developments under this issue, but momentum ran out sometime in 2023. It's unclear what usecases remain unhandled. Please open new issues for related, specific needs!

PR #782, opened 2022-11-18
infer :enum and := transformers from int? or keyword? values

ikitommi:
Similar to Plumatic [solution](https://github.com/plumatic/schema/blob/13725858557d247008c458b76841cba8546bd0de/src/cljc/schema/coerce.cljc#L74).

```clojure
;; uniform :enum type
(m/decode [:enum :A :B :C] "A" (mt/string-transformer))
; => :A

;; non-uniform :enum type
(m/decoder [:map [:x [:enum :A :B "C"]]] (mt/string-transformer))
; => #object[clojure.core$identity]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Malli is in well matured [alpha](README.md#alpha).
 * **BREAKING**: Change default of `:malli.provider/map-of-threshold` from 3 to 8
 * New `malli.core/coercer` and `malli.core/coerce` to decode and validate a value, throws on validation error.
 * New `malli.core/-no-op-transformer`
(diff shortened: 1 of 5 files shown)


ikitommi:
note - this is not a full solution to the problem. see https://github.com/metosin/malli/issues/264

Deraen:
Makes sense.

What about other "simple" types in json/string-decoders that might be used?
Symbols, uuid, double, float?

ikitommi:
added support for symbol, double and float (double via float). I think this is good enough now, as said, we should have a `-value-type` definition ("what type of values does this accept") separate of `-type` ("name of the schema") as this applies to many applications like:

* JSON Schema mapping
* Value generation
* Value providing/inferring
* transforming

PR #783, opened 2022-11-19
Add -parser, -unparser to -util-schema

abcdw:
@ikitommi @vharmain I was playing around with :merge schema and realized that if used it breaks parse/parser, I did a quick research and made a fix.

### DIFF
--- a/src/malli/util.cljc
+++ b/src/malli/util.cljc
@@ -416,6 +416,8 @@
           m/Schema
           (-validator [_] (m/-validator schema))
           (-explainer [_ path] (m/-explainer schema path))
+          (-parser [this] (m/-parser schema))
+          (-unparser [this] (m/-unparser schema))
           (-transformer [this transformer method options]
             (m/-parent-children-transformer this [schema] transformer method options))
           (-walk [this walker path options]
--- a/test/malli/util_test.cljc
+++ b/test/malli/util_test.cljc
@@ -912,52 +912,59 @@
 
     (testing "merge"
       (let [s (->> [:merge
-                    [:map [:x :string]]
+                    [:map [:x [:orn [:str :string]]]]
                     [:map [:y :int]]
                     [:schema [:map [:z {:optional true} :boolean]]]])]
         (is (= [:merge
-                [:map [:x :string]]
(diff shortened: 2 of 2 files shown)


src/malli/util.cljc
```diff
@@ -408,14 +408,18 @@
...
-      (let [[children forms schema] (fn properties (vec children) options)
+      (let [[children forms schema] (func properties (vec children) options)
             form (delay (m/-create-form type properties forms options))
             cache (m/-create-cache options)]
         ^{:type ::m/schema}
         (reify
           m/Schema
           (-validator [_] (m/-validator schema))
           (-explainer [_ path] (m/-explainer schema path))
+          (-parser [this]
```
ikitommi, line 419: shoudn't these be forwarded to the schema, e.g. `(m/parser schema)` and `(m/unparser schema)`?

src/malli/util.cljc
```diff
@@ -399,7 +399,7 @@
      (clojure.core/update children 0 #(m/form % options))
      (apply f (conj children options))]))
 
-(defn -util-schema [{:keys [type min max childs type-properties fn]}]
+(defn -util-schema [{:keys [type min max childs type-properties func]}]
```
ikitommi, line 402: why the rename?

ikitommi (changes requested):
please add a test to avoid future regression.

src/malli/util.cljc
abcdw: Yep, you are right, I already figured it out when was playing with new version. Updated implementation accordingly. Thank you!

src/malli/util.cljc
abcdw: It was conflicting with clojure.core/fn macro I used in first iteration, and because util-schema is private I decided to rename argument rather than using fully qualified name of the macro. Anyway, I reverted this change in the second iteration.

abcdw:
Updated implementation to use parser from original schema, added parser asserts to tests for each declarative util counterpart, rebased the branch.

ikitommi (approved):
Looks great!

ikitommi:
Big thanks for this.

PR #784, opened 2022-11-19
m/-infer + json-schema mapping

ikitommi:
```clojure
(json-schema/transform [:enum "perch" "pike"])
{:type "string", :enum ["perch" "pike"]}

(json-schema/transform [:enum 1 2 3])
{:type "integer", :enum [1 2 3]}
```


NoahTheDuke:
This is very cool.

Issue #786, opened 2022-11-23
Consider adding a clojure.spec multi-spec equivalent


jeroenvandijk:
In `clojure.spec` [**multi-spec**](https://clojure.org/guides/spec#_multi_spec) allows to dynamically extend specs. This gives a lot of flexibility. For example, it allows to create a plugin system where definitions are loaded at a later time, in a different part of the code. Quoting spec's [documentation](https://clojure.org/guides/spec#_multi_spec): 

> The multi-spec approach allows us to create an open system for spec validation

E.g. consider the small repl session below, it shows that a `spec` can be extended and redefined at runtime.

```clojure
(require '[clojure.spec.alpha :as s])

(defmulti event-type :event/type)
(s/def :event/event (s/multi-spec event-type :event/type))

(s/valid? :event/event
          {:event/type :event/search
           :search/url "https://clojure.org"}) ;=> false

(defmethod event-type :event/search [_]
  (s/keys :req [:event/type :search/url]))

(s/valid? :event/event
  {:event/type :event/search
   :search/url "https://clojure.org"}) ;=> true

;; Setting event-type to nil shows that the multispec is connected to the var and not the multimethod itself
(def event-type nil)
(s/valid? :event/event
          {:event/type :event/search
           :search/url "https://clojure.org"}) ;=> Exception

(defmulti event-type :event/type)
(s/valid? :event/event
          {:event/type :event/search
           :search/url "https://clojure.org"}) ;=> false
;; etc ..
```

In Malli there is the `:multi` schema. Currently this supports loading predefined keys lazily. It does not support a clean way to redefine, let alone add or remove keys at a later stage. In a Slack discussion with @ikitommi it was suggested to add a `:methods` option to dynamically define the keys. I tried this in a seperate PR (I'll add link in a moment). It was also implied that it should be a small addition. Maybe I misunderstood the suggestion, but I failed to do it in a way that adds little code.

If I have to think of an implementation similar to `multispec` it has to be with a dynamic entry parser (not just lazy) and maybe `add-watch` on the multimethod var.  The [`clojure.spec` implementation](https://github.com/clojure/spec.alpha/blob/aab86721189f2193f8bab6ea4d8966cce772a8b4/src/main/clojure/clojure/spec/alpha.clj#L948) looks different though. And after some experimentation I noticed that adding a `defmethod` doesn't trigger a var change, so `add-watch` would work here.

[1] https://clojurians.slack.com/archives/CLDK6MFMK/p1669045351866979

[...]

PR #787, opened 2022-11-23
Add :methods option to :multi

jeroenvandijk:
An experiment to have multispec like support, but I found out it is a little bit more complicated.

See #786 for the background

The code addition in this PR doesn't satisfy my request in #786 and therefore I marked this PR as a draft

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -473,17 +473,23 @@
           (recur (int (-parse-entry (aget ca i) naked-keys lazy-refs options i -children -forms -keyset))
                  (unchecked-inc-int ci)))))))
 
-(defn -lazy-entry-parser [?children props options]
-  (let [parser (delay (-eager-entry-parser ?children props options))]
-    (reify EntryParser
-      (-entry-keyset [_] (-entry-keyset @parser))
-      (-entry-children [_] (-entry-children @parser))
-      (-entry-entries [_] (-entry-entries @parser))
-      (-entry-forms [_] (-entry-forms @parser)))))
+
+(defn -lazy-entry-parser [?children props options]    
+    (let [children-fn (fn []
+                        (if-let [method-lookup (:methods props)]
+                          (into [] (concat ?children (method-lookup)))
+                          ?children))
+          parser (delay (-eager-entry-parser (children-fn) props options))]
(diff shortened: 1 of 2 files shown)


ikitommi:
I see. I think it would be better to have a Schema wrapper that allows one to create the current snapshot of the `Schema` instance programmatically. This would work with all Schema types, not just with `:multi`.

Did a prototype, works like this:

```clojure
(def schema* (atom :int))

(def mut (mu/mutable #(m/schema @schema*)))

(m/form [:map [:mut mut]])
; => [:map [:mut :int]]

(reset! schema* [:enum "so" "mutable"])

(m/form [:map [:mut mut]])
; => [:map [:mut [:enum "so" "mutable"]]]
```

with multimethods:

```clojure
(defmulti my-schema :type)

(defmethod my-schema :user [_]
  [:map
   [:type [:= :user]]
   [:name :string]])

(def multi
  (mu/mutable
   #(m/into-schema
     :multi
     {:dispatch :type}
     (map (fn [[type f]] [type (f {:type type})]) (methods my-schema)))))

multi
;[:multi {:dispatch :type}
; [:user [:map
;         [:type [:= :user]]
;         [:name :string]]]]

(m/validate multi {:type :user, :name "kikka"})
; => true

(m/validate multi {:type :fruit, :taste "sweet"})
; => false

(defmethod my-schema :fruit [_]
  [:map
   [:type [:= :fruit]]
   [:taste [:enum "sweet" "sour"]]])

multi
;[:multi
; {:dispatch :type}
; [:fruit [:map
;          [:type [:= :fruit]]
;          [:taste [:enum "sweet" "sour"]]]]
; [:user [:map
;         [:type [:= :user]]
;         [:name :string]]]]

(m/validate multi {:type :user, :name "kikka"})
; => true

(m/validate multi {:type :fruit, :taste "sweet"})
; => true
```

the `Schema` instance is mutable itself, with no caching enabled, so calling `(m/validate multi ...)` always uses the latest value. Calling `(m/validator multi)` will create a immutable snapshot with the current value.

would think work with you?

ikitommi:
... moved the response into #786

jeroenvandijk:
See #786

PR #790, opened 2022-12-02
Fix enum ast losing properties and options

rralston-sfin:
With the following enum schema 
```clojure
(def enum-example
    [:enum {:encode/json name :decode/json keyword} :e-1 :e-2])
```
`(properties (from-ast (ast enum-example)))` returns `nil` rather than `{:encode/json name :decode/json keyword}`.

The fix persists the properties and options from the schema into the ast.


ikitommi:
thanks!

PR #792, opened 2022-12-06
Fix recursive `:+` generators

frenchy64:
`:+` generators with recursive variables can become `[:+ :UNREACHABLE]`, which corresponds to the impossible generator `(gen/not-empty (gen/return ()))`. In this case, it should reduce to `(-never-gen)`.

Added a useful debugging namespace that generates AST's for generators, which is how I found the root problem. Because of the way it's generated, I had to qualify some keywords using an alias instead of auto-namespacing.

--------------------

Original report: https://clojurians.slack.com/archives/CLDK6MFMK/p1670138741691259

```
Carlo:
A question about recursive generators and :+. Consider this recursive spec that describes a boolean formula:
(def formula
  [:schema
   {:registry
    {::formula
     [:or
      :boolean
      [:tuple [:enum :not] :boolean]
      [:tuple [:enum :and] [:* [:ref ::formula]]]
      [:tuple [:enum :or]  [:* [:ref ::formula]]]]}}
   [:ref ::formula]])
We can of course generate examples via:
(repeatedly 20 #(mg/generate formula))
But, if I change :* with :+ (rest in :thread:)

I now have:
(def formula
  [:schema
   {:registry
    {::formula
     [:or
      :boolean
      [:tuple [:enum :not] :boolean]
      [:tuple [:enum :and] [:+ [:ref ::formula]]]
      [:tuple [:enum :or]  [:+ [:ref ::formula]]]]}}
   [:ref ::formula]])

(repeatedly 20 #(mg/generate formula))
and the generator for this seems broken to me (for starters, it usually doesn't return 20 things. I suspect the reason is that it makes more difficult to generate values (as a naive sampling terminates more rarely). Is that correct?

By a brute force search, I saw that this problem is solved if I wrap :+ into a :schema. A more thorough explanation would be useful though
```

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -1,3 +1,6 @@
+;; See also `malli.generator-ast` for viewing generators as data
+;; Note: use `::mg/foo` instead of `::foo` in this namespace
+;; to assist in generating `malli.generator-ast`.
 (ns malli.generator
   (:require [clojure.spec.gen.alpha :as ga]
             [clojure.string :as str]
@@ -8,6 +11,7 @@
             [clojure.test.check.rose-tree :as rose]
             [malli.core :as m]
             [malli.registry :as mr]
+            [malli.generator :as-alias mg]
             #?(:clj [borkdude.dynaload :as dynaload])))
 
 (declare generator generate -create)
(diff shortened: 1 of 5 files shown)


ikitommi:
Thanks! The `generator-debug` is really nice. It would be useful as part of test.check itself? Or as a standalone library? It definitely can be part of malli, but I think people would find it useful in a larger context too.

Anyway, big thanks for resolving this.

PR #794, opened 2022-12-06
Minor generator optimizations, add generator code generator

frenchy64:
Perhaps these little optimizations might be significant in large cases. `generator-ast` made at least their ugliness painfully obvious, as you can see in the updated tests (this is more obvious in `test/malli/generator_test.cljc` after whitespace changes are hidden).

1. Use `gen/return` for singleton `gen/elements` calls.
2. Use constrained `gen/vector` instead of `(gen/not-empty (gen/vector ..))`.

Add dev helper `generator-code` to help generate the large `recursive-gen` test cases.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -92,12 +92,19 @@
 
 (defn- -double-gen [options] (gen/double* (merge {:infinite? false, :NaN? false} options)))
 
+(defn- gen-vector-min [gen min options]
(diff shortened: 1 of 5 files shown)


ikitommi:
Thanks! I think `gen-one-of` and `gen-elements` would be good optimizations in `test.check` itself. Maybe try a PR there?

PR #795, opened 2022-12-06
Add Sequence Schemas to JSON Schema generation

### LINKED ISSUE #793: `:?` vs `:sequential` in JSON Schema transformers
Hey friends, I've found a small difference between `[:sequential {:max 1} schema]` and `[:? schema]` when transformed with `malli.json-schema`:

```clojure
(m/encode
  (m/schema
    [:map {:closed true}
     [:data [:? [:map {:closed true}
                 [:id :int]
                 [:description [:maybe :string]]
                 [:name :string]]]]])
  {:data [{:id 1
           :description nil
           :name "hello"
           :type "extra-type"
           :extra "extra-key"}]}
  mt/strip-extra-keys-transformer)
; {:data
;  [{:id 1,
;    :description nil,
;    :name "hello",
;    :type "extra-type",
;    :extra "extra-key"}]}
(m/encode
  (m/schema
    [:map {:closed true}
     [:data [:sequential {:max 1}
             [:map {:closed true}
              [:id :int]
              [:description [:maybe :string]]
              [:name :string]]]]])
  {:data [{:id 1
           :description nil
           :name "hello"
           :type "extra-type"
           :extra "extra-key"}]}
  mt/strip-extra-keys-transformer)
; {:data [{:id 1, :description nil, :name "hello"}]}
```
Is this intentional? If not, I can dig in to try to fix it.

NoahTheDuke:
Closes #793

Tried to stick with how things work in the existing code. Looks like Malli's generated JSON Schema isn't up to date with the most recent JSON Schema specs, so I'm not entirely sure I did this right.


ikitommi:
Does this work correctly with inlined sequence scehmas? e.g. `[:cat [:cat :string] [:+ [:+ :string]]]` presents a flat list of strings.

NoahTheDuke:
Oh that's a good point. I didn't know that Malli inlined those schemas. Is there a way to apply that inlining upfront?

ikitommi:
I don't think there is. Some options:

## 1) resolve locally

* return sequence schema json schemas maps a tag "these can be inlined", e.g. `(with-meta cat-json-schema {::regex-schema true})` and inline those in the with sequence schemas json schema functions
* GOOD: simple
* BAD: not generic

## 2) just mark them as "arrays of anything"

* GOOD: no work
* BAD: not describing the real thing

## 3) schema simplifier

* this has been drafted few times already: given any schema, return a simplified schema. e.g. `[:and :keyword :keyword keyword?]` is just `:keyword`. Could be used here too. If done via a protocol, could be just implemented first for sequence schemas
 
* GOOD: the ultimate solution
* BAD: lot of work
* !!!: should not increase JS-bundle size, e.g. not used by default, optional thing that can be DCE'd
 
@frenchy64 had one (partial) solution for this, with the first version of the recursive regex.

ikitommi:
here's one old test to this: https://github.com/miikka/boolean-simplifier

meander has good tools for rewriting code.

NoahTheDuke:
After thinking about it for a little while, I feel like 1) is the way to go. 2) is basically what Malli has now which is unintuitive ("why doesn't json schema transformation work?").

3\) doesn't work because `malli.json-schema/-transform` uses `m/walk` which is a post-walk. This means we'd have to walk the schema 2+ times (first to do any merging/inlining, then second to transform resulting schema).

1\) can be done in a single pass in `malli.json-schema/accept`, where the sequence schemas could add metadata and check their children for the metadata and merge as necessary.

What do you think?

NoahTheDuke:
I realize now that I basically said the same thing as you regarding 1). lol Sorry about that. I don't think it's hacky. I don't know that there's a good abstraction that doesn't require a full rewrite of the existing json schema architecture.

ikitommi:
If you think you can make it work with 1, let's go with that.

NoahTheDuke:
I spent some time on this, and I don't know that the "resolve locally" solution is possible. I find myself building an ad-hoc version of "schema simplifier", which I don't have the skill or time to develop the skill to properly implement.

Given that, how would you feel about some sort of "NOT CURRENTLY WORKING" warning when converting a sequence schema to json schema? They currently just return `{}`, which I think is reasonable but surprising without any warning.

ikitommi:
Thanks for giving it a shot. I think a reasonable default would be a vector of anything. At least it's of right type.

When we have a generic simplifier, we can make this better.

ikitommi:
could you change these to return "vector of anything"?

ikitommi:
ping @NoahTheDuke

mattiuusitalo:
@NoahTheDuke is this still relevant?

opqdonut:
Closing as inactive.

NoahTheDuke:
Ah dang, sorry for never following up on this. I stopped working at the job that used malli to generate json schemas, and don't have a good way to test this anymore. Maybe some day I'll pick it back up, but not in the foreseeable future.

PR #809, opened 2023-01-03
mr/set-default-registry! takes varargs

ikitommi:
Not sure if this is a good idea, but this:

```clojure
(mr/set-default-registry!
  ;; linear search
  (mr/composite-registry
    ;; immutable registry
    {:map (m/-map-schema)}
    ;; mutable (spec-like) registry
    (mr/mutable-registry registry)
    ;; on the perils of dynamic scope
    (mr/dynamic-registry)))
```

... could be written as:

```clojure
;; linear search
(mr/set-default-registry!
 ;; immutable registry
 {:map (m/-map-schema)}
 ;; mutable (spec-like) registry
 (mr/mutable-registry registry)
 ;; on the perils of dynamic scope
 (mr/dynamic-registry))
```

comments welcome.


ikitommi:
Maybe not. Saving few keystrokes.

PR #814, opened 2023-01-07
Tweaks to malli experimental describe

escherize:
1. Using "a" prefix does not compose, so it's been removed. 
    - (e.g. `[:vector map?]` => "a map of a vector", is now fixed)
2. Differentiate between min/max length of collection vs. min/max numeral constraints
3. a couple linting fixes
1. adds repeat verbiage that makes sense
1. removed [duplicate int defmultis](https://github.com/metosin/malli/pull/814/files#diff-91e4afd493bcf0c63e0117f025a9f42e7e44123a813ef2f9f18ac0fa2202a222L191-L194)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2944,8 +2944,7 @@ You can call describe on a schema to get its description in english:
 (med/describe [:map {:closed true}
                [:x {:optional true} int?]
                [:y :boolean]])
-
-;;=> "a map where {:x (optional) -> <integer>, :y -> <boolean>} with no other keys"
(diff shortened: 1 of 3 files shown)


ikitommi:
Could you rebase the changes with master, which has already some (style) changes.

ikitommi (changes requested):
please rebase from master, tests broken.

ikitommi (approved):

ikitommi:
Looks great, thanks!

PR #825, opened 2023-01-26
Turn on instrumentation for mx/defn with :malli/always meta

### LINKED ISSUE #823: `mx/defn` improvement: always-check inputs and outputs
It's often surprising to users when mx/defn behaves differently than (prismatic) schema.core/defn. Probably because they have very similar syntax.

It'd be cool if mx/defn provided a way to always check inputs and outputs [like this defn macro does](https://github.com/metabase/metabase/blob/bcm-reopen-chain-graph-data-permissions/src/metabase/util/malli.clj#L35-L78).

```clojure
(mx/defn ^:always-check g :- :int [x :- :keyword] 3)

(g "non-keyword")
;; ==throws==> clojure.lang.ExceptionInfo
;; Message: 
;; "[[\"should be a keyword\"]]"
;; Data: 
;; {:type :malli.core/invalid-input,
;;  :data
;;  {:input [:cat :keyword], :args ["non-keyword"], :schema [:=> [:cat :keyword] :int], :fn-name malli.experimental/g},
;;  :humanized [["should be a keyword"]]}
```

### Open Questions

Q: Is there another piece of information that would be good to include here?
Q: Is it worth differentiating between `^:always-check-inputs` and `^:always-check-outputs`?

opqdonut:
Fixes #823


opqdonut:
Need to think about how `^:malli/always-check` interacts with `malli.dev/start!`. Currently `malli.dev/start!` reinstruments the function and disables the throwing!

opqdonut:
With my current approach, the malli.dev instrumentation goes "on top" of the always-check instrumentation. So after running malli.dev/start!, you get both the nice printed error and the exception:

```
user=> (require '[malli.experimental :as mx])
nil
user=> (mx/defn ^:malli/always-check f :- :int [x :- :int] x)
#'user/f
user=> (f 1)
1
user=> (f "1")
Execution error (ExceptionInfo) at malli.core/-exception (core.cljc:138).
:malli.core/invalid-input
user=> (require 'malli.dev)
nil
user=> (malli.dev/start!)
..instrumented #'user/f
started instrumentation
nil
user=> (f 1)
1
user=> (f "1")
-- Schema Error ---------------------------------------------------------------- NO_SOURCE_FILE:1 --

Invalid function arguments:

  ["1"]

Function Var:

  user/f

Input Schema:

  [:cat :int]

Errors:

  {:in [0], :message "should be an integer", :path [0], :schema :int, :value "1"}

More information:

  https://cljdoc.org/d/metosin/malli/CURRENT/doc/function-schemas

----------------------------------------------------------------------------------------------------
Execution error (ExceptionInfo) at malli.core/-exception (core.cljc:138).
:malli.core/invalid-input
```

docs/function-schemas.md
```diff
@@ -479,6 +479,23 @@ with instrumentation:
 ; =throws=> :malli.core/invalid-input {:input [:cat :int [:int {:max 6}]], :args [10 10], :schema [:=> [:cat :int [:int {:max 6}]] :int]}
 ```
 
+... or by using metadata `^:malli/always-check`:
```
ikitommi, line 482: I would use `:malli/always-validate` (like plumatic) or even `:malli/always` (aesthetics). I like the latter as it's shorter.

src/malli/experimental.cljc
```diff
@@ -3,7 +3,8 @@
   #?(:cljs (:require-macros malli.experimental))
   (:require [clojure.core :as c]
             [malli.core :as m]
-            [malli.destructure :as md]))
+            [malli.destructure :as md]
+            [malli.instrument :as mi]))
```
ikitommi, line 7: `malli.instrument` pulls `malli.generator`, which pulls `test.check`. These include lot's of things that can't be DCEd (e.g. multimethods). This adds ~100KB of Javascript (uncompressed). Malli starts with ~10KB (uncompressed), so this matters.

Using `mx/defn` without always-validate should not contribute to js code size.

sample report, before:

```
| Resource                            |   Optimized |  Total % |          JS |      Source |
|-------------------------------------+-------------+----------+-------------+-------------|
| jar | cljs/core.cljs                |    156,9 KB |  53,71 % |      1,3 MB |    348,0 KB |
|     | malli/core.cljc               |    100,4 KB |  34,38 % |    628,2 KB |    115,3 KB |
|     | malli/impl/regex.cljc         |     15,4 KB |   5,28 % |     97,6 KB |     26,5 KB |
| jar | borkdude/dynaload.cljc        |      6,8 KB |   2,34 % |     15,6 KB |      8,8 KB |
|     | malli/destructure.cljc        |      4,1 KB |   1,41 % |     37,4 KB |      7,4 KB |
|     | malli/registry.cljc           |      2,5 KB |   0,84 % |     26,0 KB |      2,8 KB |
|     | malli/experimental.cljc       |      2,1 KB |   0,73 % |     13,2 KB |      2,6 KB |
|     | malli/sci.cljc                |      1,4 KB |   0,49 % |      3,9 KB |    751,0 B  |
|     | malli/impl/util.cljc          |    947,0 B  |   0,32 % |      4,5 KB |      2,9 KB |
|     | malli/app.cljc                |    563,0 B  |   0,19 % |      1,8 KB |    396,0 B  |
| jar | goog/string/stringbuffer.js   |    341,0 B  |   0,12 % |      2,3 KB |      2,3 KB |
| jar | goog/object/object.js         |    251,0 B  |   0,09 % |     21,4 KB |     21,4 KB |
| jar | goog/base.js                  |    235,0 B  |   0,08 % |    128,7 KB |    128,7 KB |
| jar | goog/string/internal.js       |     40,0 B  |   0,01 % |     12,3 KB |     12,3 KB |
```

after:

```
| Resource                                                    |   Optimized |  Total % |          JS |      Source |
|-------------------------------------------------------------+-------------+----------+-------------+-------------|
| jar | cljs/core.cljs                                        |    189,1 KB |  47,21 % |      1,3 MB |    348,0 KB |
|     | malli/core.cljc                                       |    113,9 KB |  28,45 % |    628,2 KB |    115,3 KB |
| jar | clojure/test/check/generators.cljc                    |     17,6 KB |   4,41 % |    114,6 KB |     66,0 KB |
|     | malli/impl/regex.cljc                                 |     15,4 KB |   3,86 % |     97,6 KB |     26,5 KB |
|     | malli/generator.cljc                                  |     13,1 KB |   3,27 % |     69,4 KB |     27,8 KB |
| jar | cljs/spec/gen/alpha.cljs                              |     11,8 KB |   2,96 % |    101,5 KB |      6,1 KB |
| jar | borkdude/dynaload.cljc                                |      6,8 KB |   1,70 % |     15,6 KB |      8,8 KB |
| jar | goog/uri/uri.js                                       |      4,8 KB |   1,21 % |     44,2 KB |     44,2 KB |
|     | malli/destructure.cljc                                |      4,1 KB |   1,03 % |     37,4 KB |      7,4 KB |
| jar | goog/math/long.js                                     |      3,9 KB |   0,98 % |     28,1 KB |     28,1 KB |
|     | malli/registry.cljc                                   |      3,5 KB |   0,87 % |     26,0 KB |      2,8 KB |
| jar | clojure/test/check/rose_tree.cljc                     |      2,8 KB |   0,69 % |     16,8 KB |      6,4 KB |
| jar | clojure/test/check/random.cljs                        |      2,4 KB |   0,60 % |     14,6 KB |      4,9 KB |
|     | malli/experimental.cljc                               |      2,1 KB |   0,54 % |     13,7 KB |      2,7 KB |
| jar | clojure/test/check/random/longs/bit_count_impl.cljs   |      1,5 KB |   0,38 % |      2,6 KB |      1,6 KB |
|     | malli/sci.cljc                                        |      1,5 KB |   0,36 % |      3,9 KB |    751,0 B  |
| jar | clojure/string.cljs                                   |      1,3 KB |   0,32 % |     14,7 KB |      8,4 KB |
|     | malli/impl/util.cljc                                  |    970,0 B  |   0,24 % |      4,5 KB |      2,9 KB |
| jar | goog/structs/structs.js                               |    868,0 B  |   0,22 % |     12,4 KB |     12,4 KB |
| jar | clojure/test/check/random/longs.cljs                  |    610,0 B  |   0,15 % |      3,1 KB |      4,2 KB |
|     | malli/app.cljc                                        |    521,0 B  |   0,13 % |      1,9 KB |    396,0 B  |
| jar | goog/uri/utils.js                                     |    361,0 B  |   0,09 % |     38,1 KB |     38,1 KB |
| jar | goog/string/stringbuffer.js                           |    349,0 B  |   0,09 % |      2,3 KB |      2,3 KB |
| jar | goog/base.js                                          |    327,0 B  |   0,08 % |    128,7 KB |    128,7 KB |
| jar | goog/string/string.js                                 |    259,0 B  |   0,06 % |     47,8 KB |     47,8 KB |
| jar | goog/object/object.js                                 |    130,0 B  |   0,03 % |     21,4 KB |     21,4 KB |
| jar | clojure/test/check/random/doubles.cljs                |    109,0 B  |   0,03 % |    983,0 B  |      1,2 KB |
| jar | goog/array/array.js                                   |     82,0 B  |   0,02 % |     62,1 KB |     62,1 KB |
| jar | goog/string/internal.js                               |     81,0 B  |   0,02 % |     12,3 KB |     12,3 KB |
|     | malli/instrument.cljs                                 |     42,0 B  |   0,01 % |     67,3 KB |      8,7 KB |
```

src/malli/experimental.cljc
```diff
@@ -47,10 +51,13 @@
...
                    ~name
                    ~@(some-> doc vector)
-                   ~(assoc meta :raw-arglists (list 'quote raw-arglists), :schema schema)
+                   ~(assoc body-meta :raw-arglists (list 'quote raw-arglists), :schema schema)
                    ~@(map (fn [{:keys [arglist prepost body]}] `(~arglist ~prepost ~@body)) parglists)
                    ~@(when-not single (some->> arities val :meta vector)))]
        (m/=> ~name ~schema)
+       ~(when (or (:malli/always-check var-meta)
+                  (:malli/always-check body-meta))
+          `(alter-var-root (var ~name) (fn [f#] (m/-instrument {:schema ~schema} f#))))
```
ikitommi, line 60: what if we wrapped the body with `m/-instrument` instead?

test/malli/experimental_test.clj
```diff
@@ -141,11 +142,52 @@
 
     (when instrumented
       (testing "instrumented calls"
-        (-strument! :instument var)
```
ikitommi, line 144: 🏆

ikitommi (changes requested):
Thanks for working on this. Quick comments attached.

src/malli/experimental.cljc
opqdonut: Oh, right, `malli.instrument` is actually an unused require now that I switched to `m/-instrument`. I'll remove it.

src/malli/experimental.cljc
opqdonut: I thought about that but generating the code for the `fn` to put inside `m/-instrument` seemed like a lot of work.

I also considered giving a name to this alter-var-root spell, something like `instrument-var!`

docs/function-schemas.md
opqdonut: `:malli/always` is fine by me

dvingo:
There is some prior work in this PR that is designed with cljs support in mind:
https://github.com/metosin/malli/pull/702

I think emitting a `defn` that delegates to the instrumented version is much simpler to deal with due to how clojurescript compilation works (multiple js functions can be emitted per one `defn` form). I also don't really know of another way because `alter-var-root` isn't available in cljs.

That PR also has some cljs tests you can copy over to ensure the var meta is still preserved and multi-arity forms work as expected.

It would be awesome to have support for this syntax in cljs too :)

opqdonut:
Huh, thanks! And sorry for not noticing your PR. That looks like a nice implementation to me. The only downside is that there's a duplicated defn impl. I'll update this PR with a similar implementation and some cljs tests.

opqdonut:
I ran into some problems trying to run all of malli.experimental-test in cljs mode, so I had to disable some tests. I'll try to get them enabled next.

src/malli/experimental.cljc
opqdonut: I ended up doing this thanks to @dvingo

dvingo:
No worries! That PR isn't directly related and is now quite old, but thanks for updating this PR!

I think the cljs test failures may be from attempting to get `meta` from a var at runtime - there are not real Vars in cljs. I haven't run the code, but looking at what is there, I think this line:
`(is (= v (k (meta var)))`
Is problematic in cljs, you can only use symbol literals - no indirection allowed because this call has to evaluate at compile time in a macro.

So to achieve this you may be able to put `:var-meta (meta f1)` etc in the `expectations` data and then use that in the assertion. I _think_ that is the issue, but I can try it out later tonight to check.

opqdonut:
Right, that's a good catch. I wonder if the instrumentation failures are because `expectations` refers to the vars, so they get stuck with the first definitions of the functions.

dvingo:
@opqdonut I played around with some cljs fixes and have the tests passing.
https://github.com/dvingo/malli/commit/2c8d99a906820fd81b466f4b6bfc4b44b0eac479

Some of the issues were that JS is just more lenient so the `:call` checks don't throw on a lot of the bad input when instrumentation is off. The other piece was that in cljs var metadata is not evaluated for almost all circumstances, which is the quote issue you ran into.

Everything is looking good aside from `f6` which is the multi-arity with varags function. The instrumented function arities don't seem to be present. There are already existing tests for this in malli.instruments-test.cljs so I think this is due to the complexities of how vars and multi-arity functions are compiled and would take a while to debug. I don't think users will be doing this sort of dynamic instrumentation with cljs so I'm fine with this for now.

opqdonut:
Yeah I also had everything passing except f6. I wonder if the best solution here would be to just create a new `malli.experimental.cljc-test` namespace with some simple cljc cases that pass, and leave the current `defn-test` out of it.

opqdonut:
I went over this with @ikitommi and we agreed to just run my new tests for cljs. No use in making the existing `defn-test` cljs-compatible.

ikitommi (approved):
Looks Great!

PR #827, opened 2023-01-28
Add clj-kondo support for cljs function schemas

dvingo:
Figured out a way to piggy-back on the shadow-cljs implementation in order to get the cljs schemas in a browser access to the filesystem so kondo config can be saved!

One question this brings up is when you're running `malli.dev/start!` from both clj and cljs the last one to write 
the file will win. I'm not sure a good solution other than attempting to perform a merge with the existing schemas instead of overwriting them.

https://github.com/metosin/malli/issues/828

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -36,7 +36,7 @@
                         jmh-clojure/task {:mvn/version "0.1.1"}}
                  :main-opts ["-m" "jmh.main"]}
            :shadow {:extra-paths ["app"]
-                    :extra-deps {thheller/shadow-cljs {:mvn/version "2.20.5"}
+                    :extra-deps {thheller/shadow-cljs {:mvn/version "2.20.20"}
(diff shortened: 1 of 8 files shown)


dvingo:
hmm, just realized there is an annoying issue with this approach.
when you don't add the hook you'll see a shadow-cljs warning:

```
2023-01-28 10:54:31.730 - WARNING] :shadow.cljs.devtools.server.worker.impl/unhandled-op - {:op :malli.clj-kondo/write-config, :build-id :main, :data {:linters {:unresolved-symbol {:exclude [(malli.core/=>)]}, :type-mismatch {:namespaces {co.app.model {make-entity {:arities
```
I have another approach I can try, where the build hook can use `cljs-eval` to ask the browser for the kondo data. I suppose going with that approach will be better then to not have to deal with this warning.

dvingo:
closing in favor of https://github.com/metosin/malli/pull/829

Issue #828, opened 2023-01-28
Add support for clj-kondo for cljs function schemas


dvingo:
Add the ability to write clj-kondo config for clojurescript function schemas.

PR #829, opened 2023-01-29
[DRAFT] Add clj-kondo support for cljs function schemas

dvingo:
https://github.com/metosin/malli/issues/828

UPDATE:
In chatting with Thomas Heller I think there is a simpler solution that doesn't need a hook. I will attempt that next.

Using a shadow-cljs hook we can send the clj-kondo config for cljs function schemas to the shadow-cljs clojure runtime to persist to disk.

A counter is used to track when the cljs is not only compiled but evaluated so that the clojure side (the hook) knows that the kondo config will reflect any code changes during hot reload.

There may be a way to determine when the latest code has been evaluated by a client from shadow-cljs's build state but I could not find a place where that information is available, thus the counter technique with a busy loop is used.

If you do not include the build hook there is no effect on the build. If you pass a `build-id` that doesn't actually exist in you `shadow-cljs.edn` config, the hook will effectively be a no-op as all `eval-cljs` calls wil return `nil`.

There is a potential problem if you have instrumentation running for clojure and clojurescript code at the same time in that they will overwrite each others' kondo config file.

I think it would make sense to have a separate output file for each side, one for cljs, one for clj. What do you think?

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -36,7 +36,7 @@
                         jmh-clojure/task {:mvn/version "0.1.1"}}
                  :main-opts ["-m" "jmh.main"]}
            :shadow {:extra-paths ["app"]
-                    :extra-deps {thheller/shadow-cljs {:mvn/version "2.20.5"}
+                    :extra-deps {thheller/shadow-cljs {:mvn/version "2.20.20"}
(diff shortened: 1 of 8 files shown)


dvingo:
closing in favor of https://github.com/metosin/malli/pull/833

PR #833, opened 2023-02-01
Add clj-kondo support for cljs function schemas

dvingo:
Fixes https://github.com/metosin/malli/issues/828

After lots of help from Thomas Heller [here](https://app.slack.com/client/T03RZGPFR/C6N245JGG/thread/C6N245JGG-1674780762.779839) I was able to put together a much simpler dev experience for cljs.

This PR adds kondo config and updates the docs with Thomas' recommended setup using a preload and namespace ordering by the compiler to ensure your function schemas are up to date when `malli.dev.cljs/start!` is evaluated at compile-time.

I updated the kondo `save!` function to emit to a different location for clj and cljs schemas so you can seamlessly use both during development. Let me know if you have any concerns or feedback about that .

### DIFF
--- a/app/malli/dev_preload.cljs
+++ b/app/malli/dev_preload.cljs
@@ -1,0 +1,7 @@
+(ns malli.dev-preload
+  {:dev/always true}
+  (:require
+    [malli.instrument-app]
+    [malli.dev.cljs :as dev]))
+
+(dev/start!)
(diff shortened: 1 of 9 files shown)


thheller (changes requested):
Small tweaks, then it should be fine.

dvingo:
Thanks @thheller ! updated per your feedback : )

thheller (approved):
Looks good. Although in plugin part you can just call the above function instead of repeating the code.

dvingo:
> n part you can just call the above function instead of repeating the code.

ah of course - updated

ikitommi:
This is really well though though, implemented and documented 🙇

PR #841, opened 2023-02-03
Allow instrumenting external functions

### LINKED ISSUE #840: Instrumenting external functions
Hey friends,

Sometimes I want to add instrumentation to code I haven't written, such as `clojure.core` functions or a library that doesn't check its inputs as closely as I would hope. Currently, that's not possible because of the way [`m/=>` is written](https://github.com/metosin/malli/blob/cab4fbfa70d01aa3d91aa4d597cd5b0a9103d6d4/src/malli/core.cljc#L2471). I think this should be possible, however, by calling `name` on the provided symbol and then using the namespace of the symbol (defaulting to `*ns*`) as the symbol.

NoahTheDuke:
~Here's a currently broken attempt at fixing #840. I don't know enough about cljs to know how to solve it, but for clj it works well enough.~

I should note that in clj, it doesn't work for any function that is "inlined". So, if you say `(m/=> clojure.core/inc [:=> [:cat :int] :int])`, calling `(inc "1")` won't be instrumented as it's compiled to a raw `Numbers.inc("1")` call. However, writing `(#'inc "1")` will be instrumented as it will be compiled into `const__0.invoke("i")`.

I think that if this is noted in the docs it's an acceptable trade-off.

Closes #840

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2469,22 +2469,24 @@
    (swap! -function-schemas* assoc-in [key ns name] (merge data {:schema (f ?schema), :ns ns, :name name}))))
 
 #?(:clj
-   (defmacro => [name value]
+   (defmacro => [given-sym value]
      (let [cljs-resolve (when (:ns &env) (ns-resolve 'cljs.analyzer.api 'resolve))
            cljs-resolve-symbols (fn [env d]
                                   (walk/postwalk (fn [x] (cond->> x (symbol? x) (or (:name (cljs-resolve env x)))))
                                                  d))
-           name' `'~(symbol (str name))
-           ns' `'~(symbol (str *ns*))
-           sym `'~(symbol (str *ns*) (str name))
+           name-str (name given-sym)
+           ns-str (str (or (not-empty (namespace given-sym)) *ns*))
+           name' `'~(symbol name-str)
+           ns' `'~(symbol ns-str)
+           sym `'~(symbol ns-str name-str)
            value' (cond->> value (:ns &env) (cljs-resolve-symbols &env))]
        ;; in cljs we need to register the schema in clojure (the cljs compiler)
        ;; so it is visible in the (function-schemas :cljs) map at macroexpansion time.
        (if (:ns &env)
          (do
-           (-register-function-schema! (symbol (str *ns*)) name value' (meta name) :cljs identity)
-           `(do (-register-function-schema! ~ns' ~name' ~value' ~(meta name) :cljs identity) ~sym))
-         `(do (-register-function-schema! ~ns' ~name' ~value' ~(meta name)) ~sym)))))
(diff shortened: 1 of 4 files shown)


src/malli/core.cljc
```diff
@@ -2469,22 +2469,24 @@
...
            value' (cond->> value (:ns &env) (cljs-resolve-symbols &env))]
        ;; in cljs we need to register the schema in clojure (the cljs compiler)
        ;; so it is visible in the (function-schemas :cljs) map at macroexpansion time.
        (if (:ns &env)
          (do
-           (-register-function-schema! (symbol (str *ns*)) name value' (meta name) :cljs identity)
-           `(do (-register-function-schema! ~ns' ~name' ~value' ~(meta name) :cljs identity) ~sym))
-         `(do (-register-function-schema! ~ns' ~name' ~value' ~(meta name)) ~sym)))))
+           (-register-function-schema! (symbol ns-str) (symbol name-str) value' (meta given-sym) :cljs identity)
+           `(do (-register-function-schema! ~ns' ~name' ~value' ~(meta given-sym) :cljs identity) ~sym))
```
dvingo, line 2488: line 2488 should have the same value as line 2487 (`ns-str` and `name-str`) then I think this should work in cljs.

test/malli/instrument_test.cljs
```diff
@@ -196,3 +196,22 @@
 (deftest ^:simple check-test
   (let [results (mi/check)]
     (is (map? results))))
+
+(deftest instrument-external-test
```
dvingo, line 200: you'll need to mark this test with `^:simple` to not run during advanced compilation on cljs.

NoahTheDuke: Interesting, did instrumentation not work with advanced compilation?

dvingo: Not this version (as opposed to the `malli.instrument.cljs` ns which uses macros) because it is implemented to replace the JS functions at runtime by setting their value by name lookup. After advanced compilation the names will all be minified and there is no way (that I know of) to access the minified name.

src/malli/core.cljc
NoahTheDuke: If I change this to match 2487, other stuff breaks. To be clear, they're fundamentally doing the same thing, but `(symbol ns-str)` is executed during macroexpansion/eval, and `~ns'` is macroexpanded to the symbol value of `ns-str` (roughly `(list 'quote ~(symbol ns-str))`) and executed at run-time.

NoahTheDuke: To fix the test, I had to change the namespace from `clojure.core` to `cljs.core`.

dvingo: ahh ok thanks for clarifying - I see what's happening now!

ikitommi:
Looks good and a change most welcome, thanks!

Issue #689, opened 2022-04-09
suppress println or log/debug when instrumenting?


csummers:
Is there any chance the following lines could be optionally suppressed or be changed to  `log/debug`?  When using `malli.dev/start!` on a large codebase, these create a lot of print statements to stdout/repl.

https://github.com/metosin/malli/blob/3599fbd7fe1eba96c35e5c39073397cba0984b6a/src/malli/instrument.clj#L27

https://github.com/metosin/malli/blob/3599fbd7fe1eba96c35e5c39073397cba0984b6a/src/malli/instrument.clj#L31

Thanks!

ikitommi:
there is no logging library used, but PR welcome to add an option that disables the printing.

aiba:
I also am getting a lot of terminal output from these printlns (a testimant to how useful we find malli.experimental/defn!).

I was going to take a stab at a PR to make this optional, but I'm stuck unable to understand how `malli.instrument/-strument` is getting called by `malli.experimental/defn`.  Perhaps someone could help with this?

I'd also be happy to just ditch the printlns if it's too impractical to make them optional.

PR #842, opened 2023-02-07
Make `instrument!` less verbose

gpind:
Addresses #689 by making `-strument!` return the affected vars instead of printing them, and then makes `malli.dev` print the count of affected vars when it calls instrument/unstrument.

### DIFF
--- a/src/malli/dev.clj
+++ b/src/malli/dev.clj
@@ -8,7 +8,10 @@
   "Stops instrumentation for all functions vars and removes clj-kondo type annotations."
   []
   (remove-watch @#'m/-function-schemas* ::watch)
-  (mi/unstrument!)
+  (->> (mi/unstrument!)
+       count
+       (format "unstrumented %d vars")
+       println)
   (clj-kondo/save! {})
   (println "stopped instrumentation"))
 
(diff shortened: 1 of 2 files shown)


ingesolvoll:
This would be great! Could you also find some place where it makes sense to document this?

gpind:
@ikitommi any objections to this?

ikitommi:
Thanks for the PR! I think reducing printing noice is really needed. But just one `:debug` is bit too simple imo. For large codebases, a full re-instrument takes some time, would be nice go get stats on the time taken too + how many fns were instrumented. We could add incremental instrumentation to make it faster if perf becomes a bottleneck.

Could we have a extension mechanism here, that would allow supporting these both cases:

1. collecting and printing individual instrumentations
2. handling the whole instrumentation process

```clojure

;; custom code to print per var
... instrumented xyz, took 12ms

;; custom code to print per whole thing
instrumented 32 vars, took 400ms
```

... I think a good default would be to print just the overall stats. Less noice, but still info that the process is running + numbers if that gets slow.

WDYT?

gpind:
I'd rather make `instrument!` return information about what it instrumented, maybe just as a collection of vars or symbols? I can work on that, if you like the idea.

As for timing: the caller can easily time the call to `instrument!`, and finer-grained timing seems like a job for a profiler.

ikitommi:
Returning data from `instrument!` seems like a good idea. In `malli.dev`, maybe default to printing just one line about the stats. If you feel it's still too much noice, the instrumentation reporting function could be configurable so one van remove it or do more reporting.

gpind:
@ikitommi I've implemented what we discussed; could you take another look?

ikitommi:
Sorry for taking so long, looks good IMO. Thanks!

Issue #821, opened 2023-01-19
Malli generates :nilable/any which is not a valid type in clj-kondo


borkdude:
`:any` already includes nilables but this generates `:nilable/any`:

```
user=> (malli.clj-kondo/accept :maybe :maybe [:any] nil)
:nilable/any
```

This causes weird output with clj-kondo linting in [this issue](https://github.com/clj-kondo/clj-kondo/issues/1961).

tvaisanen:
fixed by #843

PR #843, opened 2023-02-07
Fix :any output for clj-kondo

dvingo:
Addresses https://github.com/metosin/malli/issues/821

Prevents adding `nilable` for `[:maybe :any]` schemas in clj-kondo config.


ikitommi:
awesome, thanks!

Issue #847, opened 2023-02-08
Qualified keyword ref schemas do not support composite schema forms


dvingo:
If your ref schema in the registry is a vector and your supplied schema is also a vector an exception is thrown.
If either of them is not in a vector then no error is thrown.

```clojure
(m/schema ::test {:registry (merge (m/default-schemas) {::test [:int]})})
=> ::test

(m/schema [::test] {:registry (merge (m/default-schemas) {::test [:int]})})

Execution error (ExceptionInfo) at malli.core/-exception (core.cljc:138).
:malli.core/invalid-schema
```

dvingo:
This was a misunderstanding on my part - closing.

PR #848, opened 2023-02-08
Fix support ref schemas as vector forms in a registry

dvingo:
Addresses https://github.com/metosin/malli/issues/847

Unwrap schemas in vector form for subsequent schema lookups.


ikitommi:
This doesn't look correct:

```clojure
(let [opts {:registry {:map (m/-map-schema)
                       :int (m/-int-schema)
                       ::test :map
                       ::test-2 [:map [:x :int]]}}]
  (m/schema [::test-2] opts)) ;=> :map
```

dvingo:
ahh, thanks for the feedback! I understand now and agree doing this will lead to other problems, I can use other strategies to get the behavior I want.

PR #850, opened 2023-02-08
test: verify that :decode {:string f} works like :decode/string f

### LINKED ISSUE #648: Value transformation with nested schema properties
I noticed that this way of specifying schema properties works:

```clojure
(def A
  [string?
   {:decode/string (constantly "string")
    :decode/json (constantly "json")}])

(malli.core/decode A "" malli.transform/string-transformer)
;; => "string"
(malli.core/decode A "" malli.transform/json-transformer)
;; => "json"
```

However, this alternate nested syntax mentioned in the [README](https://github.com/metosin/malli/blob/master/README.md#value-transformation) doesn't:

```clojure
(def B
  [string?
   {:decode {:string (constantly "string")
             :json (constantly "json")}}])

(malli.core/decode B "" malli.transform/string-transformer)
;; => ""
(malli.core/decode B "" malli.transform/json-transformer)
;; => ""
```

opqdonut:
closes #648

### DIFF
--- a/test/malli/transform_test.cljc
+++ b/test/malli/transform_test.cljc
@@ -635,16 +635,26 @@
   (let [schema [string? {:title "lower-upper-string"
                          :decode/string 'str/upper-case
                          :encode/string 'str/lower-case}]
+        schema2 [string? {:title "lower-upper-string"
+                          :decode {:string 'str/upper-case}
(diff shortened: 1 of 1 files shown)


ikitommi:
that was elegant :)

Issue #839, opened 2023-02-03
mx/defn broken with destructuring


opqdonut:
Two related issues: 

1. Sequence destructuring generates invalid types. The type for a `[i :- int, s :- string]` destructuring should be `[:cat :int :string [:* :any]]`, not `[:cat [:? :int] [:? :string] [:* :any]]` which admits any sequence

2. Using a `:cat` schema for an argument doesn't work. The `:cat` schema for the argument gets combined with the top-level `:cat`. This doesn't work even with a `[:schema [:cat ..]]` workaround for the argument. However `[:tuple ...]` luckily works.

Repl session below.

```
;; 1. Sequence destructuring
user=> (require '[malli.experimental :as mx])
nil
user=> (require 'malli.dev)
nil
user=> (mx/defn foo [[i :- :int, s :- :string]] {:int i :string s})
#'user/foo
user=> (malli.dev/start!)
..instrumented #'user/foo
started instrumentation
nil
user=> (foo ["x"])
{:int "x", :string nil}
user=> (foo ["x" 1])
{:int "x", :string 1}
user=> (:schema (meta #'foo))
[:=> [:cat [:maybe [:cat [:? :int] [:? :string] [:* :any]]]] :any]

;; 2. :cat arguments
user=> (mx/defn bar [things :- [:cat :int :string]] things)
..instrumented #'user/bar
#'user/bar
user=> (bar [1 "x"])
-- Schema Error ---------------------------------------------------------------- NO_SOURCE_FILE:1 --

Invalid function arity (1):

  [[1 "x"]]

Function Schema:

  [:=> [:cat [:cat :int :string]] :any]



More information:

  https://cljdoc.org/d/metosin/malli/CURRENT/doc/function-schemas

----------------------------------------------------------------------------------------------------
-- Schema Error ---------------------------------------------------------------- NO_SOURCE_FILE:1 --

Invalid function arguments:

  [[1 "x"]]

Function Var:

  user/bar

Input Schema:

  [:cat [:cat :int :string]]

Errors:

  {:in [0], :message "should be an integer", :path [0 0], :schema :int, :value [1 "x"]}

More information:

  https://cljdoc.org/d/metosin/malli/CURRENT/doc/function-schemas

----------------------------------------------------------------------------------------------------
[1 "x"]
user=> (mx/defn bar2 [things :- [:schema [:cat :int :string]]] things)
..instrumented #'user/bar2
#'user/bar2
user=> (bar2 [1 "x"])
[...]

PR #851, opened 2023-02-10
fix: -regex-min-max for :schema inside :cat

opqdonut:
Fixes part 2 of #839


opqdonut:
This is the minimal fix to make `regex-min-max` work properly on `[:cat [:schema [:cat ...]]]`.

A cleaner solution might be to use a special-purpose marker schema for this instead of the generic `:schema`. That would break backwards compatibility, unfortunately.

src/malli/core.cljc
```diff
@@ -232,6 +232,14 @@
   (let [{min'' :min max'' :max} (-regex-min-max child)]
     (cond-> {:min (f (or min' 0) min'')} (and max' max'') (assoc :max (f max' max'')))))
 
+(defn- -re-cat-min-max
```
ikitommi, line 235: anything that is not a regex schema takes on slot. e.g. `[:cat [:and [:cat :int :int]]]`.

ikitommi (changes requested):
looks good, but should be more generic.

src/malli/core.cljc
opqdonut: That's handled by the default `RegexSchema` instance `-regex-min-max` here: 

https://github.com/metosin/malli/pull/851/files#diff-c10cf1bd7f29baf493dfd1fb727c955f2bf09a256cdc7ff690343b5940f14940L125

There are also existing tests for this behaviour, which my change doesn't break.

ikitommi: failing case:

```clojure
(m/-regex-min-max
 (m/schema
  [:cat {:registry {::cat [:cat string? int?]}}
   int? ::cat]))
; => {:min 3, :max 3}
```

There are currently one `RefSchema` that is also `RegexSchema`, which is the one created with `m/-schema-schema`. It is used for two concrete Schema types, `:schema` and `::m/schema`. The latter is used to mark eager schema references. Quick fix would be to allow them both in line 238, but this is not future-proof as users can define their own reference schema-types which can't contribute to the internal `if` we have here.

I'm ok with the quick fix here (with test), but if there is a simple way to make this future-proof, I'm all ears.

opqdonut: Arguably a user might be expecting the behaviour from your "failing case". Perhaps they want to name regex fragments:

```clj
[:cat {:registry {::option-and-value [:cat keyword? [:maybe int?]]}}
 ::option-and-value
 ::option-and-value]
```
vs.
```clj
[:cat {:registry {::option-and-value [:cat keyword? [:maybe int?]]}}
 [:schema ::option-and-value]
 [:schema ::option-and-value]]
```

ikitommi: so, options:

1. add the `::m/schema` to the if and this fix is good enough for now
2. better way to escape the sequences

for 2, what do you think of the following:
* continue to use `:schema` as (the default) sequence schema escaper
* mark `::m/schema` as a regex-op (BREAKING)

with this, the code would simplify a lot:
* no need for `-re-cat-min-max` => the responsibility to handle escaping the sequence is purely on the Schema implementation
* would support the fragments you presented

I think the only change needed would be to change the following `-schema-schema` `RegexSchema` methods:

```clojure
(-regex-op? [_] internal)

(-regex-min-max [_]
  (if internal
    (-regex-min-max child)
    {:min 1, :max 1}))
```

Would be a breaking change, but smaller than dropping the `:schema` as an way to escape the sequence and for the better (allowing "fragments").

WDYT?

opqdonut:
In master:
```
user=> (m/-regex-min-max (m/schema [:schema [:cat int? int?]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [::m/schema [:cat int? int?]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [:cat int? [:schema [:cat int? int?]]]))
{:min 3, :max 3}
user=> (m/-regex-min-max (m/schema [:cat int? [::m/schema [:cat int? int?]]]))
{:min 3, :max 3}
```

In this PR:
```
user=> (m/-regex-min-max (m/schema [:schema [:cat int? int?]]))
{:min 1, :max 1}
user=> (m/-regex-min-max (m/schema [::m/schema [:cat int? int?]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [:cat int? [:schema [:cat int? int?]]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [:cat int? [::m/schema [:cat int? int?]]]))
{:min 3, :max 3}
```

opqdonut:
Just for completeness, this PR doesn't change the validation behaviour:

```clj
user=> (m/validate (m/schema [:schema [:cat int? int?]]) [1 2])
true
user=> (m/validate (m/schema [::m/schema [:cat int? int?]]) [1 2])
true
user=> (m/validate (m/schema [:cat int? [:schema [:cat int? int?]]]) [1 2 3])
false
user=> (m/validate (m/schema [:cat int? [:schema [:cat int? int?]]]) [1 [2 3]])
true
user=> (m/validate (m/schema [:cat int? [::m/schema [:cat int? int?]]]) [1 2 3])
true
user=> (m/validate (m/schema [:cat int? [::m/schema [:cat int? int?]]]) [1 [2 3]])
false
```

ikitommi (commented):
my bad for not seeing that of this returning min=1, max=1 now:

```clojure
[:schema [:cat int? int?]]
```

as the validation works with sequences of different size than `m/-regex-min-max` reports, I think this is wrong. your code of calculating the min-max differently within `:cat` and `:catn` schemas would have calculated those correctly, right? could we merge the two approaches to get the top-level counting + the fragments working?

ikitommi (approved):
approving this, we can also merge this like it is + write an issue that it reports things weirdly. Can be resolved as a separate (smaller) issue.

opqdonut:
Yeah my original fix returns `{:min 2 :max 2}` for `[:schema [:cat int? int?]]`. The point of `-re-cat-min-max` was to handle `:schema` inside `:cat` differently from `:schema` outside `:cat`. There's no way to do a "local" fix that would have `:schema` behave differently based on context.

The fragments worked with my previous fix as well:

```clj
user=> (m/-regex-min-max (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? ::cat]))
{:min 3, :max 3}
user=> (m/validate (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? ::cat]) [1 "2" 3])
true
user=> (m/-regex-min-max (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? [:schema ::cat]]))
{:min 2, :max 2}
user=> (m/validate (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? [:schema ::cat]]) [1 "2" 3])
false
user=> (m/validate (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? [:schema ::cat]]) [1 ["2" 3]])
true
user=> (m/-regex-min-max (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? [::m/schema ::cat]]))
{:min 3, :max 3}
user=> (m/validate (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? [::m/schema ::cat]]) [1 "2" 3])
true
user=> (m/validate (m/schema [:cat {:registry {::cat [:cat string? int?]}} int? [::m/schema ::cat]]) [1 ["2" 3]])
false
```

opqdonut:
So is there a reason not to merge my original fix? It seems to have the behaviour we agree on. I can add test cases to document these decisions as well.

opqdonut:
Oh now I see, with my previous fix `:schema` doesn't work under other seqex modifiers:
```
(m/-regex-min-max (m/schema [:+ [:schema [:cat int? int?]]]))
 ==> {:min 2}
```

I'll need to go back to the hammock for a while.

ikitommi:
hammock time :+1:. maybe the caller of the `RegexSchema` methods should tell the callee that "you are under a sequence schema", e.g. extra parameter into the methods. This way, the responsibility to work correctly is with the the Schemas.

I somehow think the process of composing sequence schemas should be done differently and it would remove all of these weird issues.

opqdonut:
Figured out a way to make it work.

```clj
user=> (m/-regex-min-max (m/schema [:schema [:cat int? int?]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [::m/schema [:cat int? int?]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [:cat int? [:schema [:cat int? int?]]]))
{:min 2, :max 2}
user=> (m/-regex-min-max (m/schema [:cat int? [::m/schema [:cat int? int?]]]))
{:min 3, :max 3}
```

ikitommi:
Looks good. Do we need the 2 arities? It's extender api, so we can break it just to have the arity-2 version.

opqdonut:
Good point. I'll remove the other arity.

PR #853, opened 2023-02-14
Implement malli.experimental.time schemas for clojurescript using js-joda

dvingo:
I picked up from the work started by @henryw374 as mentioned in this comment: https://github.com/metosin/malli/issues/49#issuecomment-1384374256 
to add time schemas support for clojurescript which match the JVM implementation.

These time namespaces do not affect users who do not require them and thus they also don't need js-joda installed to use malli, unless they're using these schemas of course.


deps.edn
```diff
@@ -25,7 +25,9 @@
                                com.bhauman/spell-spec {:mvn/version "0.1.2"}
                                org.clojure/spec-alpha2 {:git/url "https://github.com/clojure/spec-alpha2.git"
                                                         :sha "3d32b5e571b98e2930a7b2ed1dd9551bb269375a"}}}
-           :cljs-test-runner {:extra-deps {olical/cljs-test-runner {:mvn/version "3.8.0"}}
+           :cljs-test-runner {:extra-deps {olical/cljs-test-runner {:mvn/version "3.8.0"}
+                                           ; used only to pull in its externs file needed to compile js-joda types under advanced compilation
+                                           com.widdindustries/cljs.java-time {:mvn/version "0.1.20"}}
```
Deraen, line 30: Did you try using this namespace from a shadow-cljs project? It has better externs inference and will warn about cases where interop should use inference hints.

I might add shadow-cljs to Malli test setup just for inference checks.

src/malli/experimental/time.cljc
```diff
@@ -0,0 +1,82 @@
...
+      :pred (fn pred [x]
+              #?(:clj  (.isInstance ^Class class x)
+                 :cljs (instance? class x)))
+      :property-pred (-min-max-pred nil)}
+     type-properties
+     (assoc :type-properties type-properties))))
+
+#?(:cljs
+   (do
+     (def Period (gobj/get js-joda "Period"))
```
Deraen, line 33: `(.-Period js-joda)`

These are static properties on the JS object, so interop property access should be used.

src/malli/experimental/time.cljc
```diff
@@ -0,0 +1,82 @@
...
+     (def OffsetTime (gobj/get js-joda "OffsetTime"))
+     (def TemporalAccessor (gobj/get js-joda "TemporalAccessor"))
+     (def TemporalQuery (gobj/get js-joda "TemporalQuery"))
+     (def DateTimeFormatter (gobj/get js-joda "DateTimeFormatter"))))
+
+#?(:cljs
+   (defn createTemporalQuery [f]
+     (let [parent (TemporalQuery. "")
+           query  (js/Object.create parent)]
+       (gobj/set query "queryFrom" (fn [t] (f t)))
```
Deraen, line 58: `(set! (.-queryFrom query) ...)`

Deraen (commented):
Looks good. Just a few minor comments about the interop calls.

Having js-joda-specific implementation in namespace matching the clj side (and not, e.g., `malli.experimental.time.js-joda`) makes sense now. It is under experimental name, so we can see how the Temporal support develops and if we will need to support multiple JS implementations later or not.

dvingo:
@Deraen Thank for the review! 

There are shadow-cljs builds in the repo but they aren't currently used for testing. I used this namespace during development of malli:
https://github.com/dvingo/malli/blob/js-joda/app/malli/instrument_app.cljs
with a cljs repl.

So that was how I determined it works with shadow-cljs. I just tried under advanced compilation as well with shadow-cljs and things are working without needing externs.

src/malli/experimental/time.cljc
dvingo: ah right, thanks!

src/malli/experimental/time.cljc
dvingo: updated

deps.edn
dvingo: I think this was the only annotation needed for shadow-cljs
https://github.com/metosin/malli/pull/853/files#diff-d3e7f4bdb74b0d6cc969a8ef92f0f5dd63c43343c07e3d1347c2f636b43da117R83

ikitommi:
Looks good to me. @Deraen there is one unresolved thread, could you merge after it/this is resolved? Thanks @dvingo!

Deraen:
The question was answered already.

Issue #855, opened 2023-02-20
Support type-properties in malli.core/-map-schema and malli.core/-tuple-schema


bgalartza:
At the moment the `malli.core/-map-schema` and `malli.core/-tuple-schema`  `IntoSchema`s don't have any `type-property`, and they don't allow setting new ones programmatically.  

We would like to be able to specify the `type-properties` for the mentioned `IntoSchema`s to be able to extend their functionality by specifying type-wide custom decoders/encoders. The `mt/transformer` function does allow specifying type-specific encoders/decoders, but it's not good enough when you have a custom registry that references the built-in types. 

For example, lets say we have the following schema with a custom registry:
```clj
(def test-schema
  (m/schema
   [::object
    [:test
     [::double]]]
   {:registry {::object (m/-map-schema)
               ::double (m/-double-schema)}}))
```
At decode time we can specify that we want to use a custom decoder for all `map`s: 
```clj

(m/decode test-schema {:test 1.0}
          (mt/transformer
           {:name :test
            :decoders {:map {:enter #(prn %)}}}))
```
But we would like to specify the decoder for all `object`s, instead of `map`s.

@ikitommi mentioned in [this Clojurians thread](https://app.slack.com/client/T03RZGPFR/CLDK6MFMK/thread/CLDK6MFMK-1676640324.276269) that if `m/-map-schema` would allow custom `type-properties` we could do the following:

```clj
(def test-schema
  (m/schema
   [::object
    [:test
     [::double]]]
   {:registry {::object (m/-map-schema {:type-properties {:decode/test {:enter #(prn %)}}})
               ::double (m/-double-schema)}}))

(m/decode
 test-schema
 {:test 1.0}
 (mt/transformer {:name :test}))
 ```
Following his example we tested the approach with `malli.core/-collection-schema` (which does allow setting the `type-properties`) and it perfectly fits our use case.  We think that others might also benefit of the change, as it would allow extending the types with more functionality without having to re-implement them completely.

ikitommi:
This would be a non-breaking change. PR most welcome!

bgalartza:
We have been having a look at the code and we have found that `-type-properties` method implementation is different in the `-simple-schema` case and `-collection-schema` case. We initially thought that using the same implementation as `-simple-schema` would be enough, but we are no longer sure because we don't fully understand why the `-collection-schema` implementation uses an `atom` for the `props`. And we are not 100% sure if we should implement it that way (using an `atom` for the `props`).

[...]

PR #856, opened 2023-02-21
Support type-properties in -map-schema, -map-of-schema and -tuple-schema

bgalartza:
Pull request for the issue #855


ikitommi:
Looks good, thanks!

PR #858, opened 2023-02-21
Simple examples of := and :enum

### LINKED ISSUE #857: Explain := and :enum
I'm not exactly sure where in the README I want these, but `:=` and `:enum` seem like they should feature pretty early.
Should I take a crack at it?

brettrowberry:
Resolves #857


ikitommi:
Thanks!

Issue #464, opened 2021-06-16
Missing "Defs" in swagger (malli.json-schema) with recursive schemes 


evilsneer:
Hi! I try to use malli as coercion&schemas with swagger ui. If i try serialise recursive schemas with registries, i get some "Could not resolve reference" of keys of recursive values. it is according to plan and swagger bad, or its just undone feature?) looks like there is some missing defs in swagger.json
![image](https://user-images.githubusercontent.com/9718734/122173728-bdcbf080-ce8a-11eb-9914-ed2bb9d9a990.png)

I use swagger-ui bundled with

`[metosin/reitit-swagger-ui "0.5.13"]
[metosin/ring-swagger-ui "3.36.0"]`

guess it 3.XX.

I'v got example of that malli:

```
[:schema
 {:registry {:namespace1/logic-pred [:or
                                        [:= #:json-schema{:val "and"} "and"]
                                        [:= #:json-schema{:val "or"} "or"]],
             :namespace1/not-pred [:= #:json-schema{:val "not"} "not"],
             :namespace1/fields-or-expression [:or
                                                [:ref "Expression"]
                                                [:or
                                                 [:tuple
                                                  [:= #:json-schema{:val "="} "="]
                                                  [:= #:json-schema{:example "phone"} "phone"]
                                                  string?]
                                                 [:tuple
                                                  [:= #:json-schema{:val "contains"} "contains"]
                                                  [:= #:json-schema{:example "phone"} "phone"]
                                                  string?]
                                                 [:tuple
                                                  [:= #:json-schema{:val "startswith"} "startswith"]
                                                  [:= #:json-schema{:example "phone"} "phone"]
                                                  string?]
                                                 [:tuple
                                                  [:= #:json-schema{:val "like"} "like"]
                                                  [:= #:json-schema{:example "phone"} "phone"]
                                                  string?]]
                                                
                                                ;; ... (more :or's)
                                                
                                                                  ],
             "Expression" [:cat
                           :namespace1/logic-pred
                           [:+
                            [:or
                             [:ref :namespace1/fields-or-expression]
                             [:tuple
                              :namespace1/not-pred
                              [:ref :namespace1/fields-or-expression]]]]]}}
 [:map
  [:where {:optional true, :json-schema/val "[and [= a 2]]"} "Expression"]
  [:limit [int? {:default 100}]]
  [:offset [int? {:default 0}]]]]
```

If i malli.json-schema/transform  this (it is used in swagger gen, right?) i get something like:

```
[...]

PR #863, opened 2023-03-06
Fix malli swagger defs

cap10morgan:
These are changes I needed to make in here to fix https://github.com/metosin/reitit/issues/558 (I will be issuing a PR for the changes needed in reitit soon too).

### DIFF
--- a/src/malli/json_schema.cljc
+++ b/src/malli/json_schema.cljc
@@ -7,7 +7,12 @@
 (defprotocol JsonSchema
   (-accept [this children options] "transforms schema to JSON Schema"))
 
-(defn -ref [x] {:$ref (str "#/definitions/" x)})
+(defn -ref [x] {:$ref (apply str "#/definitions/"
+                             (cond
+                               (qualified-keyword? x) [(namespace x) "~1"
+                                                       (name x)]
+                               (keyword? x) [(name x)]
+                               :else [x]))})
 
 (defn -schema [schema {::keys [transform definitions] :as options}]
   (let [result (transform (m/deref schema) options)]
--- a/src/malli/json_schema.cljc
+++ b/src/malli/json_schema.cljc
@@ -17,7 +17,11 @@
 (defn -schema [schema {::keys [transform definitions] :as options}]
   (let [result (transform (m/deref schema) options)]
     (if-let [ref (m/-ref schema)]
-      (do (swap! definitions assoc ref result) (-ref ref))
+      (do
+        (when-not (and (contains? result :$ref)
+                       (= (:$ref result) ref)) ; don't create circular defs
+          (swap! definitions assoc ref result))
+        (-ref ref))
       result)))
 
 (defn select [m] (select-keys m [:title :description :default]))
--- a/src/malli/swagger.cljc
+++ b/src/malli/swagger.cljc
@@ -1,5 +1,6 @@
 (ns malli.swagger
   (:require [clojure.set :as set]
+            [clojure.walk :as walk]
             [malli.core :as m]
             [malli.json-schema :as json-schema]))
 
@@ -67,3 +68,102 @@
                                  ::json-schema/definitions definitions
                                  ::json-schema/transform -transform})]
      (cond-> (-transform ?schema options) (seq @definitions) (assoc :definitions @definitions)))))
+
+(defn remove-empty-keys
+  [m]
+  (into (empty m) (filter (comp not nil? val) m)))
+
+(defmulti extract-parameter (fn [in _] in))
+
+(defmethod extract-parameter :body [_ schema]
+  (let [swagger-schema (transform schema {:in :body, :type :parameter})]
+    [{:in          "body"
+      :name        (:title swagger-schema "body")
+      :description (:description swagger-schema "")
+      :required    (not= :maybe (m/type schema))
+      :schema      swagger-schema}]))
+
+(defmethod extract-parameter :default [in schema]
+  (let [{:keys [properties required definitions]} (transform schema {:in in, :type :parameter})]
+    (println "\nextract-parameter definitions:"
+             (with-out-str (clojure.pprint/pprint definitions)))
+    (mapv
+      (fn [[k {:keys [type] :as schema}]]
+        (merge
+          {:in          (name in)
+           :name        k
+           :description (:description schema "")
+           :type        type
+           :required    (contains? (set required) k)}
+          schema))
+      properties)))
+
+(defmulti expand (fn [k _ _ _] k))
+
+#_(defn lift-response-definitions
+    [response]
+    (if-let [definitions (get-in response [:schema :definitions])]
+      (-> response
+          (update :schema dissoc :definitions)
+          (assoc :definitions definitions))))
+
+(defmethod expand ::responses [_ v acc _]
+  {:responses
+   (into
+     (or (:responses acc) {})
+     (for [[status response] v]
+       [status (-> response
+                   (update :schema transform {:type :schema})
+                   (update :description (fnil identity ""))
+                   remove-empty-keys)]))})
+
+(defmethod expand ::parameters [_ v acc _]
+  (let [old    (or (:parameters acc) [])
+        new    (mapcat (fn [[in spec]] (extract-parameter in spec)) v)
+        merged (->> (into old new)
+                    reverse
+                    (reduce
+                      (fn [[ps cache :as acc] p]
+                        (let [c (select-keys p [:in :name])]
+                          (if (cache c)
+                            acc
+                            [(conj ps p) (conj cache c)])))
+                      [[] #{}])
+                    first
+                    reverse
+                    vec)]
+    {:parameters merged}))
+
+(defn expand-qualified-keywords
+  [x options]
+  (let [accept? (-> expand methods keys set)]
+    (walk/postwalk
+      (fn [x]
+        (if (map? x)
+          (reduce-kv
+            (fn [acc k v]
+              (if (accept? k)
+                (let [expanded    (expand k v acc options)
+                      parameters  (:parameters expanded)
+                      _           (println "\nPARAMETERS:"
+                                           (with-out-str (clojure.pprint/pprint parameters)))
+                      responses   (:responses expanded)
+                      _           (println "\nRESPONSES:"
+                                           (with-out-str (clojure.pprint/pprint responses)))
+                      definitions (if parameters
+                                    (-> parameters first :schema :definitions)
+                                    (->> responses vals (map :schema)
+                                         (map :definitions) (apply merge)))
+                      _           (println "\nDEFINITIONS:"
+                                           (with-out-str (clojure.pprint/pprint definitions)))]
+                  (-> acc (dissoc k) (merge expanded) (update :definitions merge definitions)))
+                acc))
+            x x)
+          x))
+      x)))
+
+(defn swagger-spec
+  ([x]
+   (swagger-spec x nil))
+  ([x options]
+   (expand-qualified-keywords x options)))


cap10morgan:
Note that if you use any recursive schemas, you'll still see some missing defs in your swagger due to #464.

src/malli/json_schema.cljc
```diff
@@ -7,12 +7,21 @@
 (defprotocol JsonSchema
   (-accept [this children options] "transforms schema to JSON Schema"))
 
-(defn -ref [x] {:$ref (str "#/definitions/" x)})
+(defn -ref [x] {:$ref (apply str "#/definitions/"
+                             (cond
+                               (qualified-keyword? x) [(namespace x) "~1"
+                                                       (name x)]
```
opqdonut, line 13: your  commit message on the `~1` encoding was good. perhaps a simple code comment here about that?

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+    [{:in          "body"
+      :name        (:title swagger-schema "body")
+      :description (:description swagger-schema "")
+      :required    (not= :maybe (m/type schema))
+      :schema      swagger-schema}]))
+
+(defmethod extract-parameter :default [in schema]
+  (let [{:keys [properties required definitions]} (transform schema {:in in, :type :parameter})]
+    (println "\nextract-parameter definitions:"
+             (with-out-str (clojure.pprint/pprint definitions)))
```
opqdonut, line 89: these printlns need to go before mergint this

src/malli/json_schema.cljc
```diff
@@ -7,12 +7,21 @@
...
 (defn -schema [schema {::keys [transform definitions] :as options}]
   (let [result (transform (m/deref schema) options)]
     (if-let [ref (m/-ref schema)]
-      (do (swap! definitions assoc ref result) (-ref ref))
+      (do
+        (when-not (and (contains? result :$ref)
+                       (= (:$ref result) ref)) ; don't create circular defs
+          (swap! definitions assoc ref result))
+        (-ref ref))
       result)))
```
opqdonut, line 25: I'd like to see a test for this new behaviour.

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+                          (if (cache c)
+                            acc
+                            [(conj ps p) (conj cache c)])))
+                      [[] #{}])
+                    first
+                    reverse
+                    vec)]
+    {:parameters merged}))
+
+(defn expand-qualified-keywords
```
opqdonut, line 137: the name here is a bit misleading: it's more like `generate-swagger-spec`, right?

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+  [x options]
+  (let [accept? (-> expand methods keys set)]
+    (walk/postwalk
+      (fn [x]
+        (if (map? x)
+          (reduce-kv
+            (fn [acc k v]
+              (if (accept? k)
```
opqdonut, line 145: you don't need the `accept?` trick if you add a default for `expand`

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+                                    (-> parameters first :schema :definitions)
+                                    (->> responses vals (map :schema)
+                                         (map :definitions) (apply merge)))
+                      _           (println "\nDEFINITIONS:"
+                                           (with-out-str (clojure.pprint/pprint definitions)))]
+                  (-> acc (dissoc k) (merge expanded) (update :definitions merge definitions)))
+                acc))
+            x x)
+          x))
+      x)))
```
opqdonut, line 163: Reusing the name `x` makes this really hard to read. Also consider flipping the conditional like this: `(if-not (map? x) x (reduce ...))`

src/malli/swagger.cljc
cap10morgan: This is what it's called in the equivalent spec and schema code. But I'm happy to change it if you don't mind the inconsistency.

src/malli/swagger.cljc
cap10morgan: Sorry about that (and the commented-out fn above). Meant to clean those up before issuing the PR.

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+                      [[] #{}])
+                    first
+                    reverse
+                    vec)]
+    {:parameters merged}))
+
+(defn expand-qualified-keywords
+  [x options]
+  (let [accept? (-> expand methods keys set)]
+    (walk/postwalk
```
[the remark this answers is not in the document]
cap10morgan: It's copying the behavior of the equivalent code for spec and schema. So perhaps it doesn't here, but it seemed like it would need to be the same.

src/malli/swagger.cljc
cap10morgan: This is another copied-from-the-spec-and-schema code. Happy to diverge from them if you want, though.

src/malli/swagger.cljc
cap10morgan: Another spec and schema copy-pasta. But I agree on both counts. :)

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
                                  ::json-schema/definitions definitions
                                  ::json-schema/transform -transform})]
      (cond-> (-transform ?schema options) (seq @definitions) (assoc :definitions @definitions)))))
+
+(defn remove-empty-keys
+  [m]
+  (into (empty m) (filter (comp not nil? val) m)))
+
+(defmulti extract-parameter (fn [in _] in))
```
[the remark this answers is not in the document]
cap10morgan: Also mirrors the spec and schema code naming currently

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+        (merge
+          {:in          (name in)
+           :name        k
+           :description (:description schema "")
+           :type        type
+           :required    (contains? (set required) k)}
+          schema))
+      properties)))
+
+(defmulti expand (fn [k _ _ _] k))
```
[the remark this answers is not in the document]
cap10morgan: Also mirrors the spec and schema code naming currently

src/malli/json_schema.cljc
cap10morgan: Hmm... I'm not sure what causes it. I just noticed that both the actual definition and a circular ref were being generated in here from my schema.

I'll see if I can figure out a minimal repro for a test.

src/malli/swagger.cljc
```diff
@@ -67,3 +68,102 @@
...
+                acc))
+            x x)
+          x))
+      x)))
+
+(defn swagger-spec
+  ([x]
+   (swagger-spec x nil))
+  ([x options]
+   (expand-qualified-keywords x options)))
```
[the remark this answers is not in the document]
cap10morgan: I'll see what I can put together from existing spec / schema swagger-spec tests and whatever was in reitit-malli for its roughly-equivalent code.

src/malli/swagger.cljc
opqdonut: Oh okay, I'll need to look at the spec-tools and schema-tools versions and see if they make any sense.

src/malli/json_schema.cljc
opqdonut: I just bumped into this:
```
user=> (malli.json-schema/transform (mu/closed-schema [:schema {:registry {"Foo" :int}} "Foo"]))
{:$ref "#/definitions/Foo", :definitions {"Foo" {:$ref "#/definitions/Foo"}}}
user=> (malli.json-schema/transform [:schema {:registry {"Foo" :int}} "Foo"])
{:$ref "#/definitions/Foo", :definitions {"Foo" {:type "integer"}}}
```
Your PR seems to fix it. I created issue #868 for this bug.

opqdonut:
Yeah, just waiting for tests.

See [my comment](https://github.com/metosin/malli/pull/863#discussion_r1133479118) for one possible test case.

src/malli/json_schema.cljc
cap10morgan: Turned that into a test in 64b15cb40a73ca40b22a9647c85ddc601e8b91e0.

cap10morgan:
OK @opqdonut I think this is ready for re-review. I even added a (commented-out) failing test for #464.

test/malli/swagger_test.cljc
```diff
@@ -276,3 +276,142 @@
...
+      (is (= {:definitions {::body {:minLength 1, :type "string"}},
+              :parameters [{:description "",
+                            :in "body",
+                            :name "body",
+                            :required true,
+                            :schema {:$ref "#/definitions/malli.swagger-test~1body",
+                                     :definitions {::body {:minLength 1, :type "string"}}}}]}
+             (swagger/swagger-spec {::swagger/parameters
+                                    {:body (m/schema ::body
+                                                     {:registry registry})}})))))
```
opqdonut, line 293: Thanks for these! Could you also have a simple non-registry test case? Something like
```
{:body [:map [:foo :string]]}
```
?

test/malli/swagger_test.cljc
```diff
@@ -276,3 +276,142 @@
...
+  (testing "generates swagger for ::parameters w/ basic schema"
+    (let [registry (merge (m/type-schemas)
+                          {::body [:string {:min 1}]})]
+      (is (= {:definitions {::body {:minLength 1, :type "string"}},
+              :parameters [{:description "",
+                            :in "body",
+                            :name "body",
+                            :required true,
+                            :schema {:$ref "#/definitions/malli.swagger-test~1body",
+                                     :definitions {::body {:minLength 1, :type "string"}}}}]}
```
opqdonut, line 290: should it be `:definitions {"malli.swagger-test~1body" ...}`?

opqdonut (changes requested):

test/malli/swagger_test.cljc
cap10morgan: That's a good question. I think it doesn't need to be because by the time it gets serialized to JSON it gets converted to that. The one aspect I'm a little surprised by is that the `/` -> `~1` conversion is happening. But maybe that's just because it's being converted by JSON-Schema-aware code? Probably worth a little more investigation.

test/malli/swagger_test.cljc
cap10morgan: Added in [2fedffc](https://github.com/metosin/malli/pull/863/commits/2fedffc14e30c7ee90048ed31a472778089c3524)

test/malli/swagger_test.cljc
opqdonut: Jsonista and other clojure json libraries just convert `:foo.bar/baz` to `"foo.bar/baz"`. Grepping for `~` doesn't give me any hits from the malli or reitit repositories.

cap10morgan: Oh! I remember now. We don't want those to have `/` changed to `~1` because they aren't pointers, they're just regular old JSON object keys. You have to encode `/` as `~1` _only_ in JSON pointers (a.k.a. the values of `$ref` keys). [See here](https://opis.io/json-schema/2.x/pointers.html).

So yeah, the qualified keywords get expanded into the appropriate string when serialized to JSON.

opqdonut: Oh right. That makes sense. Could you add that link as a comment to  `-ref`? I think this is ready merge after that!

opqdonut: Actually, I'll just do that for you so we can avoid the round-trip.

opqdonut (approved):
I'll add the links I requested

src/malli/swagger.cljc
```diff
@@ -67,3 +72,85 @@
...
+  (let [swagger-schema (transform schema {:in :body, :type :parameter})]
+    [{:in          "body"
+      :name        (:title swagger-schema "body")
+      :description (:description swagger-schema "")
+      :required    (not= :maybe (m/type schema))
+      :schema      swagger-schema}]))
+
+(defmethod extract-parameter :default [in schema]
+  (let [{:keys [properties required definitions]} (transform schema {:in in, :type :parameter})]
+    (println "\nextract-parameter definitions:"
```
ikitommi, line 88: this broke reitit build, fixed.

ikitommi: https://github.com/metosin/malli/pull/878/files?w=1

opqdonut: Oof sorry for not noticing that last println!

ikitommi: no prob, the build broke with clojure 1.10 because of not requiring the `clojure.pprint`.

Issue #860, opened 2023-02-22
Validation of map schemas with keys missing types


brettrowberry:
I would expect this to throw/return some error about this being an invalid map schema:

```clojure
(malli.core/schema [:map [:foo]])
;; => :map 
```

Is this by design?

ikitommi:
This is not intentional, should fail.

vharmain:
Closed by https://github.com/metosin/malli/pull/864

brettrowberry:
Thank you!

PR #864, opened 2023-03-09
Fail when map schema is missing a type for a key

tvaisanen:
Fixes issue #860

Invalid entries should not be accepted.

For example the following should fail.

```clojure
(malli.core/schema [:map [:foo]])
;; => :map
```

The issue was that `malli.core/-eager-entry-parser`'s loop iterator was not incremented properly when given entry such as `[:foo]`. This was caused by `malli.core/-parse-entry` returning the same iterator `i` value that it was given.

Fail instead of "skipping" the entry.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -440,7 +440,9 @@
             n (alength ea)
             e0 (aget ea 0)]
         (if (== n 1)
-          (if (and (-reference? e0) naked-keys) (-parse-ref-vector1 e e0) i)
+          (if (and (-reference? e0) naked-keys)
+            (-parse-ref-vector1 e e0)
+            (-fail! ::invalid-children {:children -children}))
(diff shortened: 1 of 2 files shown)


src/malli/core.cljc
```diff
@@ -440,7 +442,7 @@
             n (alength ea)
             e0 (aget ea 0)]
         (if (== n 1)
-          (if (and (-reference? e0) naked-keys) (-parse-ref-vector1 e e0) i)
+          (if (and (-reference? e0) naked-keys) (-parse-ref-vector1 e e0) (unchecked-inc-int i))
```
ikitommi, line 445: instead of `(unchecked-inc-int i)` (e.g. "skip"), could we just fail there?

ikitommi (approved):
Looks great!

PR #870, opened 2023-03-14
Enable (leaf-)Schema caching

ikitommi:
Leaf-schemas are usually created without properties or children. This MR introduces `IntoSchema` level caching. `Schema` instance cache is enabled if all of the following apply:
* `IntoSchema` instance 
* which also support `Cached`
* is instantiated without children, properties or options

Currently, this is enabled just for `m/-simple-schemas`, yielding 15% better performance on schema creation + smaller memory footprint (not tested)

### DIFF
--- a/perf/malli/perf/core.cljc
+++ b/perf/malli/perf/core.cljc
@@ -13,10 +13,10 @@
   `(cc/quick-bench ~@body))
 
 (defmacro profile [& body]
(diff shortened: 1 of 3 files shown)


ikitommi:
marked as `invalid` as not sure if this makes sense. More code, small benefits.

PR #882, opened 2023-03-24
Fix: ignore schemaless responses

cap10morgan:
Adds a test for and fixes the issue causing the CI failures on [this branch of reitit](https://github.com/metosin/reitit/pull/589).

### DIFF
--- a/src/malli/swagger.cljc
+++ b/src/malli/swagger.cljc
@@ -102,11 +102,11 @@
   {:responses
    (into
     (or (:responses acc) {})
-    (for [[status response] (filter #(-> % second :schema) v)]
-      [status (-> response
-                  (update :schema transform {:type :schema})
-                  (update :description (fnil identity ""))
-                  -remove-empty-keys)]))})
+    (for [[status response] v]
+      [status (cond-> response
+                      (:schema response) (update :schema transform {:type :schema})
+                      true (update :description (fnil identity ""))
+                      true -remove-empty-keys)]))})
 
 (defmethod expand ::parameters [_ v acc _]
   (let [old (or (:parameters acc) [])
--- a/src/malli/swagger.cljc
+++ b/src/malli/swagger.cljc
@@ -101,12 +101,12 @@
 (defmethod expand ::responses [_ v acc _]
   {:responses
    (into
-    (or (:responses acc) {})
-    (for [[status response] v]
-      [status (-> response
-                  (update :schema transform {:type :schema})
-                  (update :description (fnil identity ""))
-                  -remove-empty-keys)]))})
+     (or (:responses acc) {})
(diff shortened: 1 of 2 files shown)


src/malli/swagger.cljc
```diff
@@ -101,12 +101,12 @@
...
-      [status (-> response
-                  (update :schema transform {:type :schema})
-                  (update :description (fnil identity ""))
-                  -remove-empty-keys)]))})
+     (or (:responses acc) {})
+     (for [[status response] (filter #(-> % second :schema) v)]
+       [status (-> response
+                   (update :schema transform {:type :schema})
+                   (update :description (fnil identity ""))
+                   -remove-empty-keys)]))})
```
opqdonut, line 109: Please don't reindent this. We use standard clojure indentation with function arguments indented one space. Tommi just fixed this in 44a9ee07.

src/malli/swagger.cljc
```diff
@@ -101,12 +101,12 @@
...
-    (or (:responses acc) {})
-    (for [[status response] v]
-      [status (-> response
-                  (update :schema transform {:type :schema})
-                  (update :description (fnil identity ""))
-                  -remove-empty-keys)]))})
+     (or (:responses acc) {})
+     (for [[status response] (filter #(-> % second :schema) v)]
+       [status (-> response
+                   (update :schema transform {:type :schema})
```
opqdonut, line 107: Instead of filtering out the responses without a schema, could you just skip the transform step for a nil schema? That way the other keys like `:description` would still end up in the swagger spec.

test/malli/swagger_test.cljc
```diff
@@ -362,6 +362,15 @@
                                      400 {:schema (m/schema ::error-resp
                                                             {:registry registry})}}})))))
 
+  (testing "no schema in responses ignored"
+    (is (= {:definitions nil
+            :responses {200 {:description "" :schema {:type "string"}}}}
+           (swagger/swagger-spec {#_::swagger/parameters
+                                  #_{:body {:description "just a body"}}
+                                  ::swagger/responses
+                                  {500 {:description "fail"}
```
opqdonut, line 371: I think the result should have a `500 {:description "fail"}`. Cf. https://github.com/metosin/reitit/blob/master/test/cljc/reitit/swagger_test.clj#L173

opqdonut (changes requested):

src/malli/swagger.cljc
cap10morgan: Sorry my editor was misconfigured. I'll revert it.

cap10morgan:
@opqdonut OK I think this is ready for re-review. Let me know if there are any more issues to address.

opqdonut (approved):
thanks! sorry for the late review, I was off on easter vacation

PR #886, opened 2023-03-29
Fix typo

simonacca:
Unless I'm missing something, present tense should be used here as it's currently true that you have to bring in your state atom in order to have a global registry


ikitommi:
thanks!

PR #887, opened 2023-03-30
Fix clj-doc API import

simonacca:
Hi,
cljdoc api generation currently [fails](https://app.circleci.com/pipelines/github/cljdoc/builder/40718/workflows/34426e51-e450-4736-8dca-dd730fe903d7/jobs/57093) because it can't find shadow-cljs as a dependency must be in [pom.xml](https://github.com/cljdoc/cljdoc/blob/master/doc/userguide/for-library-authors.adoc#getting-dependencies-right).
However that dependency is only needed when analyzing `malli.dev.*` namespaces, which I don't think we want in cljdocs anyway, so I'm just [excluding](https://github.com/cljdoc/cljdoc/blob/master/doc/userguide/for-library-authors.adoc#declaring-your-public-api) them from cljdoc analysis in this PR.
Let me know if this approach works for you.

Simon

### DIFF
--- a/src/malli/dev/cljs_kondo_preload.cljc
+++ b/src/malli/dev/cljs_kondo_preload.cljc
@@ -1,4 +1,4 @@
-(ns malli.dev.cljs-kondo-preload
+(ns ^:no-doc malli.dev.cljs-kondo-preload
   "Shadow-cljs preload for browser builds, used to persist clj-kondo config collected from function schemas to disk during development."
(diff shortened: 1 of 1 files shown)


ikitommi:
`malli.dev.pretty` and `malli.dev.virhe` are something that people can use, so would like to have it in the docs. Maybe just exclude the one that doesn't work?

PR #889, opened 2023-04-05
Simplify uuid regex for accept non-standard and zero uuids

niwinz:
The previous regular expression only accepted a very small subset of standard versions (a commonly used uuid v1-v6). This commit replaces it with a more general/simplified regex that also accepts uuid-zero, uuid-max and all other non-standardarized but still valid uuid values (including all the upcoming new versions https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis)


piotr-yuxuan:
Glad to read that!

ikitommi:
Thanks!!

PR #890, opened 2023-04-09
Replace goog/mixin with Object.assign

dvingo:
The latest version of shadow-cljs uses a newer version of the closure compiler and library which removed `goog.mixin` here:
https://github.com/google/closure-library/commit/d9aace7423bcee7fc700e6f68db23e6ba9670f25

resulting in this error during cljs instrumentation:
goog_mixin_err

This PR replaces `goog/mixin` use with `js/Object.assign`.


ikitommi:
Thanks!

Issue #775, opened 2022-10-27
`[:set [:enum …]]` results in error in some case for reitit + malli request coercion


piotr-yuxuan:
In some code I have a schema that is as close as possible to this one (edited for confidentiality):
``` clojure
(def MySchema
  [:map [:my-field {:default #{:created-at :id}}
         [:set {:min 1 :decode {:string #(if (string? %) (str/split % #",") %)}}
          [:enum {:decode {:string keyword
                           :json keyword}}
           :created-at :id]]]])
```

I use reitit 0.5.18 with malli coercion. When a request body contains a slightly invalid value, like:
``` clojure
{:my-field #{:credated-at}}
```

then I can catch an explanation value passed to `humanize` that essentially boils down to:

``` clojure
(def explanation
  {:value {:my-field ["id" "credated-at"]} ; Come from an http request, decoded with malli.
   :errors (list {:path [:my-field 0]
                  :in [:my-field :credated-at]
                  :schema [:enum :created-at :id]
                  :value :credated-at})})
```

The issue is that it somehow breaks `humanize`:

``` clojure
(malli.error/humanize explanation)
; => (IllegalArgumentException. "Key must be integer")
```

However, I'm very puzzled, because I can't reproduce the bug above:

``` clojure
(me/humanize
  (m/explain
    MySchema
    {:my-field #{:credated-at}}))
=> {:my-field #{["should be either :created-at or :id"]}}
```

I tried to suggest a fix in https://github.com/metosin/malli/pull/774 but obviously more work, and a better understanding, are needed. I tried to make this bug report as accurate as possible but I know it doesn't pinpoint the root cause – I haven't been able to find it.

piotr-yuxuan:
In my use case, passing down to `malli.error/humanize` the decoded value instead of the input value removes the exception. Not sure if it is a good way to fix it though.

PR #894, opened 2023-04-12
Fix issue #775

cap10morgan:
I also ran into issue #775 and saw that there was a draft PR (#774) to fix it but wasn't sure what the status of that was. A similar approach seemed to fix it, so here's a non-draft PR. Thanks @piotr-yuxuan for the shoulders to stand on!

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -146,9 +146,12 @@
 (defn -fill [x i fill] (-concat x (repeat (- i (count x)) fill)))
 
 (defn -push [x k v fill]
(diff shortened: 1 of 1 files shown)


cap10morgan:
Actually it looks like this doesn't solve the problem for me. Maybe that's why #774 was in draft. :)

I'm closing in on a test case that demonstrates the problem at a higher level, so hopefully I can figure out a real fix from that.

No need to review this yet.

cap10morgan:
I wasn't able to repro this in a simple test after all. Closing for now.

PR #895, opened 2023-04-13
zone-id gen - do not use registry to lookup enum

dvingo:
If you're using a custom default registry and require the `experimental.time.generator` namespace `:enum` will not be available yet when the `experimental.time.generator` namespace is compiled. 

This PR switches to getting the `:enum` schema directly from `malli.core`.

Here is an example error in cljs with a custom default registry:

Screenshot_20230412_233325

### DIFF
--- a/src/malli/experimental/time/generator.cljc
+++ b/src/malli/experimental/time/generator.cljc
@@ -11,7 +11,7 @@
 #?(:clj (set! *warn-on-reflection* true))
 
 (def zone-id-gen
-  (mg/generator (into [:enum] (map #(. ZoneId of ^String %)) (. ZoneId getAvailableZoneIds))))
+  (mg/generator (m/into-schema (m/-enum-schema) nil (map #(. ZoneId of ^String %) (. ZoneId getAvailableZoneIds)))))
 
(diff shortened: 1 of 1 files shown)


src/malli/experimental/time/generator.cljc
```diff
@@ -11,7 +11,7 @@
 #?(:clj (set! *warn-on-reflection* true))
 
 (def zone-id-gen
-  (mg/generator (into [:enum] (map #(. ZoneId of ^String %)) (. ZoneId getAvailableZoneIds))))
+  (mg/generator (into [(:enum (m/base-schemas))] (map #(. ZoneId of ^String %)) (. ZoneId getAvailableZoneIds))))
```
ikitommi, line 14: `m/base-schemas` pull a lot of schemas in, you can use instead:

```clojure
(into [(m/-enum-schema)] ..children...
```

or directly:

```clojure
(m/int-schema (m/-enum-schema) nil ...children...
```

the latter being faster.

dvingo: very cool! thanks, today I learned...

ikitommi:
Thanks!!

PR #898, opened 2023-04-18
Do not require timezone data directly for cljs

dvingo:
When making a release cljs build I noticed js-joda was around 700kb and then realized the `malli.experimental.time` namespace was pulling in all historical timezone data.
See the note at the bottom of the package for more info: https://www.npmjs.com/package/@js-joda/timezone
using the +/- 5 year range adds ~30kb.

This change requires ClojureScript users of malli to include the timezone data of their choice explicitly.

I had to convert the  `zone-id-gen` to a function due to `ZoneId` now not being available until runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3031,6 +3031,21 @@ To use these schemas in ClojureScript you will need to install the npm packages 
 npm install @js-joda/core @js-joda/timezone
 ```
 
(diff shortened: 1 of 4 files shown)


ikitommi:
This makes sense. Thanks!!

PR #900, opened 2023-05-01
Remove non-root swagger definitions

cap10morgan:
As per discussion in https://github.com/metosin/reitit/pull/589#discussion_r1174793547 this removes the non-root `:definitions` from the swagger output.

### DIFF
--- a/src/malli/swagger.cljc
+++ b/src/malli/swagger.cljc
@@ -125,6 +125,22 @@
                     vec)]
     {:parameters merged}))
 
+(defn dissoc-non-root-definitions
+  [{:keys [parameters responses] :as x}]
+  (cond-> x
+          parameters (update :parameters
+                             #(mapv (fn [p]
+                                      (update p :schema dissoc :definitions))
+                                    %))
+          responses (update :responses
+                            #(reduce-kv (fn [rs k v]
+                                          (assoc rs k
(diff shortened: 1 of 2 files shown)


cap10morgan:
I removed all of the trailing commas in the swagger-test data in https://github.com/metosin/malli/pull/900/commits/26bda9b12bbc14b00f5006aab59fb52423ffba3c, but let me know if you'd rather I not do that. I was having to remove the ones from the deleted non-root definitions in the test expectations and decided it might be better to just clean it all up in one commit first.

cap10morgan:
The one failing test also failed before my changes.

opqdonut:
Thanks for removing the commas in a separate commit!

All tests seem to pass now. The failures were caused by a clojars outage, but now all the deps downloaded fine and the tests passed as well.

I have a nagging thought that the `:definitions` inside `:schema` might be worth leaving in. I'll investigate a bit and see what various validators say.

opqdonut:
Ok, we only need the top-level `:definitions` since all the references are of the form `"$ref": "#/definitions/...`. I think this is good to go.

opqdonut (approved):
I'll merge this later today

cap10morgan:
OK, and now with this merged reitit-malli's test expectations need to be updated, at least. I'll work up a PR for that later today.

PR #901, opened 2023-05-10
Add a note to the README about humanizing exceptions

cap10morgan:
This has proven very useful in our codebase so thought I'd share.


ikitommi:
Thanks for sharing this!

PR #911, opened 2023-06-19
Fix typo in README.md

FrankApiyo:
## Small change in PR
- [x] Fix typo in readme: `arbitraty` -> `arbitrarily`

PR #917, opened 2023-07-09
Update README.md (minor typo)

tlight:
fixed a typo


tlight:
close since same as #911.

PR #919, opened 2023-07-11
Remove duplicated prefix

RokLenarcic:
Fix for a typo


ikitommi:
It's a stack of namespaces, but it seems that the second `malli` is not needed anymore. Thanks!

PR #927, opened 2023-08-09
add :default/fn prop for default-value-transformer

handerpeder:
use case:
we want the default value to be current time.

previously you would have to do something like this:

```clojure
(m/decode
 [:time/instant {:default :now}]
 nil
 (mt/default-value-transformer {:default-fn
                                (fn [_ x ] (case x
                                             :now (java.time.Instant/now)
                                             x))}))
```

this patch lets you declare this without the indirection:

```clojure
(m/decode [:time/instant {:default/fn #(java.time.Instant/now)}]
          nil
          mt/default-value-transformer)
```

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -397,9 +397,10 @@
   ([] (default-value-transformer nil))
   ([{:keys [key default-fn defaults ::add-optional-keys] :or {key :default, default-fn (fn [_ x] x)}}]
    (let [get-default (fn [schema]
-                       (if-some [e (some-> schema m/properties (find key))]
-                         (constantly (val e))
-                         (some->> schema m/type (get defaults) (#(constantly (% schema))))))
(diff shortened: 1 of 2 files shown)


src/malli/transform.cljc
```diff
@@ -397,9 +397,10 @@
   ([] (default-value-transformer nil))
   ([{:keys [key default-fn defaults ::add-optional-keys] :or {key :default, default-fn (fn [_ x] x)}}]
    (let [get-default (fn [schema]
-                       (if-some [e (some-> schema m/properties (find key))]
-                         (constantly (val e))
-                         (some->> schema m/type (get defaults) (#(constantly (% schema))))))
+                       (or (some-> schema m/properties :default/fn)
```
ikitommi, line 400: we should call `m/eval` on this to support sci/cherry too.

ikitommi:
Thanks!

Issue #922, opened 2023-07-29
`mt/string-transformer` is not case insensitive for UUIDs


axvr:
When given an upper case UUID string, the `mt/string-transformer` does not convert it to a UUID object.

```clojure
(require '[malli.core :as m])
(require '[malli.transform :as mt])

(m/decode :uuid "b3c4e6b4-6304-4a52-99c3-cb50e737bb94" mt/string-transformer)
;; => #uuid "b3c4e6b4-6304-4a52-99c3-cb50e737bb94"

(m/decode :uuid "B3C4E6B4-6304-4A52-99C3-CB50E737BB94" mt/string-transformer)
;; => "B3C4E6B4-6304-4A52-99C3-CB50E737BB94"
```

The [UUID RFC](https://datatracker.ietf.org/doc/html/rfc4122) also states (emphasis mine):

> Each field is treated as an integer and has its value printed as a zero-filled hexadecimal digit string with the most significant digit first.  The hexadecimal values "a" through "f" are output as lower case characters and **are case insensitive on input**.

While generated UUIDs *should* be lower case as per the RFC, this is unfortunately not always the case (e.g. the `uuidgen` command).

The `clojure.core/parse-uuid` function (and the `UUID/fromString` static method that backs it) are also case insensitive.

jasonjckn:
I can do the PR for this... it's a beginner malli PR for me. 

Is there someone with merge authority that can comment whether they want this fixed?

ikitommi:
@niwinz you have been doing UUID fixes for Malli. What do you think about this?

niwinz:
I think the issue is valid, and for to be RFC compliant, we need to accept uuids in case insensitive way. I created a PR for address this.

PR #929, opened 2023-08-14
Make uuid transformer to be case insensitive

niwinz:
Fixes: https://github.com/metosin/malli/issues/922


ikitommi:
That was quick, thanks!

PR #933, opened 2023-08-16
Add :gen/return support in malli.generator

### LINKED ISSUE #932: Add malli.generator support for :gen/return
When generating values from schemas, it is not possible to directly express returning a single constant value, as in `clojure.test.check.generators/return`.

**Expected:**
`(mg/generate [:int {:gen/return 1}])` => constantly `1`

**Actual:**
Other values are returned because `:gen/return` is not recognized.

**Workarounds:**
* `(mg/generate [:int {:gen/gen (gen/return 1)}])`
* `(mg/generate [:int {:gen/elements [1]}])`
* `(mg/generate [:int {:gen/fmap (constantly 1)}])`

Although workarounds are available, they are not as idiomatic as `:gen/return x` corresponding to `(gen/return x)`.

radhikalism:
Resolves #932.

Adds idiom for `:gen/return x` in properties, just like `clojure.test.check.generators/return x`. It has similar effect to more indirect workarounds previously like `:gen/elements [x]` or `:gen/gen (gen/return x)` or `:gen/fmap (constantly x)`.

Combining `:gen/return` and `:gen/fmap` works similarly to combining `:gen/return` and `:gen/elements`. Returning `nil` from `:gen/return nil` is supported.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -469,6 +469,10 @@
 ;; Creating a generator by different means, centralized under [[-create]]
 ;;
 
+(defn- -create-from-return [props]
+  (when (contains? props :gen/return)
+    (gen/return (:gen/return props))))
+
 (defn- -create-from-elements [props]
   (some-> (:gen/elements props) gen-elements))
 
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks!

PR #942, opened 2023-09-01
fix: -string->double for cljs shouldn't allow trailing garbage

opqdonut:
(js/parseFloat "1.2abba") ==> 1.2

This was a difference between clj and cljs behaviour, but didn't have
a test.

Potentially a breaking change since somebody could've unknowingly
relied on this.

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -68,9 +68,7 @@
 
 (defn -string->double [x]
   (if (string? x)
-    (try #?(:clj  (Double/parseDouble x)
-            :cljs (let [x' (js/Number x)] (if (js/isNaN x') x x')))
-         (catch #?(:clj Exception, :cljs js/Error) _ x))
+    (or (parse-double x) x)
     x))
 
 (defn -number->double [x]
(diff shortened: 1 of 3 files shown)


src/malli/transform.cljc
```diff
@@ -69,7 +69,7 @@
 (defn -string->double [x]
   (if (string? x)
     (try #?(:clj  (Double/parseDouble x)
-            :cljs (let [x' (js/parseFloat x)] (if (js/isNaN x') x x')))
+            :cljs (let [x' (js/Number x)] (if (js/isNaN x') x x')))
```
opqdonut, line 72: or should we use `cljs.core/parse-double` & `clojure.core/parse-double`?

opqdonut: added a commit that switches to parse-double. note that parse-double was introduced recently in clojure(script) 1.11

ikitommi (approved):

opqdonut:
Merging this despite the 1.10 vs 1.11 question. Tests don't pass currently on 1.10 anyway, see #943

PR #943, opened 2023-09-01
don't pretend to support clojure 1.10

opqdonut:
We weren't really running tests on clojure 1.10 due to a mistake in the CI script (see first commit). Unsurprisingly the tests don't pass on 1.10 due to code using new functions like `random-uuid`.

Drop the fake 1.10 testing.


opqdonut:
Should we have a (backdated?) changelog entry for dropping 1.10 support? Looks like tests were broken for 1.10 as early as 2022-10, see #767

ikitommi (approved):

Issue #910, opened 2023-06-18
Validation happening during decoding when using :or


irigarae:
This is just a question to wrap my head at what's the logic behind this behaviour.
I don't know if it's exclusive to `:or` but maybe after understanding it it all makes sense. It caught me by surprise because I read [here](https://github.com/metosin/malli/blob/master/README.md#to-and-from-json): _since m/decode doesn't check the schema, you need to run m/validate (or m/explain) if you want to make sure your data conforms to your schema_.

Basically I think I'm seeing some sort of validation inside an `:or` and the decoding rolls back if it sees the decoding didn't provide valid values. This doesn't happen with basic types like `:string` `:int`, even with `:and` I couldn't find the common pattern.

- While decoding in general, there's no validation, if the transformers transforms the value, it's left as is
  ```clojure
  (m/decode
   :int :x
   (mt/transformer {:decoders {:int (constantly "1")}}))
  #_=> "1"
  ```
- same happens with `:and`
  ```clojure
  (m/decode
   [:and :int] :x ;; even with [:and :int :keyword] or [:and :keyword :int]
   (mt/transformer {:decoders {:int (constantly "1")}}))
  #_=> "1"
  ```
- however when there's an `:or` it's not transformed (**this is the cause of my confusion**)
  ```clojure
  (m/decode
   [:or :int] :x ;; same with [:or :string :int]
   (mt/transformer {:decoders {:int (constantly "1")}}))
  #_=> :x
  ```
- when the output of the transformer validates, then it's kept
  ```clojure
  (m/decode
   [:or :int] :x ;; same with [:or :string :int] and [:or :int :string]
   (mt/transformer {:decoders {:int (constantly 1)}}))
  #_=> 1
  ```
- in case one of the `:or` branches matches earlier, the transformer function doesn't even run
  ```clojure
  (m/decode
   [:or :keyword :int] :x
   (mt/transformer {:decoders {:int (fn [x] (prn x) "1")}}))
  #_=> :x
  ```
- if something matches later, then it decodes first and keeps the decoded one (if the return validates)
  ```clojure
  (m/decode
   [:or :int :keyword] :x
   (mt/transformer {:decoders {:int (fn [x] (prn x) 1)}}))
  #_=> 1
  ```
- if no result of decoding matches the schema, then all are discarded
  ```clojure
  (m/decode
   [:or :int :string] :x
   (mt/transformer {:decoders {:int (fn [x] (prn x) "1")
[...]

PR #946, opened 2023-09-03
Fallback to use result of first branch when decoding :or and :orn

ikitommi:
from #910: 

> I see. Here's one repro:
> 
> ```clojure
> (m/decode [:int {:decode/math str}] 1 (mt/transformer {:name :math})) ; => "1"
> (m/decode [:and [:int {:decode/math str}]] 1 (mt/transformer {:name :math})) ; => "1"
> (m/decode [:or [:int {:decode/math str}]] 1 (mt/transformer {:name :math})) ; => 1
> ```
> 
> so, options being:
> 
> 1. do nothing: it's a feature
> 2. use first branch in case none is valid after decode
> 3. use first branch where value is changed in decode
> 
> Not sure what is the most correct, 1 or 2 here 🤔

this PR implements option 2.

This is a **BREAKING** change as the behavior changes. But, I think the current implementation is wrong and this is (more) correct.


src/malli/core.cljc
```diff
@@ -547,6 +547,27 @@
...
+        (-intercepting this-transformer
+                       (if (= :decode method)
+                         (fn [x]
+                           (key (reduce-kv
+                                 (fn [acc i transformer]
+                                   (let [x* (transformer (val acc))]
+                                     (if ((nth validators i) x*)
+                                       (reduced (miu/-tagged x* nil))
+                                       (miu/-tagged (or (key acc) x*) x))))
+                                 (miu/-tagged nil x) transformers)))
```
opqdonut, line 564: This loop feels too complicated for what it's doing. It looks like `(val acc)` is always just the original `x`. It also looks like this could just be an application of `some` instead of a `reduce-kv`.

ikitommi: good catch, the `miu/-tagged` was not needed! I don't think `some` is best here as:
1) we have to memoize the first value (the fallback value)
2) some uses sequence abstraction, which is much slower

ikitommi: committed a [simpler loop](https://github.com/metosin/malli/pull/946/commits/85de30f2c118dad854e4244fae88413e65fe7392).

src/malli/core.cljc
```diff
@@ -555,13 +555,13 @@
...
-                                     (if ((nth validators i) x*)
-                                       (reduced (miu/-tagged x* nil))
-                                       (miu/-tagged (or (key acc) x*) x))))
-                                 (miu/-tagged nil x) transformers)))
+                           (reduce-kv
+                            (fn [acc i transformer]
+                              (let [x* (transformer x)]
+                                (if ((nth validators i) x*)
+                                  (reduced x*)
+                                  (or acc x*))))
```
opqdonut, line 563: What about when the first `transformer` returns a `nil`? Could that happen?

I'm thinking of something like `(m/decode [:maybe :int] false false-to-nil-transformer) ==> nil`

Or perhaps even just `(m/decode [:or nil :int] nil trans)`?

opqdonut (approved):
change looks good, I like the tests. one question though.

src/malli/core.cljc
ikitommi: I thought it wouldn't matter, as `[:or nil :int]` with `nil` value succeeds and doesn't use the fallback but here's a case where is fails:

```clojure
(m/decode
 [:or
  [:string {:decode/remove-secrets (constantly nil)}]
  :keyword]
 "SECRET!"
 (mt/transformer {:name :remove-secrets}))
; => "SECRET!"
```

Fixed it so that value of the first branch is always used as fallback, regardless of value.

PR #948, opened 2023-09-05
Improve map gen creation performance

bsless:
This MR improves the performance for creating a map gen ~x5, mostly by removing a call to `satisfies?`, and further by choosing performant functions in the implementation.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -157,27 +157,41 @@
     (gen-one-of gs)
     (-never-gen options)))
 
+(defn- -build-map
+  [[req opt]]
+  (persistent!
+   (reduce
+    (fn [acc [k v]]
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks for the improvements! This only effects the case where the generator is not cached, right?

```clojure
;; no cache
(mg/generate [:map [:x :int] [:y :int]])

;; instance cache
(let [s (m/schema [:map [:x :int] [:y :int]])]
  (mg/generate s))
  
;; cached function
(let [g (mg/generator [:map [:x :int] [:y :int]])]
  (mg/generate g))
```

bsless:
Gladly :)
For reference, this is the use case that made me care about construction performance
https://github.com/bsless/fsm-test-check/blob/master/examples/add_delete_tx_malli.clj

PR #949, opened 2023-09-06
Improve map gen construction and usage performance

bsless:
Round 2 :)
See detailed commit message 
Benchmark results:

```clojure
(require '[malli.generator :as mg])

(dotimes [_ 10]
  (time
   (dotimes [_ 1e5]
     (mg/generator [:map [:a :int] [:b :int]]))))

"Elapsed time: 377.558951 msecs" ;; before
"Elapsed time: 359.188257 msecs" ;; after

(def g (mg/generator [:map [:a :int] [:b :int]]))

(dotimes [_ 10]
  (time
   (dotimes [_ 1e5]
     (mg/generate g))))

"Elapsed time: 511.243682 msecs" ;; before
"Elapsed time: 370.793034 msecs" ;; after
```

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -9,7 +9,7 @@
             [clojure.test.check.rose-tree :as rose]
             [malli.core :as m]
             [malli.registry :as mr]
-            [malli.impl.util :refer [-not-any? -last -merge]]
+            [malli.impl.util :refer [-last -merge]]
             #?(:clj [borkdude.dynaload :as dynaload])))
 
 (declare generator generate -create)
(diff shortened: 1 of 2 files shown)


ikitommi:
code looks much cleaner, thanks!

bsless:
Cheers
I still wonder if generation could be faster
Three options:
- tuple of entries into map
- fmap zipmap ks to tuple gen
- tower of `gen/bind` assoc to transient map

But I'm pretty happy with how much I've been able to trim this down, too

bsless:
For the record, this brings the malli example to (only?) 2x slower than the pure gen composition
https://github.com/bsless/fsm-test-check/blob/master/examples/add_delete_tx_malli.clj

PR #952, opened 2023-09-10
Use clj-kondo `:seqable` for `:+`, `:*`, and `:?` not `{:op :rest}`

### LINKED ISSUE #820: Malli generates invalid clj-kondo type spec for [:map [:keys [:+ :keyword]]]
See https://github.com/clj-kondo/clj-kondo/issues/1922

Malli generates an invalid clj-kondo type spec for `[:map [:keys [:+ :keyword]]]`: it emits an `{:op :rest}` which should only be used to describe the type of varargs arguments. It is invalid to use this for describing map keys. Just use `:seqable` or something.

tvaisanen:
Update clj-kondo type generation. `{:op :rest}` should be used only for varargs.

```
This can be used to match remaining arguments in vararg signatures.

https://github.com/clj-kondo/clj-kondo/blob/master/doc/types.md
```

Setup for validating the changes with clj-kondo: https://github.com/tvaisanen/clj-kondo-malli-error-2022-12-20

closes #820

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -124,15 +124,15 @@
 (defmethod accept :qualified-symbol [_ _ _ _] :symbol)
 (defmethod accept :uuid [_ _ _ _] :any) ;;??
 
-(defn -seqable-or-rest [_ _ [child] {:keys [arity]}]
+(defn -seqable-or-rest [[child] {:keys [arity]}]
   (if (= arity :varargs)
     {:op :rest :spec child}
     :seqable))
 
-(defmethod accept :+ [_ _  children options] (-seqable-or-rest nil nil children options))
-(defmethod accept :* [_ _  children options] (-seqable-or-rest nil nil children options))
-(defmethod accept :? [_ _  children options] (-seqable-or-rest nil nil children options))
-(defmethod accept :repeat [_ _  children options] (-seqable-or-rest nil nil children options))
+(defmethod accept :+ [_ _  children options] (-seqable-or-rest children options))
+(defmethod accept :* [_ _  children options] (-seqable-or-rest children options))
+(defmethod accept :? [_ _  children options] (-seqable-or-rest children options))
+(defmethod accept :repeat [_ _  children options] (-seqable-or-rest children options))
 
 (defmethod accept :cat [_ _ children _] children)
 (defmethod accept :catn [_ _ children _] (mapv last children))
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -124,9 +124,9 @@
 (defmethod accept :qualified-symbol [_ _ _ _] :symbol)
 (defmethod accept :uuid [_ _ _ _] :any) ;;??
 
-(defmethod accept :+ [_ _ [child] _] {:op :rest, :spec child})
-(defmethod accept :* [_ _ [child] _] {:op :rest, :spec child})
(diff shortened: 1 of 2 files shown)


src/malli/clj_kondo.cljc
```diff
@@ -124,9 +124,9 @@
 (defmethod accept :qualified-symbol [_ _ _ _] :symbol)
 (defmethod accept :uuid [_ _ _ _] :any) ;;??
 
-(defmethod accept :+ [_ _ [child] _] {:op :rest, :spec child})
-(defmethod accept :* [_ _ [child] _] {:op :rest, :spec child})
-(defmethod accept :? [_ _ [child] _] {:op :rest, :spec child})
+(defmethod accept :+ [_ _ _ _] :seqable)
+(defmethod accept :* [_ _ _ _] :seqable)
+(defmethod accept :? [_ _ _ _] :seqable)
 (defmethod accept :repeat [_ _ [child] _] {:op :rest, :spec child})
```
ikitommi, line 130: `:repeat` is also one of the sequence schemas that can present more than 1 values.

test/malli/clj_kondo_test.cljc
```diff
@@ -63,11 +75,27 @@
          {'kikka
           {:arities {1 {:args [:int],
                         :ret :int},
-                     :varargs {:args [:int :int {:op :rest, :spec :int}],
```
ikitommi, line 66: This is a real varargs case and should generate `:rest` definition instead `:seqable`. From clj-kondo:

```markdown
{:op :rest, :spec :int}. This can be used to match remaining arguments in vararg signatures. 
```

test/malli/clj_kondo_test.cljc
```diff
@@ -63,11 +75,27 @@
...
+                     :varargs {:args [:int :int :seqable],
                                :ret :int,
                                :min-arity 2}}}
           'siren
-          {:arities {2 {:args [:ifn :coll], :ret :map}}}}}]
```
ikitommi, line 70: Good tests 💪.

ikitommi (changes requested):
Thanks for the PR! I think there are two use cases here:

1. when used in function variable arguments, `{:op :rest ...}` should be used.
2. when used in any other place, `:seqable` should be used

e.g. the `accept` should know which of the two use cases it is and return accordingly.

test/malli/clj_kondo_test.cljc
tvaisanen: I see!  Thanks.

src/malli/clj_kondo.cljc
```diff
@@ -124,10 +124,16 @@
...
-(defmethod accept :+ [_ _ [child] _] {:op :rest, :spec child})
-(defmethod accept :* [_ _ [child] _] {:op :rest, :spec child})
-(defmethod accept :? [_ _ [child] _] {:op :rest, :spec child})
-(defmethod accept :repeat [_ _ [child] _] {:op :rest, :spec child})
+(defn -seqable-or-rest [_ _ [child] {:keys [arity]}]
+  (if (= arity :varargs)
+    {:op :rest :spec child}
+    :seqable))
+
+(defmethod accept :+ [_ _  children options] (-seqable-or-rest nil nil children options))
```
ikitommi, line 132: no need to pass the extra `nil` values here?

ikitommi (approved):
Looks good, feel free to merge 💪

src/malli/clj_kondo.cljc
tvaisanen: Updated the required args.

PR #953, opened 2023-09-10
feat(assert): adds new assert-like functionality to check malli schema

jasonjckn:
First *draft* of assert functionality, 

Couple explanations on approach here
(1) the on/off switch for assert checking is modeled off of clojure.core/assert, not clojure.spec.alpha, the latter uses volatile! for on/off switch... the former is more efficient because if assertions are disabled, it won't exist in byte-code. I'm open to changing it, but that's my preference. 

(2) I added 'no color' / 'nil color' to virhe, in order to play nicely with emacs which won't render ANSI color codes in stack traces by default. 

(3) This is in malli.assert namespace temporarily, just let me know where you want it. 

image

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1,5 +1,5 @@
 (ns malli.core
-  (:refer-clojure :exclude [eval type -deref deref -lookup -key])
+  (:refer-clojure :exclude [eval type -deref deref -lookup -key assert])
   #?(:cljs (:require-macros malli.core))
   (:require #?(:clj [clojure.walk :as walk])
             [clojure.core :as c]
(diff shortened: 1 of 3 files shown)


ikitommi:
Hi. Thanks for working on this. 

> (1) the on/off switch for assert checking is modeled off of clojure.core/assert, not clojure.spec.alpha, the latter uses volatile! for on/off switch... the former is more efficient because if assertions are disabled, it won't exist in byte-code. I'm open to changing it, but that's my preference.

I like this.

> (2) I added 'no color' / 'nil color' to virhe, in order to play nicely with emacs which won't render ANSI color codes in stack traces by default.

:+1:

> (3) This is in malli.assert namespace temporarily, just let me know where you want it.

I think `assert` is important and should be in `malli.core`. Being there, it can't depend on pretty printing, could just use `malli.core/coerce` under the hood? Just created a separate issue to support pluggable pretty printing in the core: https://github.com/metosin/malli/issues/956

WDYT?

jasonjckn:
> I think assert is important and should be in malli.core. Being there, it can't depend on pretty printing, could just use malli.core/coerce under the hood? Just created a separate issue to support pluggable pretty printing in the core: https://github.com/metosin/malli/issues/956 WDYT?

Agree with everything you wrote,

I ended up calling `m/coerce` like you suggested, i was trying to think of a way to cache the result of `m/coercer` in an atom as a performance optimization, but seemed like overkill. 

 i updated the PR, please take a look

ikitommi:
Thank You.

PR #954, opened 2023-09-14
Use bound-fn in malli.dev/start! to preserve *out*

### LINKED ISSUE #920: trivial: make instrumentation less spammy
i have a lot of vars, and my unit test output is spammed with '..instrumented', perhaps we can move this to clojure.tools.logging trace? (or just delete it. ) 

https://github.com/metosin/malli/blob/master/src/malli/instrument.clj#L29

aiba:
Fixes #920


jasonjckn:
+1

ikitommi:
thanks!

PR #955, opened 2023-09-14
Add missing dependency to malli.dev.cljs

aiba:
`malli.dev.cljs/start!` references `malli.instrument` and therefore should require it from `:cljs` context.


ikitommi:
Thanks!

PR #957, opened 2023-09-20
Add :time/period schema to experimental.time

dvingo:
Adds `:time/period` to the experimental time schemas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3018,6 +3018,7 @@ The following schemas and their respective types are provided:
 | Schema                   | Example                                              | JVM/js-joda Type (`java.time`) |
 |:-------------------------|:-----------------------------------------------------|:-------------------------------|
 | `:time/duration`         | PT0.01S                                              | `Duration`                     |
+| `:time/period`           | P-1Y100D                                             | `Period`                       |
 | `:time/instant`          | 2022-12-18T12:00:25.840823567Z                       | `Instant`                      |
 | `:time/local-date`       | 2020-01-01                                           | `LocalDate`                    |
 | `:time/local-date-time`  | 2020-01-01T12:00:00                                  | `LocalDateTime`                |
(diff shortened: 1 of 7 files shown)


ikitommi:
Thanks for the PR! 

I looked quickly at [ISO_8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) and it describes also time-based durations, e.g. syntax of `PnYnMnDTnHnMnS`. 

> For example, "P3Y6M4DT12H30M5S" represents a duration of "three years, six months, four days, twelve hours, thirty minutes, and five seconds".

[TC39](https://tc39.es/proposal-temporal/docs/duration.html), it's called duration and supports also both dates and times.

In [js-joda](https://js-joda.github.io/js-joda/manual/Duration.html) & java-time, Period and Duration are separate things:

> When you write code to specify an amount of time, use the class or method that best meets your needs: the [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) class, [Period](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html) class, or the [ChronoUnit.between](https://docs.oracle.com/javase/8/docs/api/java/time/temporal/ChronoUnit.html#between-java.time.temporal.Temporal-java.time.temporal.Temporal-) method. A Duration measures an amount of time using time-based values (seconds, nanoseconds). A Period uses date-based values (years, months, days).

options:

1. go just with java-time/js-joda `Period`, add later separate `Duration`
2. go just with java-time/js-joda `Period`, add later `Duration` that is the full ISO-thing
3. go now with `:time/duration` which supports both Period & Duration

compliancy-wise I would like to see 3, but might be awkward with java-time & js-joda apis?

dvingo:
Thanks for the review and feedback! 

I wasn't aware about ISO_8601 and TC39 combining the two together, I have only been using the `java.time` (and js-joda) versions in my applications and so that's where I the `Period`/`Duration` distinction is coming from.

I have felt the need for combining the two together for a while and made my own abstraction called `Offset`  here: https://github.com/dvingo/tick-util/blob/master/src/main/space/matterandvoid/tick_utils/offset.cljc  it serializes to a string differently than described by ISO_8601, but that could easily be changed.

I was under the impression that the malli `:time/*` schemas are intended to follow the semantics of `java.time`, but it seems like the desire is to have something that matches the standards instead.

Perhaps the `Offset` idea could be adapted to implement a malli `:time/iso-duration` (or different name) to avoid conflict with the `java.time` semantics of the `duration` name?

ikitommi:
The original idea was to make something "final" but I don't think it's possible here. e.g. `java.time` is de facto and I guess TC39 will be too. But they have different concepts.

Current `time` is modelled based on java.time & js-joda, so I think this PR makes a lot of sense - and the PR is well crafted.

Will get back to this soon.

ikitommi:
yes, let's do this. Thanks!!

PR #962, opened 2023-09-24
Malli tuple should generate clj-kondo seqable

ingesolvoll:
The existing malli code generated invalid clj-kondo config. 

Before:
`[:tuple :int :int] => [:int :int]`

After
`[:tuple :int :int] => :seqable`

Context: https://clojurians.slack.com/archives/CHY97NXE2/p1676474936435069


ikitommi:
Thanks!!

PR #963, opened 2023-09-26
Fixing strip-extra-keys-transformer for recursive map encoding

tomasd:
This PR fixes an issue when `strip-extra-keys-transformer` was used to encode the `:map-of` used in a recursive way. Transformer needs to run before the encoder otherwise it sees already encoded value which does not pass the validation. 

There is a pending issue when an invalid key in `:map-of` instead of being removed from the value invalidates the whole `:map-of`. This would probably require recursive validation, which may add some computational complexity.


ikitommi:
> There is a pending issue when an invalid key in :map-of instead of being removed from the value invalidates the whole :map-of. This would probably require recursive validation, which may add some computational complexity.

a pending issue? but, merging this, thanks!

PR #966, opened 2023-10-11
Add support for uri? in malli.transform (Clojure only)

stig:
With Reitit (0.7.0-alpha7), using the below route definition, I get a failure during the coercion when passing the uri "http://example.com".

    ["/uri"
      {:post {:summary "Post a URI - fails coercion"
              :coercion reitit.coercion.malli/coercion
              :parameters {:body {:uri uri?}}
              :responses {200 {:body {:uri uri?}}}
              :handler (fn [{{{uri :uri} :body} :parameters}]
                         {:status 200 :body {:uri uri}})}}]]

When I replaced the :coercion with the Clojure Spec version, the coercion works.

With this change the Malli coercer should also work.

I made it Clojure specific since it relies on `java.net.URI`.

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -6,6 +6,7 @@
   #?(:clj (:import (java.time Instant ZoneId)
                    (java.time.format DateTimeFormatter DateTimeFormatterBuilder)
                    (java.time.temporal ChronoField)
+                   (java.net URI)
                    (java.util Date UUID))))
 
 (def ^:dynamic *max-compile-depth* 10)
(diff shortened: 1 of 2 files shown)


src/malli/transform.cljc
```diff
@@ -95,6 +96,15 @@
       x)
     x))
 
+#?(:clj
+   (defn -string->uri [x]
+     (if (string? x)
+       (try
+         (URI. x)
+         (catch Exception _
```
stig, line 104: Would it be better to catch `java.net.URISyntaxException` here?

opqdonut: Your call. The body of the `try` is so small there's little possibility of masking unrelated exceptions.

stig: You're right that the body is small, but I prefer `URISyntaxException` for explicitness - so I have updated it.

stig: Hmm, that fails the Babashka job, so I guess I change it back :-)

https://github.com/metosin/malli/actions/runs/6493910884/job/17635777011

stig: I couldn't get the recommended macro approach to work, so I just put a comment in to tighten up the catch clause once our version of babashka has the required commit.

opqdonut:
good catch re: babashka!

PR #968, opened 2023-10-23
refactor(core): support passing custom ':type' in into-schema opt for :map and :map-of

jasonjckn:
since the into-schema/'higher kinded schema' `:map`'s uses `map?` as its predicate, and no instance of datomic.query.EntityMap is a `map?` but supports most of the interfaces&protocols, i need to define my own variant, namely `:entity-map`, i saw you support changing the predicate, so why not the `:type` tag as well? This PR does that.

image


ikitommi:
thanks!

PR #973, opened 2023-11-21
Document :or vs :union vs :merge

frenchy64:
Responding to a query on slack.


README.md
colinphill-3ds, line 1: It looks like you forgot to change the top-level operation in the example schemas – they all use `:merge`

ikitommi:
Thanks!!

PR #976, opened 2023-11-30
chore(doc): updating README provider example

ptdel:
example reflects that provider creates real schemas instead of predicates (e.g. `:string` over `string?`), this applies to `value-decoders`


ikitommi:
Thanks!

PR #995, opened 2024-01-20
Update README.md

eploko:
Fixed a typo in key names in the `UnionMaps` example snippet.

The example had me scratch my head a little as there was a typo. The combined schema used to verify a map with both `:x` and `:y` keys, while the schema only had descriptions for `:x`.


ikitommi:
Thanks!

PR #1008, opened 2024-02-17
support passing cause to -fail!

flyingmachine:
add `cause` argument to `-fail!` so that it can get passed through to `ex-info`


ikitommi:
Unclear use case and tests don't pass. Feel free to reopen with use case. thanks.

PR #1012, opened 2024-02-28
README: add dedicated sections for validating vector schemas and `:and` generation

frenchy64:


README.md
```diff
@@ -500,6 +494,54 @@ it is always better to use less general tools whenever possible:
...
+```
+
+A `:tuple` schema describes a fixed length Clojure vector of heterogeneous elements:
+
+```clojure
+(m/validate [:tuple keyword? string? number?] [:bing "bang" 42])
+;; => true
+```
+
+To create a vector schema based on a seqex, use `:and`.
```
NoahTheDuke, line 516: What's a `seqex`?

ikitommi: "Sequence Expressions", many names for the same thing. I'll add that.

PR #1016, opened 2024-03-07
goog/mixin was deprecated and is now removed

ingesolvoll:
This is same fix as https://github.com/metosin/malli/pull/890

[See slack discussion for some more context](https://clojurians.slack.com/archives/CLDK6MFMK/p1709638318412439). Short version: This code is still useful for instrumenting advanced compile code, so we might want to keep it around.


ikitommi:
thanks!

PR #1017, opened 2024-03-11
Fix fn guard default configuration

### LINKED ISSUE #1018: function guard should be instrumented by default

allentiak:
Current default does not include the new fn guard.
Fixes #1018


NoahTheDuke:
Looks like your merge wasn't successful, so there are no changes present in the PR. Might want to rebase/rollback that commit.

allentiak:
Thanks for the comment.
Yeah, it seems the Codespace in which I started working used an old commit as a baseline...

I guess automated web environments are not perfect?

BTW: @ikitommi in an earlier commit on this file (https://github.com/metosin/malli/commit/3b73f64d00e14593f9c47fd546eb32d159fe8a4c) you did use the "guard" as a default parameter when validating.
Why did you change your mind later?

ikitommi:
not sure what is the queestion here. Here's how it works now:

```clojure
(m/=> kikka6 [:=>
              [:cat :int]
              :int
              [:fn {:error/message "input should be output"}
               (fn [[[arg] ret]] (= arg ret))]])
(defn kikka6 [x] (inc x))

(kikka6 1)
```

gives:

```clojure
-- Function Guard Error ----------------------------------------- Thread:1583 --

Function Var

  malli.core-test/kikka6

Guard arguments

  [[1] 2]

Guard Schema

  [:fn
   {:error/message "input should be output"}
   #object["malli.core_test$eval17032$fn__17034@1803b826"]]

Errors

  {:in [],
   :message "input should be output",
   :path [],
   :schema [:fn
            {:error/message "input should be output"}
            #object["malli.core_test$eval17032$fn__17034@1803b826"]],
   :value [[1] 2]}

More information

  https://cljdoc.org/d/metosin/malli/CURRENT/doc/function-schemas

--------------------------------------------------------------------------------
```

allentiak:
Hi,
Thanks for the message, @ikitommi 
You're right: my bad.
Sorry for the misunderstanding.
Closing...

PR #1019, opened 2024-03-14
Use `:orn` instead of `:altn` for map destructuring

frenchy64:
Related https://github.com/metosin/malli/issues/1020

Map destructuring shouldn't convert to an `:altn` because it turns into a sequence spec when nested:

```clojure
(malli.generator/generate
  [:map [:outer [:altn
                 [:map [:map [:inner :any]]]]]])
;=> {:outer [{:inner nil}]}

(m/validate
  [:map [:outer [:altn
                 [:map [:map [:inner :any]]]
                 [:seq [:* :any]]]]]
  {:outer {:inner "a"}})
;=> false
```

`:orn` is better for the job here:

```clojure
(m/validate
  [:map [:outer [:orn
                 [:map [:map [:inner :any]]]
                 [:seq [:* :any]]]]]
  {:outer {:inner "a"}})
;=> true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,7 +393,7 @@ You can use `:sequential` to describe homogeneous sequential Clojure collections
 ;; => false
 ```
 
-Malli also supports sequence regexes (also called sequence expresions) like [Seqexp](https://github.com/cgrand/seqexp) and Spec.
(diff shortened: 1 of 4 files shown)


ikitommi:
makes sense!

jasonjckn:
+1

ikitommi:
Good catch, thanks!!

PR #1021, opened 2024-03-14
Enforce entry specs in open map destructurings

frenchy64:
A spec of `[:altn [:cat k1 s1] ... [:cat kn sn] [:cat :any :any]]` is equivalent to `[:cat :any :any]`. This results in specs for destructured open maps that skip all entry checks when providing kw args.


ikitommi:
indeed. good catch. thanks!

PR #1023, opened 2024-03-16
Close #1006: choose non-null base for swagger

### LINKED ISSUE #1006: Invalid Swagger JSON with [:or :nil] alternatives
Depending on the order of alternative types including **:nil**, the Swagger JSON can become invalid.

With **:nil** at the end, it is fine:
`[:days-paid-leave-sick-child {:optional true} [:or [:int {:min 0 :max 30}] :nil]]`
becomes 
```yaml
          days-paid-leave-sick-child:
            type: integer
            format: int64
            minimum: 0
            maximum: 30
            x-anyOf:
              - type: integer
                format: int64
                minimum: 0
                maximum: 30
              - type: 'null'
```
which is valid, but with **:nil** as first item, it is invalid:
`[:days-until-sick-leave-notice {:optional true} [:or :nil [:int {:min 1 :max 4}]]]`
becomes
```yaml
          days-until-sick-leave-notice:
            type: 'null'
            x-anyOf:
              - type: 'null'
              - type: integer
                format: int64
                minimum: 1
                maximum: 4
```
which is invalid:
```yaml
type should be equal to one of the allowed values
allowedValues: array, boolean, integer, number, object, string
```

frenchy64:
Also adds logic to catch top-level usage of `:type "null"`.

### DIFF
--- a/src/malli/swagger.cljc
+++ b/src/malli/swagger.cljc
@@ -14,14 +14,17 @@
 
 (defmethod accept :not [_ _ children _] {:x-not (first children)})
 
-(defn non-null-nth [schema children]
+(defn non-null-nth
+  "Attempts to find a non-null schema to use as base.
+  If all are nullable, picks the first. This usually works out
+  because children are processed first, but can fail on
+  schemas that only contain nil like [:or :nil :nil]."
+  [schema children]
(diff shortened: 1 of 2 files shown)


ikitommi:
Thank you.

PR #1024, opened 2024-03-23
Fix `:repeat` infinite loops by bailing early

### LINKED ISSUE #761: nested :repeat sequence schema's doesn't seem to work
using 0.8.9,
if I evaluate ` (m/validate [:repeat [:repeat :int]] [[1 2 3] [4 5]])`
I'm getting
```
Execution error (OutOfMemoryError) at malli.impl.regex/repeat-validator$optionals (regex.cljc:348).
Java heap space
```
Maybe this is not the right way to create a schema for multi-dimensional seqs, but that is not obvious, and there is nothing about it in the docs...

frenchy64:
Close #761

I can't say I fully understand what's going on but this seems to fix all infinite loops, e.g.,:

```clojure
(validate [:repeat [:repeat :int]] [])
;; before: diverge
;; after: true
```

There are lots of examples like this.

### DIFF
--- a/src/malli/impl/regex.cljc
+++ b/src/malli/impl/regex.cljc
@@ -333,6 +333,14 @@
 
 ;;;; ## Repeat
 
+;; eagerly repeat a child until either:
+;; - the child consumes no elements
+;;   - then bail to check for remaining elements
+;; - we run out of repetitions via :max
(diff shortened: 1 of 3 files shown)


ikitommi:
thanks for looking into this.

Issue #1001, opened 2024-01-28
:and should short-circuit explanation - in some cases


ikitommi:
pulled from #993

# Tangential

Depending on the validation it doesn't make sense to display all error messages. Consider:

```clojure
(-> [:and
     [:vector :any]
     [:fn {:error/message "should be distinct"} #(= % (distinct %))]]
    (m/explain :invalid)
    (me/humanize))
; => ["invalid type" "should be distinct"]
```

But in other cases it might, such as a password field:
- Must contain special characters
- Must contain numbers
- Must contain lowercase and uppercase letters

I don't have a suggestion and I'd need some hammock time pondering about this problem.

frenchy64:
I don't think this is a solvable problem if we require `:and` to be involved.

If instead, we have a `:distinct` property, we can then have the explainer short circuit if the schema pred fails. Some schemas do this, but not `-simple-schema`.

If we make the necessary changes to `-simple-schema`'s explainer we can get this behavior:

```clojure
    (is (= ["invalid type"]
           (me/humanize (m/explain [:string {:distinct true}] nil))))
    (is (= ["should be distinct: \\[ provided 2 times"]
           (me/humanize (m/explain [:string {:distinct true}] "[["))))
```

To be an equivalent solution to `:and` + `:fn` it needs to be extensible, which I have some ideas here: https://github.com/metosin/malli/pull/1025

PR #1025, opened 2024-03-27
WIP: Generalize properties to constraints

frenchy64:
Merge first: https://github.com/metosin/malli/pull/1029

In prototyping phase. Overall goal is to obviate the need for `:and`, `:not`, and `:fn` schemas, in particular because of their brittle generators. It also leads more more relevant error messages https://github.com/metosin/malli/issues/1001.

TODO
- [x] validate
- [x] explain
- [x] generate
- [ ] transform
- [x] parse
- [ ] merge/union

Ideas:
- [x] (Noah) allow `:or` property `[:map {:or [:a :b :c]} ..]`
- [x] rename `:keys` to `:keyset`
- [x] enhance `:set` with the same properties for heterogeneous sets
- [x] enhance `:map-of` with keyset constraints
- [x] generalize beyond keysets, just call them "constraints".
- [x] rename `:keyset` to `:and`.
- [x] integer constraints `[:int {:> 5 :< 10}]`, easier to generate than `[:and :int [:> 5] [:< 10]]`.
- [x] make `:min` `:max` constraints so they can participate in constraint resolution
- [x] extend to `:double`
- [x] take `:gen/min` `:gen/max` into account earlier during solution generation
- [x] perhaps `:gen/and` `:gen/or` etc?
- [x] `:string` constraints
- [x] extend humanize output to structure logical operators explicitly ("should either a or b" => [:or "should a" "should b"])
- [x] `:dispatch` constraint
```clojure
[:multi {:dispatch :db/valueType}
 [:db.type/tuple [:map {:xor [:db/tupleType :db/tupleAttrs :db/tupleTypes]}
                  [:db/tupleType {:optional true} ...]
                  [:db/tupleAttrs {:optional true} ...]
                  [:db/tupleTypes {:optional true} ...]]]]
==>
[:map {:dispatch [:db/valueType
                  [:db.type/tuple [:xor :db/tupleType :db/tupleAttrs :db/tupleTypes]]]
                  [:db.type/string ...]
                  [:db.type/ref ...]}
 [:db/tupleType {:optional true} ...]
 [:db/tupleAttrs {:optional true} ...]
 [:db/tupleTypes {:optional true} ...]]


frenchy64:
Moved to https://github.com/frenchy64/malli/pull/12

Issue #780, opened 2022-11-16
Merging registries produces `child-error`


bartkl:
As asked on the Slack channel, I was wondering if it is expected behavior to get an error when explicitly stating to use the default schemas registry:

```clojure
(m/validate
  [:map
   [:x string?]]
  {:a 1})  ;; => false
```

```clojure
(m/validate
  [:map {:registry (merge (m/default-schemas))}
   [:x string?]]
  {:a 1})
;; => clojure.lang.ExceptionInfo: :malli.core/child-error {:type :enum, :properties nil, :children nil, :min 1, :max nil}
;; {:type :malli.core/child-error, :message :malli.core/child-error, :data {:type :enum, :properties nil, :children nil, :min 1, :max nil}}
```

@ikitommi already pointed out that:

> currently, the property-registry doesn’t allow `IntoSchema` instances, just `Schema`s. Original idea was that you can only present serializable things. Not sure if this is a good constraint, as you can anyway have non-serializable things in schemas, e.g `[:string {:gen/gen …generator-function-here…}]`

PR #1026, opened 2024-03-27
Better error handling when passing IntoSchema to `m/schema`

frenchy64:
Related https://github.com/metosin/malli/issues/780
Related https://github.com/metosin/malli/issues/1048

For backwards-compatibility, don't throw if an IntoSchema is provided as a property and happens to coerce to schema. Maybe we should warn in this case, because I don't think it has much utility.


ikitommi:
Let's close this and solve this via #1181 instead.

PR #1027, opened 2024-03-28
WIP: `:->`

frenchy64:
Introduce a flat syntax for function arguments.

```clojure
[:-> :any] ; [:=> :cat :any]
[:-> :int :any] ; [:=> [:cat :int] :any]
[:-> [:cat :int] :any]  ; [:=> [:cat [:cat :int]] :any]
[:-> a b c d :any] ; [:=> [:cat a b c d] :any]

;; guard property
[:-> {:fn (fn [[[arg] ret]] ...)}
 :arg :ret]
; [:=> [:cat :arg] :ret [:fn (fn [[[arg] ret]] ...)]]
```
  
- [x] disallow `:->` without children, looks like an infix op `[:=> :a :-> :b]`.
- [x] test `:fn` property
- [x] add generators for `:ifn` `:->`


src/malli/core.cljc
```diff
@@ -1760,9 +1760,28 @@
     (-type [_] :=>)
```
ikitommi, line 1760: this should be changed too

src/malli/core.cljc
```diff
@@ -1749,7 +1749,7 @@
                 {:min 1 :max 1}
                 (-regex-min-max child nested?)))))))))
 
-(defn -=>-schema []
+(defn -=>-schema [{:keys [flat-arrow?]}]
```
ikitommi, line 1752: keys with `?` in the end should be predicate functions, so `flat-arrow` would suffice here. https://github.com/bbatsov/clojure-style-guide?tab=readme-ov-file#naming-predicates

ikitommi (commented):
Thanks! I also too a stab at this earlier. Looking at my stash, the idea was:

* move the `malli.util/-util-schema` as `malli.core/-proxy-schema`
* implement `:->` using that (you can `m/deref` to get the underlaying `:=>` of this)
* there are few places where one needs to check "if the schema is a  function schema". There could be a new `FunctionSchema` protocol or some other & simpler way to tag both variants as valid.
* I just listed both e.g. in `:function` schema (oh, the crappy name of that!)

```clojure
-        (when-not (every? #(= :=> (type %)) children)
+        (when-not (every? #(= #{:=> :->} (type %)) children)
```

the actual `:->` impl was:

```clojure
(defn -->-schema []
  (-proxy-schema {:type :->
                  :fn (fn [p c o] (let [c (map #(schema % o) c), cc [(into [:cat] (butlast c)) (last c)]]
                                    [c (map -form c) (into-schema :=> p cc o)]))}))
```

not sure if that worked in all possible places, was a quick test, see https://clojurians.slack.com/archives/CLDK6MFMK/p1706446416126059

I think having a new Schema would be simpler here. WDYT?

frenchy64:
@ikitommi I took your idea and ran with it. Since you were unhappy with `:function`, I added a new proxy schema `:ifn`. LMK what you think.

```clojure
[:function [:=> [:cat] :int] [:=> [:cat :int] :int]]
==
[:ifn [:-> :int] [:-> :int :int]]
```

ikitommi:
Hey, finally had time to dig into this. I think the `:ifn` ~= `:function` gives too many ways to do one thing:

```clojure
[:function [:=> :cat :int]]
[:function [:-> :int]]
[:ifn [:=> :cat :int]]
[:ifn [:-> :int]]
```

also, `:ifn` sound bit like `ifn?` which is a different thing, e.g. a map gives true to that.

otherwise, looks good! I'll pick up this branch and play with it + add if something is missing or would like to have it differently.

ikitommi:
I dropped the `AritySchema` and moved the `-function-info` into the `Function` protocol + added some tests + renamed `:fn` property as `:guard`.

Few things I noticed:

1. explain now explains on underlaying :=> schema, might be ok?
2. there is no custom ast & to-ast for :->

ikitommi:
I dropped `:->` from the default registry and marked `m/-->-schema` as experimental due to AST and explain results potentially being unstable. But, good to go to master.

ikitommi:
Thanks!!

frenchy64:
> also, `:ifn` sound bit like `ifn?` which is a different thing, e.g. a map gives true to that.

I don't see the problem, that's also true for `:function`.

```clojure
user=> (m/validate [:function [:=> [:cat [:enum 1]] [:enum 1]]] {1 1})
true
```

PR #1029, opened 2024-03-30
Fix :map-of :min and unreachable generator, explain such-that failures

frenchy64:
1. The `:map-of` generator should use `(gen/vector-distinct-by first ..)` in order to satisfy `:min`, otherwise it can generate less than the minimum number of entries by generating the same key multiple times with different vals.

2. `such-that` and functions built on it accept a custom error message on generation failures. It's very helpful to have the precise schema that failed in the error message.

3. The unreachable elements case for `:map-of` disallows generating the empty map if `:max` was non-zero. It's perfectly fine to generate empty regardless of `:max`. Other collection generators almost made this mistake by checking `(<= (or min 0) 0 (or max 0))`, which is technically correct but simplifiable to `(= 0 (or min 0))`.

4. `coll-distinct-gen` nested generators incorrectly in the unreachable elements case. We want `(-never-gen options)`, not `(gen/fmap f (-never-gen options))` which defeats our simplification logic of checking for `-never-gen`'s metadata.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -113,7 +113,7 @@
         child (-> schema m/children first)
         gen (generator child options)]
     (if (-unreachable-gen? gen)
-      (if (<= (or min 0) 0 (or max 0))
+      (if (= 0 (or min 0))
         (gen/fmap f (gen/return []))
         (-never-gen options))
       (gen/fmap f (cond
@@ -127,11 +127,11 @@
   (let [{:keys [min max]} (-min-max schema options)
         child (-> schema m/children first)
         gen (generator child options)]
-    (gen/fmap f (if (-unreachable-gen? gen)
-                  (if (<= (or min 0) 0 (or max 0))
-                    (gen/return [])
-                    (-never-gen options))
-                  (gen/vector-distinct gen {:min-elements min, :max-elements max, :max-tries 100
+    (if (-unreachable-gen? gen)
+      (if (= 0 (or min 0))
+        (gen/return (f []))
+        (-never-gen options))
+      (gen/fmap f (gen/vector-distinct gen {:min-elements min, :max-elements max, :max-tries 100
                                             :ex-fn #(ex-info (str "Could not generate enough distinct elements for schema "
                                                                   (pr-str (m/form schema))
                                                                   ". Consider providing a custom generator.")
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -202,7 +202,7 @@
   (let [{:keys [min max]} (-min-max schema options)
         [k-gen v-gen :as gs] (map #(generator % options) (m/children schema options))]
     (if (some -unreachable-gen? gs)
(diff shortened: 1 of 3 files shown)


src/malli/generator.cljc
```diff
@@ -127,15 +127,24 @@
...
-                  (if (<= (or min 0) 0 (or max 0))
-                    (gen/return [])
-                    (-never-gen options))
-                  (gen/vector-distinct gen {:min-elements min, :max-elements max, :max-tries 100})))))
+    (if (-unreachable-gen? gen)
+      (if (= 0 (or min 0))
+        (gen/return (f []))
+        (-never-gen options))
+      (gen/fmap f (gen/vector-distinct gen {:min-elements min, :max-elements max, :max-tries 100
+                                            :ex-fn #(ex-info (str "Could not generate enough distinct elements for schema "
```
ikitommi, line 135: Malli follow a "fail fast with data, pretty errors in dev-mode" philosophy. This enables both smaller bundle sizes on ClojureScript and fast runtime in prod (things like `pr-str` are really slow).

e.g. 

```clojure
(m/schema :enum)
```
fails with just enough data, while:

```clojure

(m/schema :enum)
```

pretty prints the proper text of what happened.


Would it be possible to do same here? Currently, the `malli.dev/start!` hi-jacks actual throwing of error, I guess you can't control that here, just generating the exception, to be throwed by `test.check`?

src/malli/generator.cljc
```diff
@@ -412,11 +427,21 @@
 (defmethod -schema-generator :< [schema options] (-double-gen {:max (-> schema (m/children options) first dec)}))
 (defmethod -schema-generator :<= [schema options] (-double-gen {:max (-> schema (m/children options) first)}))
 (defmethod -schema-generator := [schema options] (gen/return (first (m/children schema options))))
-(defmethod -schema-generator :not= [schema options] (gen/such-that #(not= % (-> schema (m/children options) first)) gen/any-printable 100))
+(defmethod -schema-generator :not= [schema options] (gen/such-that #(not= % (-> schema (m/children options) first)) gen/any-printable
+                                                                   {:max-tries 100
```
ikitommi, line 431: some of these have `:max-tries`, but not all, is there a reason for this?

ikitommi (commented):
Looks good, comment about throwing & errors.

src/malli/generator.cljc
frenchy64: My best guess is that the ones without `:max-tries` always succeed in practice.

src/malli/generator.cljc
frenchy64: Thanks for the feedback.

I think it's just a matter of calling `m/-exception`, I'll look into it.

frenchy64: Could you take another look?

ikitommi (approved):

ikitommi:
Thanks! Calling the `m/-exception` doesn't currently enable the pretty printing, but will figure how to do that separate of this.

PR #1030, opened 2024-04-01
Improve range of :< :> pos? neg? generators

frenchy64:
I don't think `:max-tries` is needed here because `such-that` will increase the `size` on failures which AFAICT reliably makes `gen/double*` choose a different value.


frenchy64:
I need to check how this affects shrinking.

frenchy64:
Maybe something like this is more appropriate, I think the extra `such-that` balloons the minimum shrunk size: https://docs.oracle.com/javase/7/docs/api/java/lang/Math.html#nextAfter(double,%20double)

Not sure about cljs.

PR #1031, opened 2024-04-02
Distinguish between symbols and strings in humanize

frenchy64:
For clj use `pr-str`, for cljs use `str`.

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -3,6 +3,10 @@
             [malli.core :as m]
             [malli.util :as mu]))
 
+(defn- -pr-str [v]
+  #?(:clj (pr-str v)
+     :cljs (str v)))
+
 (defn -pred-min-max-error-fn [{:keys [pred message]}]
   (fn [{:keys [schema value]} _]
     (let [{:keys [min max]} (m/properties schema)]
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks!

danielcompton:
@frenchy64 could you explain a bit more about why you made this change? Why can’t CLJ and CLJS both use pr-str?

We have CLJC tests which include Malli error messages. Now we need to use reader conditionals on the error message to have them pass in both CLJS and CLJ.

frenchy64:
I'm told pr-str in cljs is too expensive to use because it pulls in too much code after minification.

-------- Original Message --------
On 6/1/24 03:51, Daniel Compton wrote:
@frenchy64 could you explain a bit more about why you made this change? Why can’t CLJ and CLJS both use pr-str?
We have CLJC tests which include Malli error messages. Now we need to use reader conditionals on the error message to have them pass in both CLJS and CLJ.

liwp:
To close the loop here: we ended up overriding `malli.error/-pr-str` in our tests to get consistent output:

```clj
(t/use-fixtures
  :once
  (fn [f]
    (with-redefs [me/-pr-str pr-str]
      (f))))
```

A bit naughty, but keeps reader conditionals away from the tests.

PR #1032, opened 2024-04-02
Only humanize one of min/max when different

frenchy64:
The error message can be misleading if min/max are different, since the user is told about both bounds even though they have met one of them.


ikitommi:
makes sense, thanks!

PR #1036, opened 2024-04-09
Fix actions cache

frenchy64:
`cache: maven` doesn't seem to work.


ikitommi:
thanks! is there a guide on caching with deps & github for me read / learn? is there a blog post of such?

PR #1037, opened 2024-04-10
Utility to update entry properties: update-entry-properties

frenchy64:
This is what I came up with to answer this slack question: https://clojurians.slack.com/archives/CLDK6MFMK/p1712696483105789

### DIFF
--- a/src/malli/util.cljc
+++ b/src/malli/util.cljc
@@ -3,7 +3,7 @@
   (:require [clojure.core :as c]
             [malli.core :as m]))
 
-(declare path->in)
+(declare path->in find)
 
 (defn ^:no-doc equals
   ([?schema1 ?schema2]
@@ -110,6 +110,17 @@
   [?schema f & args]
   (let [schema (m/schema ?schema)]
     (apply m/-update-properties schema f args)))
+
+(defn update-entry-properties
+  "Returns a Schema instance with updated properties for entry k."
+  [?schema k f & args]
+  (let [schema (m/schema ?schema)
+        e (or (find schema k)
+              (m/-fail! ::no-entry {:schema schema :k k}))
+        [k p v] (case (count e)
+                  2 [(first e) nil (second e)]
+                  3 e)]
+    (m/-set-entries schema [k (apply f p args)] v)))
 
 (defn closed-schema
   "Maps are implicitly open by default. They can be explicitly closed or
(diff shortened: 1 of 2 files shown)


src/malli/util.cljc
```diff
@@ -111,6 +111,17 @@
   (let [schema (m/schema ?schema)]
     (apply m/-update-properties schema f args)))
 
+(defn update-entry-properties
+  "Returns a Schema instance with updated properties for entry k."
+  [?schema k f & args]
+  (let [schema (m/schema ?schema)
+        e (or (find schema k)
+              (m/-fail! ::no-entry {:schema schema :k k}))
+        [k p v] (case (count e)
```
ikitommi, line 120: entry contains always 3 elements I believe.

test/malli/util_test.cljc
```diff
@@ -990,3 +990,25 @@
                  (mu/assoc-in [:foo :bar] :int)
                  (mu/assoc-in [:foo :baz] :int)
                  (mu/closed-schema)))))
+
+(deftest update-entry-properties-test
+  (is (= [:map [:me {:a 1, :b 1} :int]]
```
ikitommi, line 995: please add a failing test here too.

ikitommi (commented):
I think the implementation can be simplified a bit

ikitommi:
well, I can do those. merging. thanks!

ikitommi:
https://github.com/metosin/malli/commit/48544d533bf700151c886c13064c29a916a69fa9?w=1

ikitommi:
```clojure
(mu/find [:map [:x :int]] :x) 
; => [:x nil :int]
```

frenchy64:
Thanks! I was trying to reverse engineer this line, tbh I still don't understand it. https://github.com/metosin/malli/blob/48544d533bf700151c886c13064c29a916a69fa9/src/malli/core.cljc#L360

ikitommi:
`mu/find` wraps the search key into a vector of `[::m/find key]`, `m/-get-entries` unwraps it and returns the full entry instead of just the value behind the key. Entries are still always of size 3.

```clojure
(m/-get-entries (m/schema [:map [:x :int]]) :x nil)
; => :int

(m/-get-entries (m/schema [:map [:x :int]]) [::m/find :x] nil)
; => [:x nil :int]
```

PR #1040, opened 2024-04-16
Fix flaky test

frenchy64:
This test was failing with ArithmeticOverflowException's. I rewrote them to manipulate strings instead of numbers. Could also add custom generators to the arg schemas, but since this code tests guards it didn't feel right.


ikitommi:
thanks!

PR #1043, opened 2024-04-19
Add :gen/resolve property

jgdavey:
This works much like `:gen/gen`, but will first attempt to resolve the fully-qualified symbol. 

This separation means that malli schemas are free to carry around this extra data about where generators live, without having to have them loaded at runtime. Instead, generators declared in this way are dynamically loaded when used.

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -504,6 +504,17 @@
       (when-not (:gen/elements props)
         (-generator schema options))))
 
+(defn- -create-from-resolve
+  [props]
+  (when (contains? props :gen/resolve)
+    (let [sym (-> (:gen/resolve props))]
+      #?(:clj (let [v (requiring-resolve sym)]
+                (if (instance? clojure.lang.IDeref v)
+                  @v
+                  v))
+         :cljs (when (cljs.core/exists? sym)
+                 (resolve sym))))))
+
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks for the PR. Quick comments:

1. CLJS tests fail currently (`resolve` is a macro)
2. this enables running arbitrary code, so there should be a mechanism to disable this (if one reads serialised schemas from untrusted sources) - like there is with SCI.

```clojure
[:int {:gen/resolve 'mount.core/stop}]
```

jgdavey:
Thanks for the feedback!

> CLJS tests fail currently (resolve is a macro)

I'm curious if there's a pattern that makes more sense for lazy-loading that would be regardless of platform.

> this enables running arbitrary code, so there should be a mechanism to disable this (if one reads serialised schemas from untrusted sources) - like there is with SCI.

That's a good idea. I might look into sci or dynaload or somesuch for this kind of dynamic loading.

Taking a step back, is this idea *useful*? And if so, is this the way that it ought to be implemented? My desire is for a serializable schema (dependency-free) that can still point at more sophisticated generators that may or may not be available at runtime. My use-case is generators that do exist on the classpath in production, but are available to tests.

ikitommi:
I agree using `m/eval` here like @frenchy64 suggests.

Issue #846, opened 2023-02-07
Schema variables, aka generics


NoahTheDuke:
Hey friends,

Sometimes I have a function schema where an argument can be one of many different schemas and I want the return value to match the input schema. From reading the docs, I don't think this is currently possible, so I would like to register my desire for some way to do this.

Given this function definition and schema:

```clojure
(defn simple [arg]
  (if (map? arg)
    (update arg :a (fnil inc 0))
    (conj arg :success)))
(m/=> simple [:=>
               [:cat [:or [:map [:a :int]] [:vector :keyword]]]
               [:or [:map [:a :int]] [:vector :keyword]]])

(simple {:a 1}) ; {:a 2}
(simple [:a]) ; [:a :success]
```

Changing the implementation continues to validate:

```clojure
(defn simple [arg]
  (if (map? arg)
    [:a :success]
    {:a 2}))
(m/=> simple [:=>
               [:cat [:or [:map [:a :int]] [:vector :keyword]]]
               [:or [:map [:a :int]] [:vector :keyword]]])

(simple {:a 1}) ; [:a :success]
(simple [:a]) ; {:a 2}
```

@erp12 extended Malli's schema system to handle this while experimenting in their [schema inference](https://github.com/erp12/schema-inference) repo; maybe their design would work here?

Something like:

```clojure
(m/=> simple [:=> {:schemes {'S [:or [:map [:a :int]] [:vector :keyword]]}}
               [:cat 'S]
               'S])
```

erp12:
Thanks for the shout out!

I agree that polymorphic schemas abstractly pop up everywhere, albeit less frequently at the edges of applications where schemas are typically used. The canonical example of a function which would logically have polymorphic schema is `identity`. 

Our extension of the malli syntax to include schemas with parametric polymorphism was motivated by a desire for static analysis of Clojure code using malli schemas as annotations. In turn, this was used by our automatic-programming tool for synthesizing "schema safe" Clojure code from a unit test suite. The paper published at the GECCO conference can be found [here](https://arxiv.org/pdf/2206.04561.pdf), and [the project is open source](https://github.com/erp12/cbgp-lite) but not production ready.

If the malli team wants to embrace any of the schema extensions or code analysis concepts from our research, I would be happy to help port them over.

colinphill-mdsol:
I've got a use case for this at the edge of an application! My API returns a boatload of statistical information, and some patterns reoccur throughout the data. In particular, I often have summary statistics that look something like this:

[...]

PR #1042, opened 2024-04-18
WIP: Deterministic and varied pure/impure fn gen

frenchy64:
Close #1039 

FWIW I haven't worked out how to make functions shrinkable.

The implementation can be further simplified by using the private `gen/randomized` function instead of `seeded`. That might fix the babashka tests.

```clojure
;; This sequence of results should be reproducible.
(def pure   (mg/generate [:=>                    [:cat :int] :int] {:seed 0 :size 10}))
(def impure (mg/generate [:=> {:gen/impure true} [:cat :int] :int] {:seed 0 :size 10}))
(repeatedly 10 #(pure 2))
;=> (-106 -106 -106 -106 -106 -106 -106 -106 -106 -106)
(repeatedly 10 #(pure 2))
;=> (-106 -106 -106 -106 -106 -106 -106 -106 -106 -106)
(repeatedly 10 #(impure 2))
;=> (0 -1 0 -3 0 1 16 0 7 3)
(repeatedly 10 #(impure 2))
;=> (0 -1 -1 1 2 3 -2 38 -5 -12)

(mapv pure (range 10))
;=> [5 511 -106 20 -51 -322 1 0 434 -1]
(mapv pure (range 10))
;=> [5 511 -106 20 -51 -322 1 0 434 -1]
(mapv impure (range 10))
;=> [-1 0 -2 -4 -3 -4 -15 -15 -16 -2]
(mapv impure (range 10))
;=> [-1 -1 -1 0 1 3 -12 30 -2 1]
```

PR #1053, opened 2024-05-02
Polymorphic schemas

frenchy64:
Close https://github.com/metosin/malli/issues/846
Depends on https://github.com/metosin/malli/pull/1071

I went through several iterations on the internal representation. It felt correct once the serialized `form` of a polymorphic schema was human readable and ergonomic (doesn't require quoting to evaluate correctly, so no symbols allowed). The tradeoff is that we need to implement capture-avoiding substitution, and that's very error-prone. I tested a few cases.

The identity schema's `form` is `[:all [:X] [:-> :X :X]]`. It's introduced with the `m/all` macro by `(m/all [X] [:-> X X])`.

To instantiate it, you perform a `walk/postwalk-replace` with `{:X this-to-instantiate}`.

```clojure
(m/inst (m/all [X] [:-> X X])
        [:int])
;=> [:-> [:schema :int] [:schema :int]]
```

The extra `:schema`'s make sure regex's don't splice. By default, variables are single schemas.

This representation is delicate, but works and is serializable if a few rules are followed:

It's a leaf schema, and it's essential that its body is not manipulated by users before instantiating it.

For example, variables are renamed if they clash with any keywords in the body:

```clojure
(m/all [cat] [:=> [:cat cat] cat])
;=> [:all [:cat0] [:=> [:cat :cat0] :cat0]]
```

`m/inst` also does renaming to avoid capture. A variable can be "captured" by any schema, not just another variable, since variables are keywords and so are schemas.

The following renames to `:y0` because if it was `:y`, then instantiation would destroy the body via `(postwalk-replace {:y ...})`.

```clojure
(m/inst (m/all [x] (m/all [y] x))
        [(m/all [y] y)])
;=> [:all [:y0] [:schema [:all [:y] :y]]]
```

I don't know how to generate polymorphic functions, but there's some ideas in my deterministic function PR https://github.com/metosin/malli/pull/1042

The `mg/check` implementation is pretty simple and can be improved. It just instantiates variables to very small schemas of a single value.

### DIFF
--- a/docs/function-schemas.md
+++ b/docs/function-schemas.md
@@ -10,6 +10,7 @@
   * [Polymorphic Functions](#polymorphic-functions)
   * [Instrumentation](#instrumentation)
     * [Instrumentation of Polymorphic Functions](#instrumentation-of-polymorphic-functions)
+  * [Flat Arrow Function Schemas](#flat-arrow-function-schemas)
 * [Defn Schemas](#defn-schemas)
   * [Defining Function Schemas](#defining-function-schemas)
     * [Function Schema Annotations](#function-schema-annotations)
@@ -83,7 +84,10 @@ Examples of function definitions:
       [:xs [:+ :int]]] :int]
 
 ;; arg:int -> ret:int, arg > ret
-[:=> [:cat :int] :int [:fn (fn [[arg] ret] (> arg ret))]]
+(defn guard [[arg] ret]
+  (> arg ret))
+
+[:=> [:cat :int] :int [:fn guard]]
 
 ;; multi-arity function
 [:function
@@ -100,6 +104,8 @@ Examples of function definitions:
         [:sequential a]]
    [:sequential b]])
 ```
+
+What is that `:cat` all about in the input schemas? Wouldn't it be simpler without it? Sure, check out [Flat Arrow Function Schema](#flat-arrow-function-schemas).
 
 The schema for `plus` looks like this:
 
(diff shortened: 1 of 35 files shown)


docs/function-schemas.md
```diff
@@ -21,6 +23,7 @@
   * [Pretty Errors](#pretty-errors)
 * [Defn Schemas via metadata](#defn-schemas-via-metadata)
   * [TL;DR](#tldr)
+* [Polymorphic Functions](#polymorphic-schemas)
```
NoahTheDuke, line 26: This is also titled "Polymorphic Functions". Did you mean "Polymorphic Schemas"?

docs/function-schemas.md
```diff
@@ -319,11 +331,29 @@ Generating multi-arity functions:
...
+```clojure
+(m/all [a] [:=> [:cat a] a])
+```
+
+is generatively tested with schemas like
+
+```clojure
+[:=> [:cat :nil] :nil]
+[:=> [:cat [:enum 50]] [:enum 50]]
+[:=> [:cat [:enum 5333344553]] [:enum 5333344553]]
```
NoahTheDuke, line 349: Should you have three different schemas here? I know the two enums are technically different, but maybe `:nil`, `[:enum 50]` and `[:map-of :keyword :any]` or something would better clarify what's happening.

src/malli/core.cljc
```diff
@@ -2607,6 +2611,277 @@
                                   :re-transformer (fn [_ children] (apply re/alt-transformer children))
                                   :re-min-max (fn [_ children] (reduce -re-alt-min-max {:max 0} (-vmap last children)))})})
 
+(defn- -all-binder-bounds [binder]
+  (into [] (map (fn [b]
```
NoahTheDuke, line 2615: Should this and the `mapv` below use `-vmap`?

docs/function-schemas.md
frenchy64: Currently it really is this simple, more like:

```clojure
[:fn {:gen/return nil} (fn [r] (= nil r))]
[:fn {:gen/return 50} (fn [r] (= 50 r))]
[:fn {:gen/return 5333344553} (fn [r] (= 5333344553 r))]
```

NoahTheDuke: Oh interesting, okay.

frenchy64: To go further we'll need to think about what it means to shrink a schema. There are ideas in the spec impl of this, but this is the minimal proof-of-concept.

frenchy64: I will elaborate on exactly what happens in the docs though, thanks for the feedback.

frenchy64: Found the reason I needed such a convoluted representation in the first place: https://github.com/metosin/malli/issues/1070

frenchy64: > I will elaborate on exactly what happens in the docs though, thanks for the feedback.

Done.

frenchy64:
Moved to https://github.com/frenchy64/malli/pull/11

PR #725, opened 2022-07-15
Add :float type schema

hgranthorner:
Closes #722.

Please let me know if there's any tests or anything that I should add to - I couldn't find anything for `-int-schema`.

Issue #722, opened 2022-07-12
There's no `:float`


vemv:
Hi there,

I noticed there's no `:float`. `float?` is recognised, however.

Probably it would be good to leave things symmetric relative to `:double`?

My request is not esoteric as it might sound - I have a work context where floats are heavily preferred over doubles. So leaving them as "first-class citizens" would be nice.

(I am assuming that malli considers floats and doubles diferent things, i.e. that the `float?` predicate isn't simply some sort of alias for getting `double?`/`:double` behavior)

Thanks - V

frenchy64:
This can be closed, merged as https://github.com/metosin/malli/pull/1055

vemv:
Cheers @seanstrom @frenchy64

PR #1055, opened 2024-05-10
Add `:float` support

seanstrom:
### Summary

* This PR attempts to add `:float` support as described in #722 
* This PR is a follow up from the original attempt to add float support in #725 

### Notes

* I've tried to add float support to Malli while being mindful of the potentially different meanings of `float` between Clojure and ClojureScript. Specifically, I've tried to make Malli to treat floats as Java/Clojure floats (instances of `Float`) while using `clj`, and fallback to `doubles` while using `cljs`


src/malli/generator.cljc
```diff
@@ -454,6 +454,20 @@
...
+        floor-min (fn [v] (if (nil? v) min-float (max min-float v)))
+        props (m/properties schema options)
+        min-max-props (-min-max schema options)
+        infinite? (get props :gen/infinite? false)
+        NaN? (get props :gen/NaN? false)]
+    (gen/double* (merge {:infinite? infinite?
+                         :NaN? NaN?}
+                        (when (or (not infinite?) NaN?)
+                          {:min (floor-min (:min min-max-props))
+                           :max (ceil-max (:max min-max-props))})))))
```
seanstrom, line 470: Here I was unsure how to exactly generate a float or a instance of `Float`, but I attempted to constrain the generated values to something that could be coerced into a `Float` instance.

If someone knows how/where to coerce the generated value into a Float instance then please let me know 🙏

src/malli/transform.cljc
```diff
@@ -67,11 +67,30 @@
...
+       (throw (IllegalArgumentException.
+               (str "Expected string, got " (if (nil? s) "nil" (-> s class .getName))))))
+     :cljs
+     (parse-double s)))
+
+(defn -string->float [x]
+  (if (string? x)
+    (or (parse-float x) x)
+    x))
+
```
seanstrom, line 85: Here I'm providing a custom definition for parsing floats from strings. I could not find a function that used `Float/parseFloat` in this way, so I copied and modified a similar function from Clojure source code.

test/malli/generator_test.cljc
```diff
@@ -61,6 +61,25 @@
...
+          special? #(or (NaN? %)
+                        (infinity? %))
+          test-presence (fn [f options]
+                          (some f (mg/sample [:float options]
+                                             {:size 1000})))]
+      (is (test-presence infinity? {:gen/infinite? true}))
+      (is (test-presence NaN? {:gen/NaN? true}))
+      (is (test-presence special? {:gen/infinite? true
+                                   :gen/NaN? true}))
+      (is (not (test-presence special? nil)))))
```
seanstrom, line 82: It was a bit tricky writing this test since I want to use `Float/isNaN`, but sometimes when generating Infinity the tests would error because of the value being too big to coerce into a float.

Note, I don't think this test generates `Float` instances, so I think it should be updated to check for that but I'm unsure how to do that. If someone know how please let me know 🙏

seanstrom (commented):
Self-review comments 🏄

src/malli/generator.cljc
frenchy64: Call `(gen/fmap float (gen/double* ...))`.

frenchy64: Although that doesn't work for `##Inf` and `##-Inf`. Might not make sense to allow them in a float generator, at least on the jvm.

src/malli/generator.cljc
```diff
@@ -454,6 +454,20 @@
                         {:infinite? (get props :gen/infinite? false)
                          :NaN? (get props :gen/NaN? false)})
                       (-min-max schema options))))
+(defmethod -schema-generator :float [schema options]
+  (let [max-float #?(:clj Float/MAX_VALUE :cljs (.-MAX_VALUE js/Number))
+        min-float #?(:clj Float/MIN_VALUE :cljs (.-MIN_VALUE js/Number))
```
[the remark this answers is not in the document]
seanstrom: Ah yes good catch 🙏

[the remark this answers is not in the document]
frenchy64: I just meant for JVM, `(.-MIN_VALUE js/Number)` seems appropriate for cljs.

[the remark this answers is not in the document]
seanstrom: Do we want the `min-float` to be a negative number? Because I think `(.-MIN_VALUE js/Number)` would be [the smallest positive number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_VALUE#description).

frenchy64:
Oh I see, I misread the docs. Do we even need to specify the bounds on cljs since it looks like the same as double?
-------- Original Message --------

seanstrom:
> Do we even need to specify the bounds on cljs since it looks like the same as double?

Yeah good point, we probably don't need to do it, so I've changed it to only set the bounds for the clj and allow cljs to behave as doubles. I've also made sure to allow for infinity to be generated for cljs too.

src/malli/generator.cljc
```diff
@@ -454,6 +454,22 @@
...
+(defmethod -schema-generator :float [schema options]
+  (let [max-float #?(:clj Float/MAX_VALUE :cljs (.-MAX_VALUE js/Number))
+        min-float (- max-float)
+        ceil-max (fn [v] (if (nil? v) max-float (min max-float v)))
+        floor-min (fn [v] (if (nil? v) min-float (max min-float v)))
+        props (m/properties schema options)
+        min-max-props (-min-max schema options)
+        min-max-clamped #?(:clj {:min (floor-min (:min min-max-props))
+                                 :max (ceil-max (:max min-max-props))}
+                           :cljs min-max-props)
```
frenchy64, line 466: I can't think of a compelling reason to clamp the bounds, perhaps an error should be thrown instead.

Relatedly, the bounds should be coerced to float so then floating point numbers are generated regardless of bounds like in this PR https://github.com/metosin/malli/pull/1052

seanstrom: > I can't think of a compelling reason to clamp the bounds, perhaps an error should be thrown instead.

Well the main reason I thought to clamp the bounds is because we're using `gen/double*`, which can produce numbers that will lead to thrown exceptions when trying to coerce them into floats. So I think that means we always need to clamp the bounds, but maybe we can also throw an error if someone tries to override those boundaries (?).

> the bounds should be coerced to float so then floating point numbers are generated regardless of bounds

I'll push up a commit to coerce the bounds to floats too 👍

ikitommi:
Would like to merge this, but not 100% sure if everything is resolved. What do you think @frenchy64 ?

frenchy64:
IMO it's a bit inconsistent with the rest of malli that the bounds are clamped on behalf of the user in the generator. Otherwise the rest looks good.

seanstrom:
Happy to make more changes to this PR, though I'm not sure if I understand the suggestions. I mentioned in an earlier comment that there was a reason to clamp the bounds by default when generating floats. The main reason is that calling `(fmap float)` after using `gen/double*` can accidentally create exceptions when coercing the 64-bit double into a 32-bit float.

Here's an example error message:
```other
Exception: java.lang.IllegalArgumentException: Value out of range for float: -2.631375492124477E72
```

To avoid these exceptions I've included the default clamping of the bounds, because it would seem unstable to accidentally generate exceptions in test suites. There may be another way of avoiding exceptions, if so let's discuss some alternatives.

cc @ikitommi @frenchy64

seanstrom:
One thing to note, I've recently seen this documentation on MDN about [creating 32-bit floats with `Math.fround` in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/fround). Some interesting things came to mind that could help:
* Maybe we could support generating 32-bit floats for ClojureScript using `Math.fround`?
* `Math.fround` rounds a 64-bit float into 32-bit float, maybe we could do something similar in Clojure/Java?
* `Math.fround` treats numbers outside the range of the minimum and maximum 32-bit float as `Infinity` and `-Infinity`, so perhaps we could clamp the bounds by returning `Infinity` or `-Infinity`?

seanstrom:
Based on some Java documentation, it seems we can use [the `.floatValue` method on a double in Java](https://developer.android.com/reference/java/lang/Double#floatValue()). This seems to something similar to `Math.fround` where it reduces the data of the 64-bit float into a 32-bit float. It also seems to return `Infinity` and `-Infinity` for numbers that are out of the minimum and maximum range.

Based on that these two functions behave similarly and support 32-bit floats on both CLJ and CLJS, I'll try moving forward with using both of these techniques when generating `:floats` in Malli.

frenchy64:
Are there other examples of truncation in malli that you are following? To my taste this should mirror bounding a double with a bigint, what happens there? Or a hypothetical bytes generator.

I think it would be nice for the distribution of the generator to correspond to the bounds provided by the user and for the generator to always return a float. If the distribution is not possible like lower bound being too small or even nonsensical like between Double/MIN_VALUE and Double/MIN_VALUE+1, throw.

-------- Original Message --------

seanstrom:
> Are there other examples of truncation in malli that you are following?

I suppose my thought process is around: "what does it mean to generate a float"?
I can understand that we don't want to clamp boundaries and override the users intent, but by default I think we should always generate valid 32-bit floats when the user doesn't specify a boundary. Otherwise the generator, built upon `double*`, could generate an invalid 32-bit float and throw an exception.

I would like to avoid throwing exceptions by default, which is what happens now if we don't include the clamped boundaries. The test suite for running the malli float generator will create errors because of the thrown exceptions.

I could specify the boundaries in the tests when using the float generator, but that would mean each time we go to use the float generator we would need to specify the default range of 32-bit floats. And that would seem tedious since we know the default range of a 32-bit float.

> To my taste this should mirror bounding a double with a bigint, what happens there? Or a hypothetical bytes generator.

Are you asking what would happen if we try to coerce a BigInt into a 64-bit double? If so, the behaviour would be to be to create a double that has as much precision as it can allow, or otherwise return Infinity or -Infinity. Here's some related documentation: https://www.javatpoint.com/java-biginteger-doublevalue-method

> I think it would be nice for the distribution of the generator to correspond to the bounds provided by the user and for the generator to always return a float. If the distribution is not possible like lower bound being too small or even nonsensical like between Double/MIN_VALUE and Double/MIN_VALUE+1, throw.

If someone inputs min/max boundaries that are outside the range/magnitude of the 32-bit float then sure we can decide to throw. However, I don't think we should throw by default when the boundaries are not specified.

---

Based on your notes, I think we can agree that we should throw when the user provides boundaries that are unsupported by 32-bit floats. Can we also agree that we don't want to throw exceptions when no boundaries are provided to the float generator?

frenchy64:
> Based on your notes, I think we can agree that we should throw when the user provides boundaries that are unsupported by 32-bit floats. Can we also agree that we don't want to throw exceptions when no boundaries are provided to the float generator?

Exactly.

seanstrom:
Okay great! I've pushed up a commit with that change, when you have a moment can you confirm that it's okay?

src/malli/generator.cljc
```diff
@@ -456,6 +456,26 @@
...
+                                 :max (if-not (nil? (:max min-max-props))
+                                        (:max min-max-props)
+                                        max-float)}
+                           :cljs min-max-props)
+        infinite? #?(:clj false :cljs (get props :gen/infinite? false))]
+    (->> (merge {:infinite? infinite?
+                 :NaN? (get props :gen/NaN? false)}
+                (-> min-max-clamped
+                    (update :min #(some-> % float))
+                    (update :max #(some-> % float))))
```
frenchy64, line 476: I think `min-max-clamped` can be replaced with something like this:

```clojure
#(or (some-> % float) max-float)
```

seanstrom: I couldn't quite replace the code exactly with your provided example, but I have refactored the code to remove the `min-max-clamped` with this:

```clojure
(-> min-max-props
  (update :min #(or (some-> % float)
                    #?(:clj min-float :cljs nil)))
  (update :max #(or (some-> % float)
                    #?(:clj max-float :cljs nil))))
```

This code is required to be this way because we do not want to default the boundaries in ClojureScript because we want to support generating Infinity.

frenchy64 (approved):

ikitommi:
Thanks @seanstrom for doing and @frenchy64 for the review 🙇

PR #1057, opened 2024-05-10
WIP: Spec1 adapter

frenchy64:
`(from/malli :int)` returns a s/Spec.

`(from/spec int?)` returns a m/Schema.


frenchy64:
Moved to https://github.com/frenchy64/malli/pull/15

PR #1059, opened 2024-05-16
fix: swagger: collecting definitions from all parameters

### LINKED ISSUE #1002: Swagger JSON missing global definitions from schema used in request spec
I think I found a bug regarding Swagger Schema generation:

Only definitions from schemas referred to in responses are shown as global definitions in `/definitions`, but those from requests are missing.

With the following Reitit route config
```
:parameters {:body [:map [:company ::validation/company1]]}
        :responses {200 {:body ::validation/company2}}
```
only **company2** will be in `/definitions` (although the `ref$` to both def’s are there, thus causing a missing/invalid reference to **company1**).

I'm using Malli 0.14.0 and Reitit 0.7.0-alpha7

opqdonut:
fixes #1002

kudos to @mkreis for the excellent bug report!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ We use [Break Versioning][breakver]. The version numbers follow a `<major>.<mino
 
 Malli is in well matured [alpha](README.md#alpha).
 
+## UNRELEASED
+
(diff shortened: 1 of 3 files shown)


opqdonut:
rebased, added changelog

PR #1062, opened 2024-05-29
Make clj-kondo dir configurable

simonacca:
Hello,
in a monorepo setting I had the need to configure the malli-kondo integration to accommodate the clj-kondo dir being in a non-standard location.

This can of course already be achieved by overwriting a few functions from `malli.clj-kondo` and `malli.dev`.
The following changes would make this kind of setup supported out of the box.

I'd also support merging only the first commit in case you don't want to merge the `instrument!` options and the `emit!` ones.

p.s. this is the monorepo structure in question:
```
├── apps
│   ├── app1    <- clj/cljs interpreter is started here
│   └── app2
└── clj_src
    ├── .clj-kondo
    └── my-corp     <- all code lives here
```

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -206,7 +206,9 @@
    {:linters {:unresolved-symbol {:exclude ['(malli.core/=>)]}}} xs))
 
 #?(:clj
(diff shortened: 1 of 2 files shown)


ikitommi:
makes sense, thanks!

Issue #1063, opened 2024-05-31
`:?` doesn't work properly in function schemas in cljs


simonacca:
Hi there,
I encountered a discrepancy in the behavior of multi-arity schemas in clj vs cljs, seems like a bug.

In the following example, the spec is defined with the [documented syntax for multi-arity functions](https://github.com/metosin/malli/blob/master/docs/function-schemas.md#multi-arity-functions). This fails both in clj and cljs, as expected, so far so good.

```clj
(defn myfn-a
  {:malli/schema [:function 
                  [:=> [:cat :int :int ] :keyword]
                  [:=> [:cat :int :int :string] :keyword]]}
  ([a b] :res1)
  ([a b c] :res2))

(myfn-a 1 2 3)
```

The following spec however fails with "invalid function arguments" in clj (expected behavior) and **succeeds silently** (that is, it doesn't throw any error) in cljs (this is surprising to me).

```clj
(defn myfn-b
  {:malli/schema [:=> [:cat :int [:? :int] :string] :keyword]}
  ([a b] :res3)
  ([a b c] :res4))
(myfn-b 1 2 3)
```

frenchy64:
Reproduced [here](https://github.com/metosin/malli/pull/1068).

My initial assessment is that because `m/-function-info` returns `:varargs` for `[:=> [:cat :int [:? :int] :string] :keyword]`, the `cljs$core$IFn$_invoke$arity$variadic` arity is wrapped instead of the `cljs$core$IFn$_invoke$arity$` arity.

PR #1068, opened 2024-07-16
WIP: CLJS instrumentation bug

frenchy64:
https://github.com/metosin/malli/issues/1063

The instrumentation logic assumes too strong a mapping between the number of args in a schema and the number of args in the fn arity.


frenchy64:
Moved to https://github.com/frenchy64/malli/pull/14

PR #1069, opened 2024-07-16
Fix OOM error with infinitely expanding schema

### LINKED ISSUE #1061: OutOfMemoryError Java heap space exhausted with wrong schema
I stumbled upon this:

```clojure
$ clj
Clojure 1.11.1
user=> (require '[malli.core :as m])
nil
user=> (m/schema [(m/schema :any)])
Execution error (OutOfMemoryError) at malli.impl.util/-vmap (util.cljc:31).
Java heap space
```

Not sure it's a big issue, but maybe it should throw earlier an invalid schema.

frenchy64:
Close #1061 

The call `(m/schema [(m/schema :any)])` infinitely wraps `:any` using `::schema` until memory is exhausted. Stop as soon as we detect the first wrapping.

In CLJS this doesn't seem to be allowed in the first place.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -264,7 +264,11 @@
 (defn- -lookup [?schema options]
   (let [registry (-registry options)]
     (or (mr/-schema registry ?schema)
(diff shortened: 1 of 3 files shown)


ikitommi:
good catch, thanks!

PR #1072, opened 2024-07-17
WIP: Support vector syntax for references

frenchy64:
Adds full support for vector syntax for `-reference?`'s that don't point to `IntoSchema`. This removes a point of confusion about the differences between using a reference and a schema.

Properties are attached to the `-pointer`. Children are not allowed.

This changes how registries resolve IntoSchema's. The code used to recursively resolve registry entries such that `{:string-alias :string, :string (-string-schema)}` would resolve `[:string-alias]` to an instance of `(-string-schema)`. Now it just returns a `-pointer` to `:string-alias` like the (non-vector) `:string-alias` syntax does.

I added a comment where future work could add support for user-defined "schema functions" that take and return schemas. I think we can reuse the same ideas as https://github.com/metosin/malli/pull/1053 to make these hygienic and serializable.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -277,8 +277,8 @@
 (defn- -lookup-into-schema [?schema options]
   (if (into-schema? ?schema)
     ?schema
-    (when-some [?schema (-lookup ?schema options)]
-      (recur ?schema options))))
+    (some-> (-lookup ?schema options)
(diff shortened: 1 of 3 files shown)


frenchy64:
Moved to https://github.com/frenchy64/malli/pull/13

Issue #1074, opened 2024-07-19
m/-min-max-pred assumes collections are counted?


ikitommi:
```clojure
(m/validate [:sequential {:min 0} :int] (eduction identity (range 10)))
; Execution error (UnsupportedOperationException) at malli.core/-min-max-pred$fn$fn (core.cljc:637).
; count not supported on this type: Eduction
```

PR #1075, opened 2024-07-19
support min/max on uncountables like eductions

frenchy64:
Close https://github.com/metosin/malli/issues/1074


ikitommi:
Thanks!!

PR #1081, opened 2024-08-01
unstrument before instrumenting

frenchy64:


NoahTheDuke:
This has bitten me many times, thank you for fixing it.

PR #1085, opened 2024-08-07
Add support for generating protobuf3 definition #166

### LINKED ISSUE #166: Support for Protobuf3 definition/schema
Would be useful if you could generate a Protobuf3 schema in the same way you can transform to a json & swagger schemas.

https://github.com/nytimes/openapi2proto

kapilreddy:
Closes #166 

# Contribution checklist 

- [x] Tests verified by `./bin/kaocha and ./bin/node`
- [x] `malli.perf.core/quick-bench` on `malli.protobuf3-schema/transform` with a fairly complex schema takes 28.656211 µs

# Overview of change

1. Add malli.protobuf3-schema which converts malli schema to protobuf3 defintions
2. Add tests covering nested data structures
3. Add perf test for protobuf3-schema
4. Add fixes for bin/node and malli.perf.core

### DIFF
--- a/bin/node
+++ b/bin/node
@@ -28,5 +28,5 @@ case $1 in
 	advanced) advanced ;;
 	cherry-none) cherry-none ;;
 	cherry-advanced) cherry-advanced ;;
-	*) none; advanced; cherry; cherry-advanced ;;
+	*) none; advanced; cherry-none; cherry-advanced ;;
 esac
--- a/perf/malli/perf/core.cljc
+++ b/perf/malli/perf/core.cljc
@@ -3,7 +3,7 @@
             [clj-async-profiler.core :as prof]))
 
 (defn serve! []
-  (with-out-str (prof/serve-files 8080))
+  (with-out-str (prof/serve-ui 8080))
   nil)
 
 (defn clear! []
--- a/perf/malli/perf/protobuf3_schema_perf_test.cljc
+++ b/perf/malli/perf/protobuf3_schema_perf_test.cljc
@@ -1,0 +1,24 @@
+(ns malli.perf.protobuf3-schema-perf-test
+  (:require [malli.protobuf3-schema :as protobuf]
+            [malli.perf.core :as p]))
+
+(defn transform-perf []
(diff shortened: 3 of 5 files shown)


ikitommi:
Hi, and thanks for the PR! Quick comments:

1. the solution should be extendable, multimethods are a great way for doing this
2. There could be a property-based way to define the mappings, e.g. `[:fn {:protobuf/type "double"} #(double? %)]`
3. all default schemas should be mapped if they have protobuf3 counterpart, e.g. `(protobuf/transform :keyword)` fails currently

I think the solution could look much like the JSON Schema transformation:

- code: https://github.com/metosin/malli/blob/master/src/malli/json_schema.cljc 
- tests: https://github.com/metosin/malli/blob/master/test/malli/json_schema_test.cljc

what do you think?

kapilreddy:
Makes sense. I'll address these comments.

ikitommi:
for perf testing, most of the time goes in constructing the schema. It can be cached to check the actual transformation code perf. So:

```clojure
(def schema
  (m/schema
   [:map
    [:id string?]
    [:metadata [:map
                [:created_at inst?]
                [:tags [:vector string?]]]]
    [:data [:vector [:map
                     [:name string?]
                     [:details [:map
                                [:type [:enum :type-a :type-b :type-c]]
                                [:properties [:vector [:map
                                                       [:key string?]
                                                       [:value [:or string? int? boolean?]]
                                                       [:nested [:vector [:map
                                                                          [:sub_key string?]
                                                                          [:sub_value any?]]]]]]]]]]]]]))

(defn transform-perf []
  ;; 28.656211 µs -> 640ns
  (p/bench (protobuf/transform schema)))
```

kapilreddy:
That's a good catch! I'll fix the perf namespace as well.

opqdonut:
Any update on this?

juanjovazquez:
+1 to this. Useful for working with Google Gemini and Vertex API.

opqdonut:
Closing as inactive. We also feel this would be better as a separate library, instead of a core malli feature.

PR #1086, opened 2024-08-08
Distributive schemas

frenchy64:
`:merge` distributes over `:multi` in the following manner:

```clojure
(m/deref
 [:merge
  [:map [:x :int]]
  [:multi {:dispatch :y}
   [1 [:map [:y [:= 1]]]]
   [2 [:map [:y [:= 2]]]]]]
 {:registry registry})
; => [:multi {:dispatch :y}
;     [1 [:map [:x :int] [:y [:= 1]]]]
;     [2 [:map [:x :int] [:y [:= 2]]]]]

(m/deref
 [:merge
  [:multi {:dispatch :y}
   [1 [:map [:y [:= 1]]]]
   [2 [:map [:y [:= 2]]]]]
  [:map [:x :int]]]
 {:registry registry})
; => [:multi {:dispatch :y}
;     [1 [:map [:y [:= 1]] [:x :int]]]
;     [2 [:map [:y [:= 2]] [:x :int]]]]
```

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -760,6 +760,9 @@
                                  #(reduce (fn [x parser] (miu/-map-invalid reduced (parser x))) % parsers)))]
         ^{:type ::schema}
         (reify
+          DistributiveSchema
+          (-distribute-to-children [this f options']
+            (-into-schema parent properties (mapv #(f % options) children) options))
           Schema
           (-validator [_]
             (let [validators (-vmap -validator children)] (miu/-every-pred validators)))
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -799,6 +799,9 @@
                                #(reduce (fn [_ parser] (miu/-map-valid reduced (parser %))) ::invalid parsers)))]
         ^{:type ::schema}
         (reify
+          DistributiveSchema
+          (-distribute-to-children [this f options']
+            (-into-schema parent properties (mapv #(f % options) children) options))
           Schema
           (-validator [_]
             (let [validators (-vmap -validator children)] (miu/-some-pred validators)))
@@ -844,6 +847,17 @@
             cache (-create-cache options)]
         ^{:type ::schema}
         (reify
+          DistributiveSchema
+          (-distribute-to-children [this f options']
+            (-into-schema parent
+                          properties
+                          (mapv (fn [c]
+                                  (when-not (and (vector? c)
+                                                 (= 2 (count c)))
+                                    (throw (ex-info "TODO" {})))
+                                  (update c 1 f options))
+                                children)
+                          options))
           AST
           (-to-ast [this _] (-entry-ast this (-entry-keyset entry-parser)))
           Schema
@@ -1619,7 +1633,6 @@
            (-to-ast [this _] (-entry-ast this (-entry-keyset entry-parser)))
            DistributiveSchema
            (-distribute-to-children [this f options']
-             (prn "-distribute-to-children")
              (-into-schema parent
                            properties
                            (mapv (fn [c]
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -92,12 +92,16 @@
   (-function-info [this])
   (-instrument-f [schema props f options]))
 
+(defprotocol DistributiveSchema
+  (-distribute-to-children [this f options]))
+
 (defn -ref-schema? [x] (#?(:clj instance?, :cljs implements?) malli.core.RefSchema x))
 (defn -entry-parser? [x] (#?(:clj instance?, :cljs implements?) malli.core.EntryParser x))
 (defn -entry-schema? [x] (#?(:clj instance?, :cljs implements?) malli.core.EntrySchema x))
 (defn -cached? [x] (#?(:clj instance?, :cljs implements?) malli.core.Cached x))
 (defn -ast? [x] (#?(:clj instance?, :cljs implements?) malli.core.AST x))
 (defn -transformer? [x] (#?(:clj instance?, :cljs implements?) malli.core.Transformer x))
+(defn -distributive? [x] (#?(:clj instance?, :cljs implements?) malli.core.DistributiveSchema x))
 
 (extend-type #?(:clj Object, :cljs default)
   FunctionSchema
@@ -1613,6 +1617,18 @@
          (reify
            AST
            (-to-ast [this _] (-entry-ast this (-entry-keyset entry-parser)))
+           DistributiveSchema
+           (-distribute-to-children [this f options']
+             (prn "-distribute-to-children")
+             (-into-schema parent
+                           properties
+                           (mapv (fn [c]
+                                   (when-not (and (vector? c)
+                                                  (= 2 (count c)))
+                                     (throw (ex-info "TODO" {})))
+                                   (update c 1 f options))
+                                 children)
+                           options))
            Schema
            (-validator [_]
              (let [find (finder (reduce-kv (fn [acc k s] (assoc acc k (-validator s))) {} @dispatch-map))]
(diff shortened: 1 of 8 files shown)


test/malli/distributive_test.cljc
```diff
@@ -0,0 +1,79 @@
...
+          [1 [:map [:y [:= 1]] [:x :int]]]
+          [2 [:map [:y [:= 2]] [:x :int]]]]))
+
+  (is (= (m/form
+           (m/deref
+             [:merge
+              [:map [:x :int]]
+              [:or
+               [:map [:y [:= 1]]]
+               [:map [:y [:= 2]]]]]
```
ikitommi, line 64: what happens if there is non-map here? gets replaced? I would not expect it to happen.

ikitommi: maybe this should work only with `:multi` 🤔

ikitommi (commented):
I really like the approach here. What is the reasoning for this working with `:or` and `:orn` too?

test/malli/distributive_test.cljc
frenchy64: > I really like the approach here.

Seems powerful! I'm still grokking the implications and applicability myself...see below.

This is something Typed Clojure does to convert
```
(t/Merge '{:a t/Int} (t/U '{:b t/Int} '{:c t/Int})
```
to
```
(t/U '{:a t/Int :b t/Int} '{:a t/Int :c t/Int})
```
but such transformations are not so obviously applicable to Malli because of `m/parse` and error messages relying on named paths.

> What is the reasoning for this working with :or and :orn too?

I think all 3 are kinds of disjunctions so it seemed natural to at least try it. I also tried `:and` which might also be useful in some (non-map) contexts.

`:or` + `:map` is not performant but I was also curious if there are other schemas analogous to `:merge` for which `[<COMBINE> X [:or Y Z]]` => `[:or [<COMBINE> X Y] [<COMBINE> X Z]]` makes sense.

> what happens if there is non-map here? gets replaced? I would not expect it to happen.

`:merge` essentially wraps the distributed schema, so it inherits the semantics of merge in that case.

`:merge` does some weird things if you give it non-maps.

```clojure
(m/deref
  [:merge
   int?
   [:map [:x :int]]]
  options)
;=> [:map [:x :int]]

(m/deref
  [:merge
   [:map [:x :int]]
   int?]
  options)
;=> int?
```

Perhaps it should error in these cases? FWIW Typed Clojure implements `t/Merge` via a protocol [`AssocableType`](https://github.com/typedclojure/typedclojure/blob/a27632474c33a74aa26f6e4a23f136829857a2dc/typed/clj.checker/src/typed/clj/checker/assoc_utils.clj#L31) which knows how to `assoc` a particular entry onto a map. Merging then happens in an analogous way to `clojure.core/merge`, but on the type level (pour entries from right-to-left one at a time).

My first impression of `mu/merge` is that its implementation is too concrete and could be extensible.

Anyway, because of `:merge`'s behavior the distributive rule drops the schema on the left if only one of them are maps:

```clojure
(m/deref
  [:merge
   [:or
    int?
    boolean?]
   [:map [:x :int]]]
  options)
=> 
[:or [:map [:x :int]] [:map [:x :int]]]

(m/deref
  [:merge
   [:map [:x :int]]
   [:or
    int?
    boolean?]]
  options)
=>
[:or int? boolean?]
```

The idea behind this PR is to implement these distributive rules as abstractly as possible to open the door to other non-map applications, i.e.:

```clojure
[<*> X [<+> Y Z]]
=>
[<+> [<*> X Y] [<*> X Z]]

[<*> [<+> Y Z] X]
=>
[<+> [<*> Y X] [<*> Z X]]
```

Going with that theme, here `:merge` is a kind of `<*>` schema that doesn't quite fit in the abstraction because of its own inconsistent semantics around non-maps.

> maybe this should work only with :multi 🤔

For maps it seems like the most obvious winning combination. Still looking for other applications, but the way it's implemented we can always add others later.

We could also support wrapper schemas like `:schema` such as

```clojure
[:merge
 X
 [:schema Y Z]]
```

using the same approach as `m/-function-schema?`.

ikitommi:
About different registries in merge, e.g. `distribute-registry-test` I thought this as "last one wins" here.

```clojure
(def schema1
  (m/schema
   [:map ::x ::y]
   {:registry (mr/composite-registry
               (m/default-schemas)
               {::x :int
                ::y :int})}))

(def schema2
  (m/schema
   [:map ::x ::z]
   {:registry (mr/composite-registry
               (m/default-schemas)
               {::x :string
                ::z :string})}))

; looks odd
(mu/merge schema1 schema2)
;[:map
; [:user/x :string]
; [:user/y :user/y]
; [:user/z :user/z]]

(m/validate *1 {::x "1", ::y 1, ::z "1"})
; => true
```

test/malli/distributive_test.cljc
frenchy64: > It would be a breaking change to start failing with. non-entry schemas.

Agreed. A non-starter.

> I was following the [meta-merge](https://github.com/weavejester/meta-merge) functionality here:

It seems like it's all working as intended then. Applying the distributive rule would only be surprising if you expected `:merge` to have different semantics in the first place.

frenchy64:
> About different registries in merge, e.g. distribute-registry-test I thought this as "last one wins" here (like meta-merge):

Yes it works perfectly fine until you deserialize the schema at that point.

> maybe a separate issue as this already solves the :multi merge thing, which is big.

https://github.com/metosin/malli/issues/1088

frenchy64:
> Yes it works perfectly fine until you deserialize the schema at that point.

I take that back, this schema cannot be dereferenced because `::y` is not in scope for some reason.

```clojure
[:merge
 [:map
  {::y boolean?}
  [:y ::y]]
 [:map]]
```

This can be duplicated via the new distributive rule, also saying `::y` is not in scope:

```clojure
[:merge
 [:map
  {::y boolean?}
  [:y ::y]]
 [:multi {:dispatch :z}
  [1 [:map]]]]
```

frenchy64:
I documented the order of operations for distributing merge. I pushed in the middle of some work and I to fix the tests. Otherwise it's feeling pretty good.

frenchy64:
I don't know which order to apply the left/right distributive rules makes sense or is the most intuitive.

src/malli/core.cljc
```diff
@@ -1613,6 +1622,18 @@
...
            (-to-ast [this _] (-entry-ast this (-entry-keyset entry-parser)))
+           DistributiveSchema
+           (-distributive-schema? [_] true)
+           (-distribute-to-children [this f options']
+             (-into-schema parent
+                           properties
+                           (mapv (fn [c]
+                                   (when-not (and (vector? c)
+                                                  (= 2 (count c)))
+                                     (throw (ex-info "TODO" {})))
```
ikitommi, line 1633: TODO.

ikitommi (approved):
this in really, really good. I'll fix the on exception. 🥇

ikitommi:
> I don't know which order to apply the left/right distributive rules makes sense or is the most intuitive. Currently it's right then left. Here's the difference.

The current makes more sense to me.

ikitommi:
Actually, the error was not needed, `-children` of `:multi` always returns size 3 vector of `key properties value`. The old code would have failed with merging with entry properties:

```clojure
[:multi {:dispatch :y}
  [1 {:entry "properties"} [:map [:y [:= 1]]]]
  [2 [:map [:y [:= 2]]]]]
```

ikitommi:
https://github.com/metosin/malli/pull/1086/commits/be2c2be1663027ae9bc1bfa7c29fbf9aa2431fe9

ikitommi:
Run some sample tests with this, really happy with it, thanks!

PR #1090, opened 2024-08-16
allow `m/-proxy-schema` child to be a `delay`

frenchy64:
This is something that I found came in handy implementing lexical scope. In that case, there's a lot of substitution happening, and it doesn't make sense to compute the proxy child if you're just going to throw them away to replace the proxy with new children.

In general, it allows you to `-walk` a proxy schema without computing its child.


ikitommi:
Ok then.

PR #1091, opened 2024-08-16
Add `-walk` option to inherit `::val` properties

frenchy64:
This simplifies `walk-properties`, but I'm not sure it's airtight enough to release as more than a suggestion.

I considered reusing `::walk-entry-vals` but it could be a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ We use [Break Versioning][breakver]. The version numbers follow a `<major>.<mino
 
 Malli is in well matured [alpha](README.md#alpha).
 
+## NEXT
+
+* add `::m/walk-inherit-entry-props` to `-walk` for unwrapping `::val` nodes and inheriting their props after walking
+
 ## 0.16.3 (2024-08-05)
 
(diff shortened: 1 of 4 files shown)


ikitommi:
Had to revisit how the entry walking works, this is not good:

```clojure
(defn walk [schema]
  (m/walk
   schema
   (m/schema-walker identity)
   {::m/walk-entry-vals true}))

(-> [:map [:x :int]] (walk) (walk) (walk))
; [:map [:x [:malli.core/val [:malli.core/val [:malli.core/val :int]]]]]
```

I guess the right thing to do would be to do automatically the opposite somewhere. Where? in the `m/schema-walker`?

frenchy64:
> I guess the right thing to do would be to do automatically the opposite unwrapping somewhere.

Yes, I noticed the same thing.

> Where? in the m/schema-walker?

I think it would reasonable for `-inner-entries` to collapse any `::val` schemas it finds on the way back up.

ikitommi:
> I considered reusing ::walk-entry-vals but it could be a breaking change.

I think it's a bug that the entries are kept after walking, so it would be ok to to de-walk them using `::walk-entry-vals`. It's a breaking change, but for the better.

PR #1092, opened 2024-08-18
Idea: align predicate and base schemas

frenchy64:
It might reduce confusion between e.g., `int?` and `:int` if we define one in terms of the other via `-proxy-schema`.

I realize the point of the the predicates is to look like spec predicates, but it's an easy mistake to make to think `:any` and `any?` are synonymous. A better example: for `int?` it's not obvious it doesn't support `:min/max` properties.

Another advantage of using proxies here is we can reduce the implementation burden on multimethods that dispatch on `m/type`. Instead of a `defmethod` for `:any` and `any?`, you could have a `defmethod :any` and in the default method recur via `m/deref` to catch the `:any` case.

It's convenient to proxy things like `map?` as `[:map-of :any :any]`, I think all current children can stay the same.

This is just a sketch and I have no idea why the test fail.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -14,7 +14,7 @@
 
 (declare schema schema? into-schema into-schema? type eval default-registry
          -simple-schema -val-schema -ref-schema -schema-schema -registry
-         parser unparser ast from-ast -instrument ^:private -safely-countable?)
+         parser unparser ast from-ast -instrument ^:private -safely-countable? base-schemas)
 
 ;;
 ;; protocols and records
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
I love this idea and fully support it. I had no idea there were differences and I don't think this is mentioned in the readme.

ikitommi:
I would like to sunset the predicate schemas from the default registry. Too many ways to do the same thing now. This could be part of Malli 1.0.0 as it would be a breaking change. There could be a predicate registry in a seprate namespace for people who want to use this.

frenchy64:
Sounds good!

PR #1095, opened 2024-08-27
Multi keyword dispatch gen

ikitommi:
# Before

```clojure
(mg/sample
 [:multi {:dispatch :type}
  ["duck" :map]
  ["boss" :map]])
;({}
; {}
; {}
; {}
; {}
; {}
; {}
; {}
; {}
; {})
```

# After

```clojure
(mg/sample
 [:multi {:dispatch :type}
  ["duck" :map]
  ["boss" :map]])
;({:type "boss"}
; {:type "boss"}
; {:type "boss"}
; {:type "boss"}
; {:type "boss"}
; {:type "duck"}
; {:type "duck"}
; {:type "duck"}
; {:type "duck"}
; {:type "boss"})
```

... especially nice with new `:multi` merging:

```clojure
(mg/sample
 [:merge
  [:map [:name :string]]
  [:multi {:dispatch :type}
   ["duck" [:map [:speed :int]]]
   ["boss" [:map [:level :int]]]]])
;({:name "", :level 0, :type "boss"}
; {:name "", :speed -1, :type "duck"}
; {:name "", :level -1, :type "boss"}
; {:name "24m", :level -1, :type "boss"}
; {:name "8ou9", :level 4, :type "boss"}
; {:name "o", :speed -3, :type "duck"}
; {:name "", :speed 2, :type "duck"}
; {:name "", :speed -1, :type "duck"}
; {:name "", :level -15, :type "boss"}
; {:name "HWw6", :level 1, :type "boss"})
```

### DIFF
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -172,7 +172,7 @@
        #(map (fn [[k :as e]]
                (cond-> e
                  (not= ::m/default k)
-                 (update 2 mu/merge [:map [dispatch [:= k]]]))) %)
+                 (update 2 mu/merge [:map [dispatch [:= nil k]]]))) %)
        (m/options schema)))))
 
 (defn -multi-gen [schema options]
--- a/src/malli/generator.cljc
+++ b/src/malli/generator.cljc
@@ -9,6 +9,7 @@
             [clojure.test.check.rose-tree :as rose]
             [malli.core :as m]
             [malli.registry :as mr]
+            [malli.util :as mu]
             [malli.impl.util :refer [-last -merge]]
             #?(:clj [borkdude.dynaload :as dynaload])))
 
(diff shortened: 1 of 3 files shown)


frenchy64:
I wonder if there's any advantage to baking this into `:multi` itself. Perhaps applicable to json schema generation?

```clojure
(deref
  [:multi {:dispatch :type}
   "a" :map])
;=> [:merge :map [:type [:eq "a"]]]
```

ikitommi:
Schema forms should not change on schema creation -> children can be changed. This is a needed property to ensure schemas can be read back as-is from their (serialised) forms.

There should be explicit action like `deref`, `expand` or custom helper do merge this. `deref` is already used for this (e.g. with `:merge`) but it would mean that merge implementation needs to be moved from `malli.util` to `malli.core` making it even bigger and slower to load - and more code on cljs.

I think this is a good baby step, maybe move it later to `malli.util` - or to core if it makes most sense. Made it private so easy to move later.

src/malli/generator.cljc
```diff
@@ -163,10 +164,21 @@
...
 
+(defn- -merge-keyword-dispatch-map-into-entries [schema]
+  (let [dispatch (-> schema m/properties :dispatch)]
+    (cond-> schema
+      (keyword? dispatch)
+      (mu/transform-entries
+       #(map (fn [[k :as e]]
+               (cond-> e
+                 (not= ::m/default k)
+                 (update 2 mu/merge [:map [dispatch [:= k]]]))) %)
```
frenchy64, line 175: Can `k` be nil? I think that needs to become `[:= nil nil]`.

frenchy64: Ditto if `k` is a map.

ikitommi: true, good catch!

PR #1100, opened 2024-08-28
feat: json-transformer decodes 123.0 into 123 for int schemas

### LINKED ISSUE #986: json-transformer should coerce doubles with zero fractional part to longs
Per the [json schema specification](https://json-schema.org/understanding-json-schema/reference/numeric#integer) integers also accept floating point numbers with zero fractional part:

> Numbers with a zero fractional part are considered integers:
> ```
> 1.0
> ;; compliant with schema { "type": "integer" }
> ```

This breaks a bit when decoding with a json-transformer.

```clojure
(m/validate :int (m/decode :int 3.0 mt/json-transformer))
#_=> false
```

I think it would be a nice addition to the json-decoders:

```clojure
;; just an example, not taking into account clj/cljs nor NaN, etc.
(defn -float->long [x]
  (if (and (float? x) (zero? (rem x 1)))
    (long x)
    x))
(m/validate :int (m/decode :int 3.0 (mt/transformer {:decoders {:int -float->long}})))
#_=> true
;; same example
(m/decode [:int {:decode/json -float->long}] 3.0 mt/json-transformer) #_=> 3
(m/decode [:int {:decode/json -float->long}] 3.2 mt/json-transformer) #_=> 3.2
```

opqdonut:
fixes #986

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Malli is in well matured [alpha](README.md#alpha).
 * Distribute `:merge` over `:multi` [#1086](https://github.com/metosin/malli/pull/1086), see [documentation](README.md#distributive-schemas)
 * allow `m/-proxy-schema` child to be a `delay`
 * Fix `malli.dev.pretty` throws when explaining errors in nested maps [#1094](https://github.com/metosin/malli/issues/1096)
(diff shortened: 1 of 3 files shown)


ikitommi (approved):

ikitommi:
Great, thanks!

PR #1101, opened 2024-08-28
fix: (-some-pred []) should return false

### LINKED ISSUE #751: Incorrect base case for `-some-pred`
This should return false in Clojure. In ClojureScript it returns true.

```clojure
((-some-pred []) true) ;=> true
```

I think this line needs to be configurable.
https://github.com/metosin/malli/blob/bf92680ad76e57697261dd45c52dd31ffa9a8e1e/src/malli/impl/util.cljc#L59

I believe this is just morally incorrect since `:or` always has non-empty children (so preds are always non-empty).

opqdonut:
Previously, it was returning false in cljs but true in clj. Now it
returns false in both.

This is in line with `(or)` returning nil and `(and)` returning true in
clojure.

fixes #751

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -102,8 +102,6 @@ jobs:
         uses: DeLaGuardo/setup-clojure@master
         with:
           cli: latest
-          # bb: latest
-      - name: Download bb master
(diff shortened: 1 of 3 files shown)


opqdonut:
simplified the implementation, but now it fails on babashka...

ikitommi:
Thanks!

PR #1104, opened 2024-09-06
Explicitly check for CharSequence in -re-schema

alexander-yakushev:
The current behavior attempts to run `re-find` against any object in a `try/catch` block, fails with a ClassCastException on non-charsequence objects, the exception is caught and false is returned. The drawback is that the exception object is generated and its stacktrace is filled. Besides, throwing and catching an exception in Java is not free, it follows a quite slow path compared to the tight situations the validation might be used.

The profile looks like this:
image

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1466,6 +1466,8 @@
       (-check-children! :re properties children 1 1)
       (let [children (vec children)
             re (re-pattern child)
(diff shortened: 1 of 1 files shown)


alexander-yakushev:
If preferred, I can preserve the current behavior for `-explainer` as it is less performance-sensitive.

ikitommi:
thanks!

PR #1105, opened 2024-09-06
Use 2-arity <= in min-max validation

alexander-yakushev:
3-arity `<` and the like use varargs and thus allocate on invocation.


ikitommi:
Thanks!

PR #1109, opened 2024-09-27
test: Add tests to show multi usage

tomjkidd:
Wasn't able to see examples of this in the docs or tests, so figured it might be useful to make this capability more discoverable

I took a quick look to see if there were some conventions to follow, and just did something simple. If you'd prefer to have a different format or style, and even to include additional tests for the rest of the api, just let me know


ikitommi:
good stuff, thanks!

PR #1113, opened 2024-10-11
Add metadata on destructure for symbols

pfeodrippe:
With this, we can make all the keys in `mx/defn` required and we can pass, for example, ^:optional to one symbol.

So we can use like this

```clojure
(mx/defn ^:malli/required-keys eita
  [{:keys [a/b
           ^:optional a/c
           a/d]}]
  ...)

(eita {:a/b 120 :a/d 40})               ;; This would then work without :a/c being required.
```

Will add tests if we are okay with it \o

### DIFF
--- a/src/malli/destructure.cljc
+++ b/src/malli/destructure.cljc
@@ -75,19 +75,35 @@
         :when f, v vs] (f (namespace k) (str v))))
 
 (defn -keys [{:keys [keys strs syms] :as arg} {:keys [::references] :or {references true} :as options}]
-  (let [any (fn [f ks] (map (fn [k] [(f k) :any]) ks))]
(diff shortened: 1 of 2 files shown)


ikitommi:
👀

ikitommi:
It sounds like a nice way to compact information into `mx/defn`, but for now we don't want to make multiple ways to define optionality. You can build you own version of it, as many users are doing. Thanks.

Default way to define this would be:

```clojure
(def MyOpts
  [:map
   [:a/b :int]
   [:a/c {:optional true} :int]
   [:a/d :int]])


(mx/defn ^:malli/always eita
  [{:keys [a/b a/c a/d]} :- MyOpts]
  [b c d])
```

pfeodrippe:
NP, thanks!!

PR #1117, opened 2024-11-06
Elaborate `optional-keys` and `required-keys`

firoxer:
The preexisting documentation did not mention the optional argument that lets you pick which entries to modify.


o-s-silva:
A good addition to the docs, thanks!

PR #1120, opened 2024-11-13
Require generator for `:gen/fmap`

frenchy64:
`:gen/fmap` is fed `nil` if its schema cannot create a generator. This does not seem like desirable semantics because silent failures could happen if the `:gen/fmap` function is permissive, like `first` or a keyword, making the entire generator return `nil`s.

Could we reconsider this? Perhaps `:gen/fmap` predated `:gen/return` and `:gen/elements`, but we can certainly now force the user to be explicit here. Importantly, is this worth the risk breaking things by fixing it?

Also includes a fix for the CLJS version of `alphanumeric-char?` in the tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ We use [Break Versioning][breakver]. The version numbers follow a `<major>.<mino
 
 Malli is in well matured [alpha](README.md#alpha).
 
+## NEXT
+
+* **BREAKING**: `:gen/fmap` property requires its schema to create a generator.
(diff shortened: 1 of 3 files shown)


ikitommi:
I think this is a good idea.

ikitommi:
> Perhaps `:gen/fmap` predated `:gen/return` and `:gen/elements`

it is, but I as the orginal test showed, I think this was just not thought of. `clojure.test.check.generators/fmap` always requires a generator, so malli should too.

ikitommi:
thanks!

PR #1137, opened 2024-11-26
Fix heading levels in README.md

stig:
The parent heading is at level 2, so these should be at level 3 rather than level 4, to match later headings in this section.


ikitommi:
thanks!

PR #1138, opened 2024-11-26
`:not` humanizer

frenchy64:
This seems like a cheap and easy trick to automatically get a `:not` humanizer for most (simple) schemas.

Before supporting negated composite schemas we need to enhance humanizing to distinguish between conjunctions and disjunctions. Then we can leverage this for explaining negations:

```clojure
;; should be [:and "should be a string" "should be a symbol"]
(me/humanize (m/explain [:and :string :symbol] nil))
;=> ["should be a string" "should be a symbol"]

;; should be [:or "should be a string" "should be a symbol"]
(me/humanize (m/explain [:or :string :symbol] nil))
;=> ["should be a string" "should be a symbol"]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -817,7 +817,50 @@ Or if you already have a malli validation exception (e.g. in a catch form):
 
 ## Custom error messages
 
-Error messages can be customized with `:error/message` and `:error/fn` properties:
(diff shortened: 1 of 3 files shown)


ikitommi (approved):
that's clever. Great tests!

PR #1140, opened 2024-11-28
feat: accept vectors of size 2 in addition to MapEntry in unparse

### LINKED ISSUE #1123: Clarify and maybe change `unparse` behavior
As posted on [slack](https://clojurians.slack.com/archives/CLDK6MFMK/p1731615638557769), data equal to the result of `malli.core/parse`  cannot be unparsed.

``` clojure
(def int-or-string [:orn [:int :int] [:str :string]])

(malli.core/parse int-or-string 1) ;; -> [:int 1]

(= [:int 1] (malli.core/parse int-or-string 1)) ;; -> true

(malli.core/unparse int-or-string [:int 1]) ;=> :malli.core/invalid

(->> 1
     (malli.core/parse int-or-string)
     (malli.core/unparse int-or-string)) ;; => 1
```
I have some data in my schema's parsed format, which was not produced by `parse`, but which I would like to unparse. It appears that's not currently possible.

opqdonut:
fixes #1123

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -872,8 +872,8 @@
             (let [unparsers (into {} (map (fn [[k _ c]] [k (-unparser c)])) (-children this))]
               (fn [x]
                 (if (miu/-tagged? x)
-                  (if-some [unparse (get unparsers (key x))]
-                    (unparse (val x))
+                  (if-some [unparse (get unparsers (first x))]
+                    (unparse (second x))
                     ::invalid)
                   ::invalid))))
           (-transformer [this transformer method options]
@@ -1650,7 +1650,7 @@
                (fn [x] (if-some [parser (find (dispatch x))] (parser x) ::invalid))))
            (-unparser [_]
              (let [unparsers (reduce-kv (fn [acc k s] (assoc acc k (-unparser s))) {} @dispatch-map)]
-               (fn [x] (if (miu/-tagged? x) (if-some [f (unparsers (key x))] (f (val x)) ::invalid) ::invalid))))
+               (fn [x] (if (miu/-tagged? x) (if-some [f (unparsers (first x))] (f (second x)) ::invalid) ::invalid))))
            (-transformer [this transformer method options]
             ;; FIXME: Probably should not use `dispatch`
             ;; Can't use `dispatch` as `x` might not be valid before it has been unparsed:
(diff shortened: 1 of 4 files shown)


src/malli/core.cljc
```diff
@@ -872,8 +872,8 @@
             (let [unparsers (into {} (map (fn [[k _ c]] [k (-unparser c)])) (-children this))]
               (fn [x]
                 (if (miu/-tagged? x)
-                  (if-some [unparse (get unparsers (key x))]
-                    (unparse (val x))
+                  (if-some [unparse (get unparsers (first x))]
+                    (unparse (second x))
```
frenchy64, line 876: `nth` might be more equivalent in performance here since I think `first` needs to go via `seq`.

ikitommi:
```clojure
(def schema
  [:or
   [:tuple :string :keyword]
   [:orn
    ["int" :int]
    ["any" :keyword]]])

(->> (m/parse schema :any) (m/unparse schema))
; => ["any" :any]
```

opqdonut:
Ouch, that's a nice corner-case! This behaviour is the same in `master` since a `MapEntry` is a `:tuple`. The safest solution would be to switch to something like a custom record, but that wouldn't help users solve the original problem (of fabricating inputs to unparse easily).

```clj
;; in master
(m/validate [:tuple :string :keyword] (clojure.lang.MapEntry. "a" :a)) ; => true
```

ikitommi:
Indeed. So, to get correct behavior, we would  need the record. Maybe for the `:catn` case too where we use regular map now? If the record would implement right interfaces / protocols, this would not even be a breaking change?

ikitommi:
.. but would this solve the original issue?

ikitommi:
```clojure
(defrecord Tagged [key value])

(defn -tagged [key value] (->Tagged key vvalue))
(defn -tagged? [x] (instance? Tagged x))
```
+ changing access to `(:key tagged)` and `(:value tagged)`:

```clojure
(def schema
  [:or
   [:tuple :string :keyword]
   [:orn ["any" :keyword]]])

(->> :any (m/parse schema))
; => #malli.impl.util.Tag{:key "any", :value :any}

(m/unparse schema *1)
; => :any

(m/unparse schema (miu/-tagged "any" :any))
; => :any
```

I think this would solve this and no ambiguity with vectors.

opqdonut:
The record wouldn't fully solve the original problem (making custom inputs to unparse easy), but would at least solve the footgun (things look like vectors but aren't). I agree that the record would be the most correct solution.

Re: breaking, I think it would be breaking for people who are consuming the output of `parse`.

ikitommi:
> Re: breaking, I think it would be breaking for people who are consuming the output of parse.

true that. Tried to look if the parse syntax is agreed to be stable or not, no mention of that so user can assume it is, thus breaking.

PR #1141, opened 2024-12-02
Close #1121: don't generate nil if :seqable is non-empty

### LINKED ISSUE #1121: `:seqable` generates `nil` when `:min` is greater than 0
```clojure
(mg/generate [:seqable {:min 4 :max 4} :int] {:seed 0})
;=> nil

;; test
(dotimes [_ 100]
  (is (not-any? nil? (mg/sample [:seqable {:min 1} :any]))))
```

frenchy64:


ikitommi:
Thanks!

PR #1143, opened 2024-12-04
align :float and :double generators

### LINKED ISSUE #1132: `:float` accepts doubles but never generates them
```clojure
(m/validate :float (double 1.0))
;=> true
(some double? (mg/sample :float {:size 1e6}))
;=> nil
```

frenchy64:
Close #1132 

My thinking is we should follow `clojure.spec.gen.alpha` where `float?` and `double?` generate the same (and thus the corresponding schemas).


ikitommi:
thx!

PR #1144, opened 2024-12-04
refactor generator ns with -never-gen helpers

frenchy64:
This undoes a lot of the verbosity introduced when supporting `recursive-gen`.


ikitommi:
Indeed, it does. 🙇

ikitommi:
Thanks!!

PR #1146, opened 2024-12-10
look inside :ref when searching for default for :map child

### LINKED ISSUE #1145: default-value-transformer doesn't work for a `:ref` inside a `:map`
When a map key has a `:ref` schema and the key is missing, `default-value-transformer` doesn't add the key at all.

```clojure
;; This doesn't work:
(malli.core/decode [:map [:a [:ref "bing"]]]
                   {}
                   registry transformer) ; => {} EXPECTED {:a 7}

;; However, these work:
(malli.core/decode :int
                   nil
                   registry transformer) ; => 7
(malli.core/decode [:ref "bing"]
                   nil
                   registry transformer) ; => 7
(malli.core/decode [:vector [:ref "bing"]]
                   [nil]
                   registry transformer) ; => [7]
(malli.core/decode [:map [:a [:ref "bing"]]]
                   {:a nil}
                   registry transformer) ; => {:a 7}
(malli.core/decode [:map [:a :int]]
                   {}
                   registry transformer) ; => {:a 7}
```

opqdonut:
When a default value gets produced up via the normal transformer
    pipeline, the -transform method of -ref-schema will delegate to the
    child schema. However, when we're building up the default values for a
    :map in add-defaults, we're not using the transformer pipeline, but
    calling get-default manually. Thus, we need to manually deref.
    
fixes #1145

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -486,7 +486,9 @@
                                                                (when (or (not optional) add-optional-keys)
                                                                  (let [e (find p key)]
                                                                    (when-some [f (if e (constantly (val e))
-                                                                                       (get-default v))]
+                                                                                       (or (get-default v)
+                                                                                           (when (m/-ref-schema? v)
+                                                                                             (get-default (m/-deref v)))))]
                                                                      [k (fn [] (default-fn schema (f)))])))))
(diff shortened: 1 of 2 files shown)


ikitommi:
Makes sense.

PR #1147, opened 2024-12-11
Propagate options to 1-child -reducing, disallow 0-child

frenchy64:
`:union` and `:merge` don't work well with 0 and 1 children.

For 1 child, they don't propagate options to the child. This happens because the `reduce` call in `-reducing` returns init, which is a `?schema`.
- this pr ensures options are always propagated

For 0 children, deref'ing returns nil. This is because the `reduce` call in `-reducing` is passed `nil` init.
- this pr adds checks for non-empty children to `:merge` and `:union`
- fallback error for other users of `-reducing`

```clojure
(m/form (m/deref-all (m/schema [:merge [:merge :map]] {:registry (merge (mu/schemas) (m/default-schemas))})))
; (err) Execution error (ExceptionInfo) at malli.core/-exception (core.cljc:157).
; (err) :malli.core/invalid-schema

(m/deref (m/schema :merge {:registry (merge (mu/schemas) (m/default-schemas))}))
;=> nil
```

### DIFF
--- a/src/malli/util.cljc
+++ b/src/malli/util.cljc
@@ -378,8 +378,8 @@
 ;;
 
 (defn -reducing [f]
-  (fn [_ [first & rest :as children] options]
-    (let [children (mapv #(m/schema % options) children)]
+  (fn [_ children options]
+    (let [[first & rest :as children] (mapv #(m/schema % options) children)]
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks!

PR #1151, opened 2024-12-17
Update json-schema for :tuple for 2020-12 draft

dekelpilli:
The current implementation was done in 2019, before the latest JSON Schema draft (2020-12). This PR updates the generated schema to match the recommended way - https://json-schema.org/understanding-json-schema/reference/array#additionalitems
>In Draft 4 - 2019-09, tuple validation was handled by an alternate form of the items keyword. When items was an array of schemas instead of a single schema, it behaved the way prefixItems behaves.

\+

> Before to Draft 2020-12, you would use the additionalItems keyword to constrain additional items on a tuple. It works the same as items, only the name has changed.


ikitommi:
Didn't notice tuples were a thing nowadays, thanks!

PR #1158, opened 2025-01-08
fix: explain :path of ref schemas

### LINKED ISSUE #1106: `-ref-schema` returns wrong path in `-explainer` when the ref is a `var`
# What

I believe the `path` returned by the explainer of `-ref-schema` is wrong when the ref is a `var`. This originally turned up as follows:

Using `malli.error/humanize` on a schema that uses `:ref` to refer to another schema in a `var` throws `:malli.core/invalid-schema`.

For example, the following will throw:
```clj
(def Referred [:map [:foo :int]])
(def Schema [:ref #'Referred])

(me/humanize
  (m/explain Schema {:foo "2"})
  {:resolve me/-resolve-root-error})
```

Analysis indicates that this is because the `path` passed to the resolver when using `:ref` is too short: it receives `[0 :foo]`, but `(mu/get-in Schema [0 :foo])` is `nil`. The `path` should be `[0 0 :foo]`, i.e. an extra leading `0` to "deref" (?) the var of the reference:

```clj
(mu/get-in Schema [0 :foo])    ; => nil
(my/get-in Schema [0])         ; => #'dev/Referred
(mu/get-in Schema [0 0 :foo])  ; => :int
```

This causes the `schema` in the `let`-binding [here](https://github.com/metosin/malli/blob/bc80f3f0a3df5ee68634e5ce2f63182d98fa106e/src/malli/error.cljc#L274) to be `nil`, which in turn causes the `properties` call on the subsequent line to throw `invalid-schema` as `nil` is not a valid schema.

# Steps to reproduce

Add the following test to `error_test.clj`:

```clj
(deftest humanize-with-refs
  (testing "direct error" ; works!
    (is (= {:foo ["should be an integer"]}
           (with-local-vars [referred [:map [:foo :int]]]
             (me/humanize
              (m/explain [:ref referred] {:foo "2"})
              {:resolve me/-resolve-direct-error})))))

  (testing "root error" ; fails :(
    (is (= {:foo ["should be an integer"]}
           (with-local-vars [referred [:map [:foo :int]]]
             (me/humanize
              (m/explain [:ref referred] {:foo "2"})
              {:resolve me/-resolve-root-error}))))))
```

opqdonut:
The path from a [:ref ::kw] or a [:ref #'Var] to the referred schema
is [0 0] due to the intermediate -pointer. See additions to
get-in-test. Thus the -explainer of a -ref-schema needs to add [0 0]
to the path, instead of the previous [0]. This mirrors the [0 0]
already present in the -walk of -ref-schema.

fixes #1106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Malli is in well matured [alpha](README.md#alpha).
 * Docs: elaborate optional-keys and required-keys [#1117](https://github.com/metosin/malli/pull/1117)
 * **BREAKING** Output of `parse` now uses new `malli.core.Tag` and `malli.core.Tags` records for `:orn`, `:multi`, `:altn`, `:catn` etc. [#1123](https://github.com/metosin/malli/issues/1123) [#1153](https://github.com/metosin/malli/issues/1153)
   * See [Parsing](#parsing-values) and [Unparsing](#unparsing-values) for docs.
+* FIX: `:path` when explaining `:ref` errors [#1106](https://github.com/metosin/malli/issues/1106)
 
 ## 0.17.0 (2024-12-08)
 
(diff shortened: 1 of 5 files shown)


ikitommi (approved):

ikitommi:
👍

PR #1161, opened 2025-01-13
Introduce schemas for map key constraints

frenchy64:
This PR adds some schemas to help constrain the keysets of `:map` schemas. Instead of `[:and [:map ..] [:fn <OPAQUE>]]`, you now write `[:and [:map ...] [:xor ..] [:iff ..] ...]`.

The next step is to improve the generators for these schemas. The basic strategy to generate `[:and S T]` will be to generate `S` but propagate `T` as a hint to the `S` generator. Most of the work will be to canonicalize `T` by expanding into disjunctive normal form.

There are more general abstractions here (like the idea of simplifying a schema) but we can build towards that later. I have sketches of using such abstractions to compile optimal validators and generators. For now, concentrating on map keyset constraints is helping move this work along.

### DIFF
--- a/test/malli/core_test.cljc
+++ b/test/malli/core_test.cljc
@@ -3547,3 +3547,344 @@
   (is (not (m/validate [:sequential {:min 11} :int] (eduction identity (range 10)))))
   (is (not (m/validate [:seqable {:min 11} :int] (eduction identity (range 10)))))
   (is (nil? (m/explain [:sequential {:min 9} :int] (eduction identity (range 10))))))
+
+(deftest xor-test
+  (is (= [:xor [:= 1] [:= 2]]
+         (m/form (m/schema [:xor [:= 1] [:= 2]] {:registry (merge (mu/schemas) (m/default-schemas))}))))
+  (is (true? (m/validate [:xor [:enum 1 2] [:enum 2 3]] 1 {:registry (merge (mu/schemas) (m/default-schemas))})))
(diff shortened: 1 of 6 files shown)


test/malli/core_test.cljc
```diff
@@ -3581,3 +3581,356 @@
   (is (not (m/validate [:sequential {:min 11} :int] (eduction identity (range 10)))))
   (is (not (m/validate [:seqable {:min 11} :int] (eduction identity (range 10)))))
   (is (nil? (m/explain [:sequential {:min 9} :int] (eduction identity (range 10))))))
+
+(deftest xor-test
+  (is (= [:xor [:= 1] [:= 2]]
+         (m/form (m/schema [:xor [:= 1] [:= 2]] {:registry (merge (mu/schemas) (m/default-schemas))}))))
```
ikitommi, line 3587: PR adds new schemas to default registry, here expectation is that they are in `malli.util`.

ikitommi (changes requested):
Hey, thanks for working on this and sorry for being out of reach for a month+. I read the code and tested the branch, like the approach and if this is a step forward in simplification etc, let's do this. But the code should be finished first. Some notes:

- some TODOs (like transformers)
- schema walking doesn't work, e.g.

```clojure
(m/walk
 [:if [:has :user] [:has :pass] [:has :secret]]
 (m/schema-walker identity))
```

Also I `json-schema` mappings need to be done.

opqdonut:
looks like there's still a lot TODO here, labeling as WIP

frenchy64:
The merge conflicts are proving insurmountable, I'm going to release this as an independent library just to get it moving again and then we can circle back to whether they belong here.

PR #1162, opened 2025-01-14
WIP: Open :multi dispatch

frenchy64:
https://clojurians.slack.com/archives/CLDK6MFMK/p1736871015607669

Combining `:multi` with mutable registries simulates `s/multi-spec`.

### DIFF
--- a/src/malli/util.cljc
+++ b/src/malli/util.cljc
@@ -375,22 +375,19 @@
     (up schema ks f args)))
 
 (defn extend-multi
-  "Extend a :multi schema. Overwrites existing entry by = dispatch value if already present.
-  If not present, added before the default entry (if any)."
-  ([?schema e] (extend-multi ?schema e nil))
-  ([?schema e options]
+  "Extend a :multi schema. Overwrites existing entry by = dispatch value if already present."
+  ([?schema entry] (extend-multi ?schema entry nil))
+  ([?schema entry options]
    (let [s (m/schema ?schema options)
          c (m/children s)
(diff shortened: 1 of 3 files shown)


src/malli/util.cljc
```diff
@@ -373,6 +373,21 @@
                               (apply f (get s k) args))))]
     (up schema ks f args)))
 
+(defn extend-multi
```
ikitommi, line 376: what is the difference with `extend-multi` and `assoc`?

```clojure
(mu/extend-multi [:multi {:dispatch :type} [:a :string] [:b :string]] [:b :int])
; => [:multi {:dispatch :type} [:a :string] [:b :int]]

(mu/assoc [:multi {:dispatch :type} [:a :string] [:b :string]] :b :int)
; => [:multi {:dispatch :type} [:a :string] [:b :int]]
```

ikitommi:
```clojure
(extend-multi [:multi {:dispatch :type} [:a :string] [:b :string]] [:c :int])
; => [:multi {:dispatch :type} [:a :string] [:b :string] [:c :int]]

(mu/assoc [:multi {:dispatch :type} [:a :string] [:b :string]] :c :int)
; => [:multi {:dispatch :type} [:a :string] [:b :string] [:c :int]]
```

frenchy64: Ah, I think they're the same. I initially thought the `::m/default` case would need special handling but then rolled that back.

frenchy64:
I didn't consider caching and after reading https://github.com/metosin/malli/issues/786 I think this needs another look.

PR #1165, opened 2025-01-23
docs: use project relative links for malli content

### LINKED ISSUE #1164: docs: consider referencing imgs with project relative paths
## Currently...
In the malli docs, I see some lovely images.

## But...
Some of these images are referenced absolutely, for example:

```
<img src="https://raw.githubusercontent.com/metosin/malli/master/docs/img/dot.png"/>
```

## This means...
Regardless of what revision of malli docs a user is looking at, the `master` revision is always used.
This is not an issue if folks are looking at the latest docs, but if looking at older docs (with tools like cljdoc), this is not so great.

## Also...
I see a couple of images referenced like so:

```
<img src="https://github.com/metosin/malli/blob/master/docs/img/pretty-coerce.png" width=800>
```
Notice that this is not referencing the raw image. GitHub must compensate for this, but cljdoc does not, and these images are broken on cljdoc.

## Proposal
I can take a sweep through the docs and fixup img refs for you.

For example this:
```
<img src="https://raw.githubusercontent.com/metosin/malli/master/docs/img/dot.png"/>
```
Would become:
```
![](/docs/img/dot.png)
```

And this:
```
<img src="https://github.com/metosin/malli/blob/master/docs/img/pretty-coerce.png" width=800>
```
Would become:
```
<img src="/docs/img/pretty-coerce.png" width=800>
```

Sound good? If so, lemme know, and I will create a PR.

lread:
Mostly images, but also fixed one doc link from changelog.

Closes #1164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@ borkdude/edamame '1.4.25' to '1.4.27'
 
 ## 0.16.3 (2024-08-05)
 
-* `:->` added to default registry, see [documentation](https://github.com/metosin/malli/blob/master/docs/function-schemas.md#flat-arrow-function-schemas).
+* `:->` added to default registry, see [documentation](/docs/function-schemas.md#flat-arrow-function-schemas).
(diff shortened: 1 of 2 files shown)


ikitommi:
thanks!!

Issue #1166, opened 2025-01-28
When parsing, the parsed value is passed into validation functions, causing failure to parse even on schemas that validate.


escherize:
It _seems_ like this is not the intended behavior, but here's what I am seeing locally:

These print different things, but that's not a big problem:

``` clojure
(mc/validate [:and [:catn ["a" :int] ["b" :keyword]]
              [:fn (fn [x] (prn x) true)]]
             [3 :x])
;; prints: [3 :x]
;; => true

(mc/parse [:and [:catn ["a" :int] ["b" :keyword]]
           [:fn (fn [x] (prn x) true)]]
          [3 :x])
;; prints: {"a" 3, "b" :x}
;; => {"a" 3, "b" :x}
```

The problem I'm hitting is when we actually try to _validate_ these things with the attached :fn schema:

``` clojure
(mc/validate [:and [:catn ["a" :int] ["b" :keyword]]
              [:fn vector?]]
             [3 :x])
;; true

(mc/parse [:and [:catn ["a" :int] ["b" :keyword]]
           [:fn vector?]]
          [3 :x])
;; => :malli.core/invalid 😢 
```

I sort of can work around it by checking for e.g. (if it's a map with these keys, then it passes validation). but that's not really ideal.

PR #1167, opened 2025-01-28
Non-flowing `:and` parser

frenchy64:
Close https://github.com/metosin/malli/issues/1166

Could also provide a separate schema (called `and-` in spec2).

[Approach from spec2](https://github.com/clojure/spec-alpha2/blob/4cbfa677c4cd66339f18e1c122222c05c69e0d8e/src/main/clojure/clojure/alpha/spec.clj#L777)

[Original slack thread](https://clojurians.slack.com/archives/CLDK6MFMK/p1738099771219139
)

TODO
- unparsing (why is this invalid?)
```
               (let [s [:and {:parse :non-flowing}
                        [:catn ["a" :int] ["b" :keyword]]
                        [:fn vector?]]]
                 (m/unparse s (m/parse s [3 :x])))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2508,6 +2508,32 @@ Parsing returns tagged values for `:orn`, `:catn`, `:altn` and `:multi`.
 ; => #malli.core.Tag{:key :malli.core/default, :value {:type "sized", :size 1}}
 ```
 
+### :and parsing
+
+By default parsing an `:and` schema allows the intermediate results of parsing the value
+flow left-to-right. For example, the following parse fails because the output of
+parsing `:catn` is a map, not a vector.
+
+```clojure
+(m/parse [:and
+          [:catn ["a" :int] ["b" :keyword]]
+          [:fn vector?]]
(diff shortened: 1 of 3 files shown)


opqdonut:
I'm not sure I see who would ever _want_ the current behaviour of `:and` + parse? It seems to me it would be pretty difficult to use the current `:and` so that it works both when validating and parsing. I guess most users only validate, and making `:and` non-flowing wouldn't affect them?

frenchy64:
Yes, the only case I can think of is `[:and [:fn vector?] [:catn ...]]` where the simpler parsing goes first and flows to the more complex one.

frenchy64:
After some thought, I agree it makes a lot of sense to only parse the first conjunct. This also aligns with the `:and` generator, where the first conjunct provides the generator and the rest filter via their validators. I'll sketch it out in another PR.

frenchy64:
Superseded by https://github.com/metosin/malli/pull/1182

PR #1171, opened 2025-03-04
WIP: delay/future/promise schemas

frenchy64:
https://clojurians.slack.com/archives/CLDK6MFMK/p1741110236353429

TODO
- [ ] document `:deref`
- [ ] update docs to for when 3 arity deref is used

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2736,7 +2736,9 @@
             sentinel form
             [deref? deref-with-timeout? pending?]
             #?(:clj [#(instance? clojure.lang.IDeref %)
-                     #?(:bb #(future? %) ;;TODO 
+                     #?(;; IBlockingDeref not supported by babashka
+                        :bb (let [c (class (promise))]
+                              #(or (future? %) (instance? c %)))
                         :default #(instance? clojure.lang.IBlockingDeref %))
                      #(instance? clojure.lang.IPending %)]
                :cljs [#(satisfies? cljs.core.IDeref %)
@@ -2751,12 +2753,10 @@
                                 (-fail! ::delay-does-not-support-timeout))
                               delay?)
                    #?@(:clj [:future future?
-                             :promise #?(:bb (let [c (class (promise))]
-                                               #(instance? c %))
-                                         :default #(and (deref? %)
-                                                        (deref-with-timeout? %)
-                                                        (ifn? %)
-                                                        (pending? %)))]))
+                             :promise #(and (deref? %)
+                                            (deref-with-timeout? %)
+                                            (ifn? %)
+                                            (pending? %))]))
             force? (if force
                      any?
                      (case type


test/malli/core_test.cljc
```diff
@@ -3581,3 +3581,25 @@
...
+  (testing "schema matches delay"
+    (is (m/validate [:delay :int] (delay 1)))
+    (is (m/validate [:delay :int] (doto (delay 1) deref)))
+    (is (m/validate [:delay {:force true} :int] (delay 1)))
+    (is (m/validate [:delay {:force true} :int] (doto (delay 1) deref)))
+    (is (nil? (m/explain [:delay :int] (delay 1))))
+    (is (nil? (m/explain [:delay :int] (doto (delay 1) deref))))
+    (is (nil? (m/explain [:delay {:force true} :int] (delay 1)))))
+  (testing "schema does not match delay"
+    (is (m/validate [:delay :boolean] (delay 1)))
```
escherize, line 3595: This seems a bit off: it should fail right?

frenchy64: The main decision here is whether we should force delays during validation. Given that they often contain blocking code, it might not be the best idea. Here, I thought I'd make the default to only check if realized, and allow the user to tell malli when it's ok to force the delay via a property.

test/malli/core_test.cljc
```diff
@@ -3581,3 +3581,72 @@
...
+       (is (not (m/validate [:promise {:force true} :boolean] (doto (promise) (deliver 1)))))
+       (is (not (m/validate [:promise {:force true} :boolean] (doto (promise) (deliver 1)))))
+       (is (not (m/validate [:promise :boolean] 1)))
+       (is (not (m/validate [:promise :boolean] (future))))
+       (is (not (m/validate [:promise :boolean] (delay))))
+       (is (nil? (m/explain [:promise :boolean] (promise))))
+       (is (m/explain [:promise :boolean] (doto (promise) (deliver 1))))
+       (is (m/explain [:promise {:force true} :boolean] (doto (promise) (deliver 1))))
+       (is (m/explain [:promise {:force true} :boolean] (doto (promise) (deliver 1))))
+       (is (m/explain [:promise :boolean] 1)))))
```
opqdonut, line 3652: consider adding tests for get-in / walk behaviour, there's been some bugs recently in those

src/malli/core.cljc
```diff
@@ -2718,6 +2718,66 @@
                                   :re-transformer (fn [_ children] (apply re/alt-transformer children))
                                   :re-min-max (fn [_ children] (reduce -re-alt-min-max {:max 0} (-vmap last children)))})})
 
+(defn -deref-schema [{:keys [type]}]
```
opqdonut, line 2721: I understand the naming, but I fear confusion with `m/deref`. Not sure what to do about that. Docstring?

opqdonut:
nice and orthogonal addition to the lib otherwise, love it

src/malli/core.cljc
frenchy64: My original intention for this name was to have an overarching `[:deref S]` schema. I couldn't decide whether it should be a simple `reify` or a disjunction of `:delay`/`:promise`/`:future`.

I'm not sure if providing a schema called `:deref` increases or reduces clarity here.

test/malli/core_test.cljc
frenchy64: Will do.

PR #1175, opened 2025-03-12
Restore clojure 1.11 CI jobs, add JDK 23

### LINKED ISSUE #1174: Restore Clojure 1.11 CI checks
All JVM jobs test 1.12 despite jobs being duplicated for 1.11 and 1.12.

frenchy64:
Closes #1174 

Clojure 1.11 jobs are redundant after removing the `:clojure-11` alias.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -12,9 +12,10 @@ jobs:
     strategy:
       matrix:
         # Supported Java versions: LTS releases and latest
-        jdk: [8, 11, 17, 21, 23]
+        jdk: [8, 11, 17, 21]
+        clojure: [11, 12]
 
-    name: Java ${{ matrix.jdk }}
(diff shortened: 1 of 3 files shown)


opqdonut:
Our README says

> Malli requires Clojure 1.11 or ClojureScript 1.11.51.

so either we should bump that to 12, or to reinstate the clojure-11 alias to keep testing with it.

Previously, we were claiming to support clojure 1.10, but since the CI config was broken, clojure 1.11-specific code had crept in, which was a bit embarrassing.

frenchy64:
Ah good catch. I will attempt to add back the Clojure 1.11 jobs to see if they still work. Then the decision can be made on its own merits.

opqdonut (approved):

frenchy64:
Whoops, I never added JDK 23.

PR #1177, opened 2025-03-17
docs: readme: update babashka badge target

lread:
New target describes babashka badges

See https://github.com/babashka/babashka/issues/1798

(also converted to markdown syntax, I assume that's fine with you).


ikitommi:
Thanks!

PR #1178, opened 2025-03-19
schematize children once in -applying (instead of 3x)

frenchy64:
I also noticed we can skip the call to `apply` for the one schema we use `-applying` for.


opqdonut (approved):

opqdonut:
thanks, understanding what `-applying` is for took a while, but now I'm richer for the experience :wink:

ikitommi (approved):

PR #1179, opened 2025-03-20
`malli.impl.util` perf ideas

frenchy64:
Here are some ideas for performance improvements to `malli.impl.util`. These were all by eye, let me know if they don't seem obviously better and I will look into closer.

The `OR`/`AND` macro compiles to leaner JVM bytecode if you're returning booleans, originally developed in collaboration with @alexander-yakushev.


opqdonut (commented):
I happened to bump into this PR only now. Looks ok, but I would appreciate a commend _and_ some benchmark results to prove the perf improvement.

PR #1180, opened 2025-03-20
Cache validators recursively

frenchy64:
In some cases, we can save a lot of work and memory by pushing in the `-cached` logic further into the schemas. The tests demonstrate a key use-case: a schema that is only ever used as a child.

This is the output of the final line of the test on master:
```clojure
{:-validator 10, :-parser 10, :-unparser 10, :-explainer [[0] [0]]}
```

i.e., this PR saves 27 calls to protocol methods, each returning fresh fns.

Moving the `explainer` cache to the entire fn also saves recreating the fn every call.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -14,7 +14,8 @@
 
 (declare schema schema? into-schema into-schema? type eval default-registry
          -simple-schema -val-schema -ref-schema -schema-schema -registry
-         parser unparser ast from-ast -instrument ^:private -safely-countable?)
+         parser unparser ast from-ast -instrument ^:private -safely-countable?
+         -cached-validator -cached-parser -cached-unparser)
 
 ;;
 ;; protocols and records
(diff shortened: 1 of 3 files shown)


frenchy64:
There's something bugging me about the dynamism being lost here. For example, does this interfere with dynamically scoped registries in a way that matters?

ikitommi:
> There's something bugging me about the dynamism being lost here. For example, does this interfere with dynamically scoped registries in a way that matters?

I think it does. I would assume that I can recreate a schema validator with current bindings when calling `-validator` on it, no caching. Caching should be a layer, no built-it. This can cause hard to debug problems.

opqdonut:
We discussed this with Tommi & some Metosin guys. We don't feel good about this PR for a couple of reasons:

- Actual performance impact is undemonstrated. We think that eg. reitit stores the m/schema instances and calls m/validator on them, so it gets caching even now.
- Potentially surprising effects with dynamic registries etc.
- Creating a validator doesn't really need to be fast, what matters is that the validator itself is fast.

opqdonut (changes requested):
(rescinding my previous approval)

frenchy64:
This has helped spawn other ongoing work, I don't see this approach going further at the moment.

Issue #1172, opened 2025-03-11
Swagger "Could not resolve pointer" with namespaced keys


arichiardi:
Hi there, from Slack, opening a bug report for a problem we have when using `reitit.swagger/create-swagger-handler`.

We register malli spec with a namespace:

```(r/register! :survey/site-type [:enum ...])```

and we use it in another schema, say:

```clojure
(def survey
  [:map [:type :survey/type])
```

The generated ref for the parent object is `"#ref": "$ref":"#/definitions/survey~1site-type"` while in `"definitions"` we see `{"survey/site-type": ...}`.

This results in a Swagger error: `Could not resolve reference: Could not resolve pointer: #/definitions/survey~1site-type`.

Thanks for taking a look - let me know if you want me to test the fix locally.

frenchy64:
This seems like the correct output. Are you able to post a more complete (but minimal) example? Or at least the relevant parts of the `swagger.json`?

arichiardi:
I can try yes to have a minimal example but I guess it cannot be correct if swagger is returning a big red error.

frenchy64:
It could be a frontend problem like:
- https://github.com/swagger-api/swagger-ui/issues/4161
- https://github.com/swagger-api/swagger-ui/issues/1793

arichiardi:
Oh I see what you mean now. I somehow thought the two would need to match.

To put it another way you are saying that `#/definitions/survey~1site-type` should correctly refer to `survey/site-type`

Thanks for the links - I have just read that I guess we probably should not use them anyway (see https://github.com/swagger-api/swagger-ui/issues/4161#issuecomment-361738365).

frenchy64:
Maybe they do need to match in some tooling, and probably openapi.

I see that even reitit's openapi support does exactly the same thing `~1` https://github.com/metosin/reitit/blob/58195eed68bc3413704c20e95b4bf7bd7abb8c7e/test/cljc/reitit/openapi_test.clj#L1002

Though it mentions this is banned in openapi https://github.com/metosin/reitit/blob/58195eed68bc3413704c20e95b4bf7bd7abb8c7e/test/cljc/reitit/openapi_test.clj#L1023-L1025

opqdonut:
Yes this whole thing with `/` is very unfortunate. The best thing for our users would probably be to map `/` to something like `_` or `.` instead.

arichiardi:
[...]

PR #1183, opened 2025-04-01
use "." instead of "/" when creating swagger/json-schema $refs 

opqdonut:
"/" had to be encoded as "~1", and the encoding wasn't really supported by many tools.

See eg. metosin/malli#1172


frenchy64:
Can we choose a string that is invalid (or at least, less common) in a keyword, such as `@`? As-is, I think `:foo.bar` and `:foo/bar` have the same representation.

opqdonut:
That's a good point. However the official schema is very restrictive: `a-zA-Z0-9._-`. I guess I could try to add a warning that fires if there is a ref conflict.

https://github.com/seriousme/openapi-schema-validator/blob/772375bf4895f0e641d103c27140cdd1d2afc34e/schemas/v3.1/schema.json#L282

frenchy64:
Another idea is to use `_SLASH_` and refuse names that contain that substring.

```clojure
user=> (munge :asdf/foo)
"_COLON_asdf_SLASH_foo"
```

opqdonut:
The schema names are shown pretty prominently in swagger ui, so I like to make them look normal-ish

PR #1191, opened 2025-04-30
Prints better the schemas and into-schemas to the console and the REPL in CLJS

### LINKED ISSUE #1186: Printing m/schema instances in CLJS
On the CLJ side, we have a print-method implementation for schemas:

```clj
user=> (m/schema [:and [:map [:a :int]] [:map [:b :int]]])
[:and [:map [:a :int]] [:map [:b :int]]]
```

CLJS doesn't have print-method, so we get ugly behaviour

```clj
cljs.user> (m/schema [:and [:map [:a :int]] [:map [:b :int]]])
#object [malli.core.t_malli$core45030]
```

Could there be some way to print things nicely in CLJS?

The official answer seems to be to use IPrintWithWriter, see:

- https://github.com/lambdaisland/data-printers/blob/main/src/lambdaisland/data_printers.cljc#L66-L70
- https://stackoverflow.com/questions/77460039/clojure-clojurescript-how-to-plug-in-custom-print-implementation-for-edn-tag-li

green-coder:
Fixes #1186


green-coder:
### Size of Malli in a Shadow-CLJS project, release build

Before this PR:

image

After this PR:

image

opqdonut (approved):
I guess we just can't avoid the bundle size increase. Thanks!

PR #1192, opened 2025-05-05
Update clojurescript-function-instrumentation.md

danielcompton:
Clojurists Together funded work to add custom formatter support to Firefox, so it should work here too.


ikitommi:
indeed, thanks!

PR #1200, opened 2025-05-12
set description in pom.xml

opqdonut:
the pom description is shown for example on clojars


pom.xml
```diff
@@ -5,6 +5,7 @@
   <artifactId>malli</artifactId>
   <version>0.18.0</version>
   <name>malli</name>
+  <description> High-performance data-driven data specification library for Clojure/Script.</description>
```
green-coder, line 8: There is an unwanted extra space at the beginning.

PR #1203, opened 2025-05-14
Automated testing for examples in README.md

mattiuusitalo:
This PR adds an automated test step which generates Clojure tests from most of the examples in the documentation and runs them to see if they still work. It uses https://github.com/lread/test-doc-blocks to do the testing.

We have over two hundred examples in the readmes and these can be verified with test-doc-blocks. The syntax is very similar to what we are already using to document the correct usage and the result it produces in the REPL. Think:

```Clojure
(assoc {} :foo :bar)
;; => {:foo
;;     :bar}
```

# Pros
* Checks that examples keep working and reflect the current state of the library
* Nice way to avoid testing cases that won't work that doesn't mess up with the rendering in Github
* Adds over a hundred new test cases (admittedly most should be covered by the test suite)
* Can also test prints to stdout and stderr.

# Cons
* Can't check certain cases without making the examples more convoluted (for example generating data)
* Have to conform to the example writing format from test-doc-blocks

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -105,3 +105,33 @@ jobs:
           bb: latest
       - name: Run tests
         run: bb test-bb
+
+  doc-tests:
+    # Builds tests from Documentation with test-doc-blocks and tests that
+    # the examples are still valid.
+    name: Doc Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Java 11
+        uses: actions/setup-java@v4
+        with:
+          distribution: zulu
+          java-version: 11
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.m2/repository
+            ~/.deps.clj
+            ~/.gitlibs
+          key: ${{ runner.os }}-test-deps-${{ hashFiles('**/deps.edn') }}-${{ hashFiles('**/bb.edn') }}
+          restore-keys: |
+            ${{ runner.os }}-test-deps-${{ hashFiles('**/deps.edn') }}-
+            ${{ runner.os }}-test-deps-
+      - name: Setup Clojure
+        uses: DeLaGuardo/setup-clojure@master
+        with:
+          cli: latest
+          bb: latest
+      - name: Run doc tests
+        run: bb test-docs
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ Usage:
 
 (m/ast non-empty-string)
 ;; => {:type :string
-;      :properties {:min 1}}
+;;     :properties {:min 1}}
 ```
 
 Map-syntax is also called the [Schema AST](#schema-ast).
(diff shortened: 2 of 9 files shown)


.github/workflows/clojure.yml
mattiuusitalo, line 1: I didn't want to add this to the other tests, since test-doc-blocks requires clojure 1.11.  I also don't think at this point testing with the whole matrix of clojure versions is necessary.

README.md
mattiuusitalo, line 1: Changes in this file were required to make the comment blocks compatible with test-doc-blocks. I also found several small discrepancies between expected results and what it actually does. Some syntax errors as well.

docs/function-schemas.md
```diff
@@ -368,17 +373,18 @@ Example of a multi-arity function with instrumentation scopes and custom reporti
       ([x y] (* x y)))))
 
 (multi-arity-pow 4)
-; =prints=> :malli.core/invalid-output {:output [:int {:max 6}], :value 16, :args [4], :schema [:=> [:cat :int] [:int {:max 6}]]}
-; => 16
+;; =stdout=> :malli.core/invalid-output {:output [:int {:max 6}], :value 16, :args [4], :schema [:=> [:cat :int] [:int {:max 6}]]}
```
mattiuusitalo, line 376: This is an example of leveraging test-doc-blocks' capability of verifying what it printed in stdout.

gen-tests.edn
mattiuusitalo, line 1: A separate tests.edn was needed because I didn't find a way to have the tests not randomized other than at the main level. For these readme tests the order of the tests is crucial.

README.md
```diff
@@ -1731,17 +1765,17 @@ Schema paths can be converted into value paths:
...
-; => [:address :lonlat]
+; [:address :lonlat]
 ```
 
 and back, returning all paths:
 
 ```clojure
 (mu/in->paths Schema [:address :lonlat])
-; => [[0 :address 0 :lonlat]]
+; [[0 :address 0 :lonlat]]
```
opqdonut, line 1778: these couldn't be made doctestable?

gen-tests.edn
opqdonut: could the name of this file match the deps.edn alias or the bb.edn task? so something like test-docs.md or test-doc-blocks.md?

gen-tests.edn
```diff
@@ -0,0 +1,5 @@
+#kaocha/v1
+{:tests [{:id :generated
+          :test-paths  ["target/test-doc-blocks/test"]
+          :kaocha.hooks/post-load-test [lread.test-doc-blocks.kaocha.hooks/fail-on-no-tests-found]}]
+ :randomize? false}
```
opqdonut, line 5: `:reporter kaocha.report/documentation` might be nice here

opqdonut (approved):
LGTM, some suggestions tho

gen-tests.edn
mattiuusitalo: Good call. Let's do this and I will add a doc in the file to describe why the separate edn was needed.

README.md
mattiuusitalo: This actually can, good call.

gen-tests.edn
mattiuusitalo: Let's do that!

opqdonut (approved):
:fire:

PR #1205, opened 2025-05-15
Fix JSON transformer string conversions when using enum or equality schemas

### LINKED ISSUE #1204: Equality schema (:=) encoding with JSON transformer converts values to strings
We're trying to get on the latest version of Malli since we're far behind - currently on 0.12.0, and we're going to the next minor version of 0.13.0.

Our tests picked up a change in how `:=` schemas are encoded when using the JSON transformer - values that were untouched on 0.12.0 are  converted to strings on 0.13.0:

```clojure
(m/encode [:= 42] 42 mt/json-transformer)
;; 0.12.0 => 42
;; 0.13.0 => "42"
```

From [the changelog](https://github.com/metosin/malli/pull/951/files), it seems like maybe this behavior was introduced with the change described as "Use type inferrer when encoding enums" in #951. I tried it out on 0.18.0, and it looked like this behavior is still there - so it's been this way for a long time for users who are up-to-date.

Main question - is the way it works now intentional and the way it should work? It was surprising to me that the JSON transformer would take an integer value that matched the schema and is representable in JSON as a number and turn it into a string - the previous behavior of just doing nothing made more sense to me. Looking at `malli.transform/-infer-child-compiler`, I don't understand why it defaults to converting any value to a string, but I don't necessarily understand the philosophy of transformers necessarily.

mike706574:
Resolves #1204

Starting with 0.13.0, when encoding or decoding enum or equality schemas using the JSON transformer, values are converted to and from strings:

```clj
(m/encode [:= 42] 42 mt/json-transformer) ;; => "42"
(m/encode [:enum 42] 42 mt/json-transformer) ;; => "42"

(m/decode [:= 42] "42" mt/json-transformer) ;; => 42
(m/decode [:enum 42] "42" mt/json-transformer) ;; => 42
```

I think this behavior is expected from the string transformer, but not from the JSON transformer. Here's what I'd expect:

```clj
(m/encode [:= 42] 42 mt/json-transformer) ;; => 42
(m/encode [:enum 42] 42 mt/json-transformer) ;; => 42

(m/decode [:= 42] "42" mt/json-transformer) ;; => "42"
(m/decode [:enum 42] "42" mt/json-transformer) ;; => "42"
```

This change addresses that by inferring the encoder or decoder to use for an equality or enum schema from all of the other encoders or decoders rather than using some manual overrides.

I don't know much about the internals of Malli, like the protocols involved or the way encoder or decoder compilation works, so there might be a better way to do this - I'm willing to take a shot at implementing that if someone has some pointers.

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -251,44 +251,47 @@
 ;; decoders
 ;;
 
+(defn -base-json-decoders []
+  {'ident? -string->keyword
+   'simple-ident? -string->keyword
+   'qualified-ident? -string->keyword
+
+   'keyword? -string->keyword
+   'simple-keyword? -string->keyword
+   'qualified-keyword? -string->keyword
+
+   'symbol? -string->symbol
+   'simple-symbol? -string->symbol
+   'qualified-symbol? -string->symbol
+
+   'uuid? -string->uuid
+   'float? -number->float
+   'double? -number->double
+   'inst? -string->date
+   'integer? -number->long
+   'int? -number->long
+   'pos-int? -number->long
+   'neg-int? -number->long
+   'nat-int? -number->long
+   'zero? -number->long
+
+   #?@(:clj ['uri? -string->uri])
+
+   :float -number->float
+   :double -number->double
+   :int -number->long
+   :keyword -string->keyword
+   :symbol -string->symbol
+   :qualified-keyword -string->keyword
+   :qualified-symbol -string->symbol
+   :uuid -string->uuid
+   ;#?@(:clj [:uri -string->uri])
+
+   :set -sequential->set})
+
 (defn -json-decoders []
   (-add-child-compilers
-    {'ident? -string->keyword
-     'simple-ident? -string->keyword
-     'qualified-ident? -string->keyword
-
-     'keyword? -string->keyword
-     'simple-keyword? -string->keyword
-     'qualified-keyword? -string->keyword
-
-     'symbol? -string->symbol
-     'simple-symbol? -string->symbol
-     'qualified-symbol? -string->symbol
-
-     'uuid? -string->uuid
-     'float? -number->float
-     'double? -number->double
-     'inst? -string->date
-     'integer? -number->long
-     'int? -number->long
-     'pos-int? -number->long
-     'neg-int? -number->long
-     'nat-int? -number->long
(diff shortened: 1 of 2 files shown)


test/malli/transform_test.cljc
```diff
@@ -1125,41 +1125,79 @@
...
+                 :enum3 1
+                 :enum4 1.1
+                 :equals1 :kikka
+                 :equals2 'kikka
+                 :equals3 1
+                 :equals4 1.1}]
+    (testing "decoding children using the json transformer works"
+      (is (= decoded (m/decode child-inference-test-schema encoded (mt/json-transformer)))))
+    (testing "encoding children using the json transformer works"
+      (is (= encoded (m/encode child-inference-test-schema decoded (mt/json-transformer)))))))
```
opqdonut, line 1181: I'd like to see a case here for what happens when using the stringy `encoded` value with the json-transformer. The old test covered that.

src/malli/transform.cljc
```diff
@@ -233,147 +233,146 @@
...
-
-    :> -any->string
-    :>= -any->string
-    :< -any->string
-    :<= -any->string
-    :not= -any->string
-
-    'float -any->string
-    'double -any->string}))
+  (-add-child-compilers
```
opqdonut, line 353: Hmm, so `-add-child-compilers` will overwrite the `:=` and `:enum` keys from `-json-encoders`. I guess that works, but somehow I feel the composability of the encoders is decreased.

opqdonut (commented):
Some comments from my first skim. Looks good in general, but I can't pass judgement yet.

test/malli/transform_test.cljc
mike706574: Added that test + threw in some extra cases that assert that encoding or decoding values that are already encoded or decoded doesn't do anything.

src/malli/transform.cljc
mike706574: I pushed a change to pull out the "base" encoders and decoders into separate functions. It doesn't change anything functionally, but gets rid of the overwriting you mentioned, so it's at least more straightforward maybe? Let me know what you think.

opqdonut (approved):
I think this is good to merge now. If some regressions crop up, we'll deal with them. Thanks!

PR #1207, opened 2025-05-21
Replace predicate schemas with malli keyword schemas when possible

### LINKED ISSUE #1193: Avoid predicate schemas like `string?` in the README
We want to recommend using keyword schemas like `:string` instead.

mattiuusitalo:
As using malli schema types is recommended, we want to make sure that most of the documentation is using these in favor of predicate schemas.

Fixes #1193

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,14 +225,14 @@ Following example schema is assumed in many of the following examples.
 ```clojure
 (def Address
   [:map
-   [:id string?]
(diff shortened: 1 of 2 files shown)


mattiuusitalo:
Is there a recommended replacement for `number?` predicate schema?

opqdonut:
let's get this in to avoid building up a huge conflict, and worry about `number?` in a separate PR

Issue #1168, opened 2025-02-13
Cat and other seqexps do not honor gen/min and gen/max


slombardi-cni:
Pre-reqs for running the examples:
```clojure
(require '[malli.generator :as mg])
```
As part of our error messages, we are looking into generating concise but valid examples of schemas to help devs integrate with APIs or catch typos or other simple mistakes. Consider the following schema. It works as you would expect. Note we commented out the `[:+ int?]` at the bottom. Thus we get: 
```clojure
(let [RouteConfig [:sequential
                   {:gen/min 3, :gen/max 5}
                   [:and
                    [:cat {:gen/fmap vec}
                     [:string {:gen/elements ["/my/path" 
                                              "/my/awesome/path" 
                                              "/another/path"]}]
                     #_[:+ int?]]
                    vector?]]]
  (mg/generate RouteConfig))

;; [["/my/path"] ["/my/path"] ["/my/awesome/path"]]
```
Now let's uncomment the `[:+ int?]` and add some gen properties for it. The gen properties are ignored:
```clojure
(let [RouteConfig [:sequential
                   {:gen/min 3, :gen/max 5}
                   [:and
                    [:cat {:gen/fmap vec}
                     [:string {:gen/elements ["/my/path"
                                              "/my/awesome/path"
                                              "/another/path"]}]
                     [:+ {:gen/min 1, :gen/max 3} int?]]
                    vector?]]]
  (mg/generate RouteConfig))

;; [["/another/path" -7661 18084512 -15057394 991560 -13 229106415 -22]
;;  ["/my/awesome/path" 3228317 -26136 196 1003842 -305]
;;  ["/my/awesome/path" 592538 -2 1392 -76]
;;  ["/my/path" -161 -6360229 -18295415 3527 25 -28 438867 44066 -101382 -79 3 21 -1048678 41 -86440 -1701 2]]
```
If we move the props to the `cat` instead, it makes no difference:
```clojure
(let [RouteConfig [:sequential
                   {:gen/min 3, :gen/max 5}
                   [:and
                    [:cat {:gen/fmap vec :gen/min 1, :gen/max 3}
                     [:string {:gen/elements ["/my/path"
                                              "/my/awesome/path"
                                              "/another/path"]}]
                     [:+ int?]]
                    vector?]]]
  (mg/generate RouteConfig))

[...]

PR #1208, opened 2025-05-21
Gen min max for :+ and :* sequences

mattiuusitalo:
This is a (partial) fix for #1168

It makes sure that `:+` and `:*` use the :gen/min :gen/max etc options that are supported by other collection schemas.

I think it fixes the issue that the OG reporter had.

### DIFF
--- a/test/malli/generator_test.cljc
+++ b/test/malli/generator_test.cljc
@@ -315,7 +315,14 @@
       [:string {:gen/min 10, :max 20}]
 
       [:+ {:min 10 :max 20} :int]
+      [:+ {:gen/min 10 :max 20} :int]
+      [:+ {:min 10 :gen/max 20} :int]
+      [:+ {:gen/min 10 :gen/max 20} :int]
+
       [:* {:min 10 :max 20} :int]
+      [:* {:gen/min 10 :max 20} :int]
+      [:* {:min 10 :gen/max 20} :int]
+      [:* {:gen/min 10 :gen/max 20} :int]
 
       [:vector {:min 1, :gen/min 10, :max 100, :gen/max 20} int?]
       [:set {:min 1, :gen/min 10, :max 100, :gen/max 20} int?]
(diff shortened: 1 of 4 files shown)


test/malli/generator_test.cljc
```diff
@@ -314,10 +314,16 @@
       [:set {:min 10, :gen/max 20} int?]
       [:string {:gen/min 10, :max 20}]
 
+      [:+ {:min 10 :max 20} :int]
+      [:* {:min 10 :max 20} :int]
```
opqdonut, line 318: I'd like to see tests for gen/min and gen/max as well.

opqdonut (commented):
also, do you think we have enough docs so that this behaviour is discoverable? or should we add an example?

test/malli/generator_test.cljc
mattiuusitalo: Good idea, adding those

opqdonut:
don't forget to update the changelog!

ikitommi:
how do nested min & max work? e.g.

```clojure
[:cat {:gen/min 2 :gen/max 10} [:+ {:gen/min 4 :gen/max 5} :int] [:* {:gen/min 1 :gen/max 5} :string]]
```

opqdonut:
Do we have `:gen/min` support for `:cat` already? This PR doesn't add it AIUI.

ikitommi:
```clojure
[:* {:gen/min 2 :gen/max 10} [:cat [:+ {:gen/min 4 :gen/max 5} :int] [:* {:gen/min 1 :gen/max 5} :string]]]
```

if we don't have `:cat` gens.

opqdonut:
Somebody might be surprised that the min and max refer to the repetitions of the child schema, not to the size of the resulting sequence, but other than that I think this generalizes nicely.

opqdonut:
In that way, the semantics are the same as for regex:

```
[:* {:gen/min 2 :gen/max 10} [:cat [:+ {:gen/min 4 :gen/max 5} :int] [:* {:gen/min 1 :gen/max 5} :string]]]
```

corresponds to `(I{4,5}S{1,5}){2,10}`, where `I` and `S` are stand-ins for regexes for `:int` and `:string`

PR #1209, opened 2025-05-21
fix and document :default/fn

opqdonut:


mattiuusitalo (approved):

test/malli/transform_test.cljc
```diff
@@ -1068,10 +1068,13 @@
         (is (= {:first 'one, :second 'two} (m/encode schema {} transformer)))
         (is (= ['one 'two] @seen)))))
 
-  (testing ":default/fn property on schema"
-    (let [schema [:string {:default/fn (fn [] "called")}]]
-      (is (= "called" (m/decode schema nil mt/default-value-transformer)))))
-
+  (testing ":default/fn property"
```
mattiuusitalo, line 1071: Do we have a test for when the default shouldn't be called?

PR #1211, opened 2025-06-05
Reduce CLJS compiled bundle size

alexander-yakushev:
When the Clojurescript compiler encounters a call to `resolve` or a `var` macro, it inlines the whole metadata right at the callsite. This produces a lot of unnecessary code in the compiled output.

My proposal here is to rewrite `predicate-schemas` function to avoid using `#'` quoting. With these changes, I was able to reduce the compiled Malli size (with advanced optimizations) by 14 KB uncompressed (4KB gzipped).

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1,6 +1,6 @@
 (ns malli.core
   (:refer-clojure :exclude [eval type -deref deref -lookup -key assert])
-  #?(:cljs (:require-macros malli.core))
+  #?(:cljs (:require-macros malli.core malli.impl.util))
   (:require #?(:clj [clojure.walk :as walk])
             [clojure.core :as c]
             [malli.impl.regex :as re]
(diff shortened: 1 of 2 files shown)


ikitommi:
Change welcome.

alexander-yakushev:
Hm, I'm actually struggling to make it work on CLJS with all the macro action that is going on. Will you accept a PR that spells out all pairs of symbols and function values explicitly? I know it's more error prone this way, but I'm not sure I can make this work in the way I wanted from the beginning.

alexander-yakushev:
Alright, managed to get it to work by extracting the code-expanding macro into another namespace. I'm not sure why it couldn't work within the same namespace, I don't have that much experience with CLJS.

ikitommi:
Great, thanks! If you want smaller cljs bundles, there is also the compiler option to omit the default registry: https://github.com/metosin/malli?tab=readme-ov-file#dce-and-schemas

PR #1216, opened 2025-07-23
Move malli-types-clj(s) file to .clj-kondo/imports

### LINKED ISSUE #1215: Update path for clj-kondo imports
clj-kondo has changed where it stores config now to `.clj-kondo/imports/` rather than `.clj-kondo`.

I think this line should be updated:

https://github.com/metosin/malli/blob/6b0862135ea58ef2ac92bda9ae605994a5fb20d6/src/malli/clj_kondo.cljc#L169

See https://clojurians.slack.com/archives/CHY97NXE2/p1721810604723449?thread_ts=1721795642.361919&cid=CHY97NXE2

cc @borkdude in case I've missed something here.

danielcompton:
This is the new recommended path for clj-kondo config to live in.

Fixes #1215

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -166,8 +166,12 @@
      ([config key options]
       (let [cfg-file (apply io/file (conj
                                      (get options :clj-kondo-dir-path [])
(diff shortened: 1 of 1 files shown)


mattiuusitalo:
Thanks!

PR #1219, opened 2025-08-14
[perf] Improve malli.core/-intercepting peformance

alexander-yakushev:
Some minor space for improvement that I've spotted while working with Metabase. `-intercepting` just combines three functions via `comp`, but skips those that weren't provided. Instead of allocating a vector and transforming it, this PR spells out all cases explicitly. Slightly wordy, but no overhead.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -558,7 +558,9 @@
 
 (defn -intercepting
   ([interceptor] (-intercepting interceptor nil))
-  ([{:keys [enter leave]} f] (some->> [leave f enter] (keep identity) (seq) (apply -comp))))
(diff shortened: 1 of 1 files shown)


opqdonut:
Always good to avoid vectors. How about something a bit less verbose like this:

```
(defn comp-some [f g] (if (and f g) (-comp f g) (or f g)))

... (comp-some leave (comp-some f enter)) ...
```

alexander-yakushev:
Neat suggestion! I've updated the PR.

opqdonut (approved):

opqdonut:
Thanks for looking into performance!

PR #1220, opened 2025-08-14
[perf] Improve malli.transform/transformer peformance

alexander-yakushev:
The bulk of overhead in the returned `-value-transformer` implementation is performing `get-in` multiple times. get-in is quite terrible, so this PR proposes to manually unroll the keys that are looked up in the maps. I hope I preserved the semantics right, this function is a convoluted one.

### DIFF
--- a/src/malli/transform.cljc
+++ b/src/malli/transform.cljc
@@ -387,10 +387,9 @@
 (defn transformer [& ?transformers]
   (let [->data (fn [ts default name key] {:transformers ts
                                           :default default
-                                          :keys (when name
-                                                  (cond-> [[(keyword key) name]]
-                                                    (not (qualified-keyword? name))
-                                                    (conj [(keyword key (clojure.core/name name))])))})
+                                          :name name
+                                          :qname (when (and name (not (qualified-keyword? name)))
+                                                   (keyword key (clojure.core/name name)))})
         ->eval (fn [x options] (if (map? x) (reduce-kv (fn [x k v] (assoc x k (m/eval v options))) x x) (m/eval x)))
         ->chain (m/-comp m/-transformer-chain m/-into-transformer)
         chain (->> ?transformers (keep identity) (mapcat #(if (map? %) [%] (->chain %))) (vec))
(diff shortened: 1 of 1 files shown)


mattiuusitalo:
Apologies for taking a long time to review this. Did you measure the performance impact and what kind of numbers did you get?

alexander-yakushev:
Hi Matti! Let me expand why this is a useful improvement for us (and possibly for somebody else).

I uncovered this potential optimization while working on Metabase. In Metabase, `:ref` schemas are used a lot for various purposes, not just for recursive schemas, but as a general rule. When `:ref` is used, Malli schema compiler cannot employ some usual optimizations that it would do without it, and thus, the time spent in functions like `transformer` becomes runtime-relevant.

Here is a small reproducible:

```clj
(require '[malli.core :as m]
         '[malli.transform :as mt])

(def my-registry (merge (m/base-schemas)
                        {::some-number (m/-int-schema)}))

(def my-coercer
  (m/coercer [:ref ::some-number]
             (mt/transformer mt/default-value-transformer
                             {:name :custom-transformer})
             {:registry my-registry}))

(criterium.core/quick-bench (my-coercer 1))
```

```
Before patch:
Evaluation count : 1313220 in 6 samples of 218870 calls.
             Execution time mean : 475.031962 ns
    Execution time std-deviation : 24.682179 ns
   Execution time lower quantile : 447.572710 ns ( 2.5%)
   Execution time upper quantile : 499.586846 ns (97.5%)
                   Overhead used : 1.816199 ns

After patch:
Evaluation count : 2010882 in 6 samples of 335147 calls.
             Execution time mean : 303.423432 ns
    Execution time std-deviation : 9.061204 ns
   Execution time lower quantile : 295.271600 ns ( 2.5%)
   Execution time upper quantile : 314.545635 ns (97.5%)
                   Overhead used : 1.816199 ns
```

_Note that the after-results include both this PR and #1219._

The improvement is ~37% time reduction, but the bigger win is allocation: 1024b (before) vs 96b (after). Here's the diffgraph between original and patched version: http://flamebin.dev/GWRmDh. No wonder it is all blue because basically all allocations disappeared.

opqdonut:
I like it!

Issue #1153, opened 2024-12-29
Ambiguity with :catn parse results


ikitommi:
like #1150 but with `:catn`:

```clojure
(def Schema
  [:or
   [:vector [:map
             [:s :string]
             [:a :any]]]
   [:* [:catn
        [:s :string]
        [:a :any]]]])

(m/parse Schema ["kikka" {:a 1}])
; => [{:s "kikka", :a {:a 1}}]

(m/unparse Schema *1)
; => [{:s "kikka", :a {:a 1}}]
```

## Proposal

* have two parse result records
* corresponding functions (names do change, but we are breaking this anyway)
* maybe move to `malli.core` as this is kinda core functionality?

```clojure
;; any concatenated tagging, e.g. :catn
(defrecord Tags [])
;; for alternatios and ord, e.g. :orn, :altn
(defrecord Tag [key value])
```

ikitommi:
... and this also applies to `:cat`:


```clojure
(def Schema 
  [:or
   [:vector [:vector :any]]
   [:* [:cat :string :any]]])

(->> ["kikka" {:a 1}]
     (m/parse Schema)
     (m/unparse Schema))
; => [{:s "kikka", :a {:a 1}}]
```

opqdonut:
Using `{}` for the `:catn` parse forgets the ordering, which can be convenient for consumers. If instead of `{:s "k", :a {:a 1}}` we produce `(Tags. [(Tagged. :s "k") (Tagged. :a {:a 1})])`, the consumer will need to handle the ordering.

One option that comes to mind is to just use a generic `(defrecord Parsed [parsed])` wrapper to disambiguate parse results from values. Like this:

```
(m/parse [:orn ["any" :keyword]] :any)
;; ==> (Parsed. ["any" :keyword])
(m/parse [:* [:catn [:s :string] [:a :any]]] ["kikka" {:a 1}])
;; ==> [(Parsed. {:s "kikka" :a {:a 1}})]
(m/parse [:* [:cat :string :any]] ["kikka" 6])
;; ==> [(Parsed. ["kikka" 6])]
```

ikitommi:
I was thinking:

```clojure
(m/parse [:* [:catn [:s :string] [:a :any]]] ["kikka" {:a 1}])
;; ==> [#malli.util.impl.Parsed{:s "kikka" :a {:a 1}})]
```

e.g. calling `map->Tags`.

[...]

PR #1223, opened 2025-09-23
Provide a function to convert new parsing result to the old one

igrishaev:
Hi!

This pull request addresses an issue I faced today. In my project, we have a kind of a DSL what we parse using Malli, and then process. This DSL is quite flexible, and thus my schemas have a lot of `nor`, `catn` and similar parsing schemas. These schemas, when the data was ok, return a pair of `[tag, value]`.

When the incoming structure is parsed, I process it as follows:
```
(let [[tag value] parsed]
  (case tag
    :integer (process-as-integer value)
    :string (process-as-string value)
    :uuid (process-as-uuid value)))
```
The thing is, today somebody has bumped Malli, and now I receive not a vector but a map like `{:key <tag>, :value <value>}`. Personally I've found it being quite inconvenient as it breaks the entire parsing pipeline. Switching from a fixed vector to a map is really frustrating because it forces people to scan entire codebase and alter `case` forms.

This PR makes the Tag object `nth` friendly. When being de-structured, it acts like a normal vector: the 0 index returns tag, index 1 returns a value. Anything else, when `not-found` is not passed, throws an error. I provide tests for that.

It's my first pull request to Malli, so please let me know if anything needs correction.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1,7 +1,7 @@
 (ns malli.core
   (:refer-clojure :exclude [eval type -deref deref -lookup -key assert])
   #?(:cljs (:require-macros malli.core malli.impl.util))
-  (:require #?(:clj [clojure.walk :as walk])
+  (:require [clojure.walk :as walk]
             [clojure.core :as c]
             [malli.impl.regex :as re]
             [malli.impl.util :as miu]


igrishaev:
M-m..., any feedback? cc @ikitommi @frenchy64

opqdonut:
I think this is a good workaround for the breakage we caused with Malli, but am a bit hesitant to merge it until I get a second opinion. This might have unforeseen consequences.

opqdonut:
for the record, this was breakage in malli 0.18.0, via #1123 #1153

igrishaev:
Just a minute ago I found something else, let me check it quickly before merging this.

igrishaev:
Ok this PR addresses only one part of parsing breakage. I'll create a separate issue for another one.

igrishaev:
OK here it is: https://github.com/metosin/malli/issues/1224

opqdonut:
I'm wondering: could it be a cleaner approach to add `:plain-parse-results` to `options`? Then it could be a documented caveat that this output doesn't work with `unparse` (which needs the explicitness of `Tag`/`Tags`)

igrishaev:
Oh, it looks like Babashka doesn't support extending records with interfaces... grief.

opqdonut:
Concretely:

```
(def Multi
  [:multi {:dispatch :type}
   [:user [:map [:size :int]]]
   [::m/default :any]])

(m/parse Multi {:type :user, :size 1})
;; => #malli.core.Tag{:key :user, :value {:type :user, :size 1}}
(m/parse Multi {:type :user, :size 1} {:plain-parse-output true})
;; => [:user {:type :user, :size 1}]
```

igrishaev:
It could have been a new function like  `parse-new`, or renaming `parse` to `parse-old`. Because the new parse implementation totally breaks the pipeline: not only in vector->map direction but also introducing an extra level like `:values`.

opqdonut:
I'll try to clarify how I see the situation:

The explicitness of Tag / Tags was needed to make sure parse->unparse roundtrips even for exotic schemas. This makes malli much more reliable when people are doing complicated things, but it does make life more difficult for people who just want to use the output of `parse`. Adding a new function or an option to get the old output of `parse` feels like a good compromise.

igrishaev:
OK, it looks like extending the Tag object with an interface works only in pure Java (I'm not too familiar with clojurescript and babashka). I think I'll close this unless I find a better solution.

igrishaev:
OK, so the following code has fixed issues I was suffering from:

```clojure
(defn remap-tags [parsed]
  (walk/postwalk
   (fn [x]
     (cond
       (malli/tag? x) [(:key x) (:value x)]
       (malli/tags? x) (:values x)
       :else x))
   parsed))
```

Is it possible to include it to malli, what do you think? If yes, I'll update the PR.

opqdonut:
Yes, for sure, that's a nice way to address the incompatibility. I propose `old-parse-format` as the name.

igrishaev:
Ok I've updated the PR: the Tag object stays untouched now, I only added a function to covert the parsing result and a test.

opqdonut:
Looks like the `clojure.walk` require in `malli.core` is clj-only. Easy to fix.

igrishaev:
Right, I have fixed that. I ran tests locally but only for native Clojure, sorry.

igrishaev:
Thank you!

PR #1234, opened 2025-10-30
Correctly handle recursive schemas in -parser-info

frenchy64:
Fixes https://github.com/metosin/oksa/issues/27

Detects ref cycles by identifying refs by their name+scope. Generalizes a concept from malli.generator, so a shared helper `m/-identify-ref-schema` seemed appropriate since it's quite handy. It could be used in place of `-memoize` in `:ref`'s validator. Unfortunately there's no `opts` arg so dynamic vars might be needed.

GPT-4.1 LLM used for debugging, design, and to generate the basis for the for the `:ref` `-parser-info` case (based on `-ref-gen`).


mattiuusitalo (approved):
Good work

PR #1235, opened 2025-10-30
Sketch: tying the knot for ref validator

frenchy64:
This way we can fully compile a ref validator (faster & less memory, don't need to stop to compile each new level and cache).

Removing the dynamic variable here is a strong motivator for `-validator` to take `opts`. OTOH, we're gaining perf and only need to use dynamic variable once before its cached.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -2114,6 +2114,8 @@
                      (fn [x]
                        (if-let [f @vol]
                          (do (println "LAZY: TIED THE KNOT")
+                             ;;TODO exercise code path
+                             (throw (ex-info "LAZY: TIED THE KNOT" {}))
                              (f x))
                          (let [f (binding [*ref-validators* (assoc ref-validators id vol)]
                                    (-validator (rf)))]


src/malli/core.cljc
```diff
@@ -1926,12 +1926,151 @@
...
+;; lazy refs only realize their child when the validator is called
+(-memoize (fn [] (schema (mr/-schema (-registry options) ref) options)))
+;;                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+;; eager refs realize their child when compiling their validators
+(when-let [s (mr/-schema (-registry options) ref)] (-memoize (fn [] (schema s options))))
+;;           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+;;
+;; Because of this essential difference, validators for eager refs can be fully compiled, but
+;; lazy refs must  <START FROM HERE, WE CAN FIX THIS>
```
opqdonut: Looks like the excellent docs trail off here. So what is the difference in the plan eager vs. lazy?

src/malli/core.cljc
```diff
@@ -1926,12 +1926,151 @@
...
+      #(@vol %)
+      (let [vol (volatile! nil)
+            s (or (when-let [s (mr/-schema (-registry options) ref)]
+                    (schema s options))
+                  (when-not allow-invalid-refs
+                    (-fail! ::invalid-ref {:type :ref, :ref ref})))
+            f (binding [*ref-validators* (assoc ref-validators id vol)]
+                (-validator s))]
+        (vreset! vol f)
+        f))))
```
opqdonut: This impl looks good, thanks! I wonder if the `s` logic could be replaced with something like `deref`?

src/malli/core.cljc
```diff
@@ -1947,10 +2086,12 @@
...
-       (let [rf (or (and lazy (-memoize (fn [] (schema (mr/-schema (-registry options) ref) options))))
-                    (when-let [s (mr/-schema (-registry options) ref)] (-memoize (fn [] (schema s options))))
-                    (when-not allow-invalid-refs
-                      (-fail! ::invalid-ref {:type :ref, :ref ref})))
+       (let [?schema (-memoize
+                       #(or (mr/-schema (-registry options) ref)
+                            (when-not allow-invalid-refs
+                              (-fail! ::invalid-ref {:type :ref, :ref ref}))))
+             _ (when-not lazy (?schema))
+             rf (-memoize #(schema (?schema) options))
```
opqdonut: why this change?

src/malli/core.cljc
```diff
@@ -1961,9 +2102,34 @@
...
+           (-validator [this]
+             (let [ref-validators *ref-validators*
+                   id (-identify-ref-schema this)]
+               (if lazy
+                 (let [rv (ref-validators id)
+                       vol (or rv (volatile! nil))]
+                   (c/assert (not @vol) "invariant: we tie the knot on the way back up")
+                   (if-some [f @vol]
+                     @f
+                     (fn [x]
```
opqdonut: Ok, so in the lazy case, the knot-tying is delayed until the validator is run. That makes sense. Could the code for the lazy and eager cases be made to look even more symmetric, so that the added `fn` wrapper stands out as the only difference?

src/malli/core.cljc
```diff
@@ -1926,12 +1926,151 @@
...
+;; ## Eager refs
+;; 
+;; Implementation:
+
+(-validator [this]
+  (let [ref-validators *ref-validators*
+        id (-identify-ref-schema this)]
+    (if-some [vol (ref-validators id)]
+      #(@vol %)
+      (let [vol (volatile! nil)
```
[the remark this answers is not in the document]
frenchy64: For posterity it's more like:
- volatile can race if multiple threads realize lazy refs and generate validators for them simultaneously
- synchronization logic adds unnecessary overhead since all racing validators are equivalent
- "extra" validators that are written to the volatile first but then overwritten will be garbage collected because we always get the latest validator from the volatile, so over time there will be one canonical validator for recursion

frenchy64:
Plumatic Schema [solves this](https://github.com/plumatic/schema/blob/74a1ecdb645d3f43b405a27e2b70f72db861b7ca/src/cljc/schema/spec/core.cljc#L74-L86) by caching validators for _all_ schemas during compilation. In addition to handling recursive schemas, it also prevents a nasty exponential blowup of compilation size for even non-recursive schemas, that Malli also suffers from.

Malli's validator compilation is exponential. This registry demonstrates how:

```clojure
(def registry {::creates-1-validator [:tuple]
               ::creates-2-validators [:tuple ::creates-1-validator ::creates-1-validator ::creates-1-validator ::creates-1-validator]
               ::creates-16-validators [:tuple ::creates-2-validators ::creates-2-validators ::creates-2-validators ::creates-2-validators]
               ::creates-64-validators [:tuple ::creates-16-validators ::creates-16-validators ::creates-16-validators ::creates-16-validators]
               ::creates-256-validators [:tuple ::creates-64-validators ::creates-64-validators ::creates-64-validators ::creates-64-validators]
               ::creates-1024-validators [:tuple ::creates-256-validators ::creates-256-validators ::creates-256-validators ::creates-256-validators]
               ::creates-4096-validators [:tuple ::creates-1024-validators ::creates-1024-validators ::creates-1024-validators ::creates-1024-validators]
               ::creates-16384-validators [:tuple ::creates-4096-validators ::creates-4096-validators ::creates-4096-validators ::creates-4096-validators]
               ::creates-65536-validators [:tuple ::creates-16384-validators ::creates-16384-validators ::creates-16384-validators ::creates-16384-validators]
               ::creates-262144-validators [:tuple ::creates-65536-validators ::creates-65536-validators ::creates-65536-validators ::creates-65536-validators]
               ::creates-1048576-validators [:tuple ::creates-262144-validators ::creates-262144-validators ::creates-262144-validators ::creates-262144-validators]
               ::creates-4194304-validators [:tuple ::creates-1048576-validators ::creates-1048576-validators ::creates-1048576-validators ::creates-1048576-validators]})
```

With this registry, each level of depth N compiles `(m/validator ::creates-1-validator)` 4^N times.

e.g., `(m/validator ::creates-4194304-validators)` compiles `(m/validator ::creates-1-validator)` 4,194,304 (4^11) times.

Plumatic Schema would only compile it once. It's not so trivial to achieve with dynamically scoped refs, but it's the same idea as detecting ref cycles, which we can now do reliably.

Here's a reproduction of the issue https://github.com/frenchy64/malli/pull/36/files which I have been pondering since discussing https://github.com/metosin/malli/pull/1180

I mention this because, like Plumatic Schema, I think it makes sense to have the same solution solve both exponential and recursive validators.

mattiuusitalo:
Good illustration of the problem. This seems worth pursuing.

frenchy64:
Extracted minimal recursive validator implementation in a separate PR: https://github.com/metosin/malli/pull/1245

I will work on the exponential problem separately, the solutions ended up being different to the recursive validator impl.

PR #1239, opened 2025-11-10
[perf] Optimize -parent-child-transformer

alexander-yakushev:
Computing a child-transformer in a single `reduce` achieves several benefits:
1. No need to accumulate intermediate child transformers into a vector.
2. No need to reverse the vector direction - calling `-comp` on the next child and the accumulator achieves that naturally.
3. For the very common case where there is just one child, the vector overhead is gone completely.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -603,11 +603,10 @@
   (let [parent-transformer (-value-transformer transformer parent method options)
         child-transformer
         (reduce (fn [acc child]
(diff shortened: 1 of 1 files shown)


alexander-yakushev:
@frenchy64 Yeah, that was my initial version, but then I decided to codegolf one line off. I've commited your suggestion.

frenchy64 (approved):

PR #1242, opened 2025-11-15
[perf] Add dissoc-map-keys to optimize -map-schema

alexander-yakushev:
The current approach of `(reduce (fn [acc k] (dissoc acc k)) m (keys keyset))` has two inefficiencies:
- `keys` is an expensive function, having to allocate a sequence of keys 
- Reducing the result of `keys` also has allocation overhead because such reduction is Iterator-based.

This PR addresses the efficiency part and also introduces a reusable function because this pattern appears multiple times in `-map-schema`.

This is optimized as part of the higher-level benchmark but I can craft a lower-level benchmark if necessary.


opqdonut (changes requested):
Otherwise good to go! No benchmark needed, `reduce-kv` tends to be faster in pretty much all cases.

alexander-yakushev:
Thanks for the review! Addressed the comment.

Issue #975, opened 2023-11-30
Custom validation functions: Allow programmatically defining `:error/path`


mkreis:
Open this feature request as discussed here: https://clojurians.slack.com/archives/CLDK6MFMK/p1701249325700779

It would be great if it was possible to programmatically define `:error/path` depending on the validation function (and the data to be validated).

In below example from the docs `:error/path` only accept vectors is hard-coded to `[:password2]`:
```
(-> [:and [:map
           [:password string?]
           [:password2 string?]]
     [:fn {:error/message "passwords don't match"
           :error/path [:password2]}
       (fn [{:keys [password password2]}]
         (= password password2))]]
    (m/explain {:password "secret"
                :password2 "faarao"})
    (me/humanize))
; {:password2 ["passwords don't match"]}
```

irigarae:
Another idea about this, in case it moves this issue a little bit for others.

Without knowing too much about the internals of `me/humanize` if it did "flatmap" on the errors, it would be possible to return multiple errors (and paths) from the same `:error/fn`. Here's an example on what the update to me/humanize could look like from the current version:

```diff
 (defn humanize
   "Humanized a explanation. Accepts the following options:
 
   - `:wrap`, a function of `error -> message`, defaulting to `:message`
   - `:resolve`, a function of `explanation error options -> path message`"
   ([explanation]
    (humanize explanation nil))
   ([{:keys [value errors] :as explanation} {:keys [wrap resolve]
                                             :or {wrap :message
                                                  resolve me/-resolve-direct-error}
                                             :as options}]
    (when errors
      (->> errors
-          (map (fn [error]
-                    (let [[path message] (resolve explanation error options)]
-                      (assoc error ::path path ::message message))))
+          (mapcat (fn [error]
+                    (let [[base-path message] (resolve explanation error options)]
+                      (if (vector? message)
+                        (for [{:error/keys [message path]} message]
+                          (assoc error ::path (into base-path path) ::message message))
+                        [(assoc error ::path base-path ::message message)]))))
           (reduce
            (fn [acc error]
              (let [{::keys [path message]} error]
                (me/-push-in acc value path (wrap (assoc error :message message)))))
            nil)))))
```

With this idea, it would allow these three to return exactly the same errors:

```clj
[:fn {:error/path [:k1 :k2]
      :error/message "foo"}
 (constantly false)]
[:fn {:error/path [:k1 :k2]
      :error/fn (fn [{:keys [value]} _]
                  "foo")}
[...]

PR #1243, opened 2025-11-22
Allow error-path to be a function

mkreis:
Implements issue #975 to programmatically define :error/path depending on the validation function (and the data to be validated).

### DIFF
--- a/README.md
+++ b/README.md
@@ -957,6 +957,31 @@ Errors can be targeted using `:error/path` property:
 ;; => {:password2 ["passwords don't match"]}
 ```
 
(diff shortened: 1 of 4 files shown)


opqdonut:
Thanks for the excellent PR! Code-wise, there's nothing to complain about, it has docs, tests, and is implemented minimally. 

However, we have some reservations about this feature in general. Concretely, we haven't seen a use-case that couldn't be expressed using other malli features already. Also, there's a mismatch between the existing `:error/fn` behaviour and `:error/path`. Please see my comments in the issue for more:

https://github.com/metosin/malli/issues/975#issuecomment-3588236180

opqdonut:
trying out #1255 instead

PR #1244, opened 2025-11-25
Document mutable registry caveats

frenchy64:
My overall goal here to better understand ref cycle detection in the presence of mutable registries, in particular what happens if you mutate a registry to change the value of a ref _during_ ref cycle detection.

This also is a response to some of the feedback from https://github.com/metosin/malli/pull/1180, in particular the concerns that this optimization would interfere with mutable/dynamic registries. Here's my attempt to describe the real-world caveats of mutable registries, and how malli itself aggressively caches the results of mutable registries in ways that might be surprising in the name of performance.

In particular, this attempts to refute/address [@ikitommi's scenario](https://github.com/metosin/malli/pull/1180#issuecomment-2742472509):

> I would assume that I can recreate a schema validator with current bindings when calling `-validator` on it, no caching.

I hope this PR describes how this is not a realistic expectation, since Schema's themselves cache registry lookups.

What I found (and tried to communicate here) is that Malli itself intrinsically does not support mutating registries _during_ schema traversals/construction. The key example is the "hello world" / "hei maailma" example, that, despite an atomic transaction updating the schemas to be consistent within the registry, it's possible to have a single schema contain "hello maailma" (see `transactions can still lead to inconsistent schemas` test).

From this, I hope that we can establish an expectation that users cannot rely on registry mutations to be reliably picked up by parsed Schema's, and they must reparse them from syntax to update their schemas reliably.

This expectation makes the original question of ref cycle detection easier to answer: ref cycle detection doesn't work if registries mutate underneath them, because malli itself cannot handle these things correctly, and doing so is user error.

### DIFF
--- a/test/malli/core_test.cljc
+++ b/test/malli/core_test.cljc
@@ -3666,28 +3666,18 @@
       (testing "registry mutation ignored due to cached schema"
         (is (= :int (-> schema m/deref-all m/form))))))
   (testing "transactions can still lead to inconsistent schemas"
-    (let [loading-order (atom [])
-          registry-atom (atom (assoc (m/default-schemas)
+    (let [registry-atom (atom (assoc (m/default-schemas)
                                      "hello" [:enum "hello"]
                                      "world" [:enum "world"]))
           reg (mr/lazy-registry
                 (m/default-schemas)
                 (fn [type registry]
-                  (when-some [?schema (@registry-atom type)]
-                    (swap! loading-order conj [type :=> ?schema])
-                    (m/schema ?schema {:registry registry}))))
-          schema (m/schema [:tuple
-                            "hello"
-                            [:multi {:lazy-refs true, :dispatch identity}
-                             "world"]]
+                  (some-> (@registry-atom type)
+                          (m/schema {:registry registry}))))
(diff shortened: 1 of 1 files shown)


test/malli/core_test.cljc
```diff
@@ -3637,3 +3637,57 @@
...
+      (swap! registry-atom assoc
+             "hello" [:enum "hei"]
+             "world" [:enum "maailma"])
+      (testing "permanently inconsistent schema, despite CAS transaction"
+        (is (= ["hello" "maailma"]
+               [(-> schema m/children first m/form)
+                (-> schema m/children second m/children first last m/deref m/form last)]))
+        (is (= [["hello" :=> [:enum "hello"]]
+                ["world" :=> [:enum "maailma"]]]
+               @loading-order))))))
```
opqdonut: I would like a simpler test case for the inconsistency, but maybe this was difficult to come up with already! Not insisting.

frenchy64: The only difficultly is crafting these examples without reaching into malli's implementation details. I updated the example to be easier to follow by using `m/-lazy` directly.

PR #1249, opened 2026-01-04
Introduce more efficient -memoize on JVM

alexander-yakushev:
While investigating Malli memory footprint in various cases, I spotted a small opportunity for improving `-memoize`. Currently, `-memoize` returns a closure over an AtomicReference and the thunk to be called. This is what it takes in heap, realized, and not:

```clj
;; Current implementation
(require '[clj-memory-meter.core :as mm])

(mm/measure (-memoize #(do true))) ;; "56 B"

(mm/measure (let [m (-memoize #(do true))]
              (m)
              m)) ;; "72 B"
```

By replacing a closure with a custom type that holds the result in a mutable volatile field (thus, avoiding separate AtomicReference), and more importantly, cleans up the thunk once it was computed, we get this:

```clj
;; After patch
(mm/measure (-memoize #(do true))) ;; "40 B"

(mm/measure (let [m (-memoize #(do true))]
              (m)
              m)) ;; "40 B"
```

The bytes improvement alone is probably not that impressive, however, Malli quite likes to spam `-memoize`, and I've seen millions of memoize objects in a single heap, so we talk about potentially reducing footprint by a few dozen megabytes. But the fact that thunks are cleared means the GC can free those, and that it is potentially more rewarding (given how thunks may contain quite heavy objects like parts of schemas etc. etc.).

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -265,9 +265,25 @@
 
 (defn -vmap ([os] (miu/-vmap identity os)) ([f os] (miu/-vmap f os)))
 
+;; Very primitive memoization/deferred value implementation that doesn't use
+;; locks (thus, allows multiple callers to race) but clears the thunk after
+;; it was computed once (to allow GC to clean it).
+#?(:clj
+   (deftype SimpleMemoize [^:volatile-mutable f ^:volatile-mutable value]
+     clojure.lang.IFn
+     (invoke [_]
+       (let [f' f] ; capture volatile `f` to avoid racy overwrites
+         (if (nil? f') ; `f=nil` means somebody has already computed value
+           value
+           (let [value' (f')]
+             (do (set! value value') ; set `value` first, then clean up `f`
+                 (set! f nil) ; JMM guarantees order here since fields are volatile
+                 value')))))))
+
 (defn -memoize [f]
-  (let [value #?(:clj (AtomicReference. nil), :cljs (atom nil))]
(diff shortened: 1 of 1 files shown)


frenchy64:
Possible cross-platform implementation: https://github.com/metosin/malli/pull/1252/changes

```clojure
(defn -memoize [f]
  (let [f (volatile! f)
        value (volatile! nil)]
    (fn [] (or @value
               (if-some [f' @f]
                 (let [v (f')]
                   (vreset! value v)
                   (vreset! f nil) ;; must set after value
                   v)
                 @value)))))
```

Hopefully we can cut the total amount of memoization by avoiding creating schemas altogether via knot-tying, sharing etc.

alexander-yakushev:
Yeah, this would work. It would have a marginally larger footprint (72B vs 40B), but I'm just counting peanuts here, really, it doesn't matter this much. Too bad that Babashka doesn't have its own conditional reader, though.

frenchy64:
Yeah. Presumably the typical memoized thunk is heavier than `#(do true)`, which I took as the main optimization here (garbage collecting entire schemas).

frenchy64:
> Too bad that Babashka doesn't have its own conditional reader, though.

It does (`:bb`), you just need to use it before `:clj`.

FWIW the cross platform version should also improve memory usage for cljs too.

alexander-yakushev:
Gotcha. Well, it's probably for the best to just keep a single cross-platform version indeed than to mess with conditionals. I defer to your solution.

PR #1252, opened 2026-01-08
Cross-platform -memoize optimization

frenchy64:
Closes https://github.com/metosin/malli/pull/1249

Alternative to https://github.com/metosin/malli/pull/1249


opqdonut:
I like it in general!

alexander-yakushev:
LGTM too.

PR #1254, opened 2026-01-09
WIP: Explain knot

frenchy64:
Avoids recreating child schemas for recursive explainers.

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -1940,17 +1940,18 @@
 ;; returns an identifier for the :ref schema in the context of its dynamic scope.
 ;; useful for detecting cycles.
 ;; copied to malli.generator
-(defn- -identify-ref-schema [schema]
-  ;; TODO mr/-schemas doesn't seem right, making defn private for now.
-  ;; e.g., we only care about property registry entries, not schema constructors.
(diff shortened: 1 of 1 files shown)


opqdonut:
do you want an initial review of this already or is it still under progress?

frenchy64:
@opqdonut an initial review would be good.

The summary is: this attempts to reuse objects in memory as much as possible without changing the arguments to `-explain` (moving the `path` to the return function). We can't reuse the result of `-explain` (because it's closed over by a different `path`, moving the `path` param would fix this), but we can reuse the schema itself at recursive points.

Open question: is this schema reuse a way to remove `*ref-validators*`? Perhaps by caching the result of `-validator` for refs? Is that too much caching? Alternative is to add `options` arg to `-validator`.

test/malli/core_test.cljc
```diff
@@ -3681,3 +3681,19 @@
              "world" [:enum "maailma"])
       (is (= ["hello" "maailma"]
              (mapv (comp peek m/form m/deref-all) (m/children schema)))))))
+
+(deftest recursive-explainer-test
```
opqdonut: thanks for the very concrete test

Issue #407, opened 2021-03-31
Unjustified stack overflow with recursive schema


helins:
The following scheme systematically produces a stack overflow. It is a recursive definition where one item (`:E`) contains another item (which might be `:E` as well, or not).

```clojure
(def registry
     {:A    [:tuple [:= :A]]
      :B    [:tuple [:= :B]]
      :C    [:tuple [:= :C]]
      :D    [:tuple [:= :D]]
      :E    [:tuple
             [:= :E]
             :item]
      :item [:multi
             {:dispatch first}
             [:A :A]
             [:B :B]
             [:C :C]
             [:D :D]
             [:E :E]]})

;; Validation or generation both produce a stack overflow

(malli/validate [:schema
                 {:registry registry}
                 :item]
                [:B])

(malli.gen/generate [:schema
                     {:registry registry}
                     :item])
```

Stack overflow:
```
Execution error (StackOverflowError) at malli.core/-schema (core.cljc:242).                                                                                                                                                                                                                                                                                                               
```

Same happens when using `:or` instead of `:multi`.

helins:
After proper usage of `:ref` in the context of recursion, closed in favor of issue #408 .

PR #1259, opened 2026-02-28
Fix for nbb (babashka ClojureScript) compatibility

robert-johansson:
This PR makes malli work with [nbb](https://github.com/babashka/nbb) — a
ClojureScript scripting environment powered by SCI (Small Clojure Interpreter).
Malli 0.17.0 works with nbb, but 0.18.0+ broke due to two issues. This PR
fixes both with minimal changes (2 files, ~20 lines), with zero behavioral
change for standard Clojure/ClojureScript users.

Depends on [babashka/nbb#407](https://github.com/babashka/nbb/issues/407)
(exposing `IPrintWithWriter` in nbb), which has been fixed on nbb master.

### Issue 1: `predicate-schemas*` JVM-only macro (introduced in 0.19.0)

The `predicate-schemas*` macro in `malli.impl.util` was wrapped in
`#?(:clj ...)`. In standard ClojureScript, the CLJS compiler runs this macro
on the JVM at compile time. In nbb/SCI (self-hosted ClojureScript), there is
no separate JVM compile step, so the macro was unavailable.

**Fix** (`impl/util.cljc`): Remove the `#?(:clj ...)` guard from the `defmacro`. The macro
body is portable Clojure (just `for` + quasiquoting, no JVM interop), so it
works in both JVM and self-hosted ClojureScript. This follows the same pattern
as `malli.core/assert`, which is also an unguarded `defmacro` in a `.cljc`
file. The existing `(:require-macros malli.impl.util)` in `core.cljc` ensures
it is loaded as a macro in ClojureScript contexts.

### Issue 2: `borkdude.dynaload` not deref-able under nbb

The `:default` evaluator path in `malli.sci` uses `dynaload` to load SCI
functions, then derefs them. In nbb, `dynaload` loads but the result cannot
be deref'd (`IDeref` protocol error), so the evaluator fails at runtime.

**Fix** (`sci.cljc`): Add `:org.babashka/nbb` reader conditional to skip the
`dynaload` require and use `load-string` directly for the evaluator (same
approach as `:bb`).

## Changes

- `src/malli/impl/util.cljc` — remove `#?(:clj)` guard from `predicate-schemas*` macro
- `src/malli/sci.cljc` — conditional require and nbb evaluator

**No changes to `core.cljc`.** The `IPrintWithWriter` issue that previously
required 46 reader conditionals is now resolved upstream in
[babashka/nbb#407](https://github.com/babashka/nbb/issues/407).

All existing tests pass.

## Testing

Verified with nbb dev build (includes nbb#407 fix) on macOS, both Node.js and bun:

```
$ node ~/nbb/lib/nbb_main.js -e '
(require (quote [malli.core :as m]))
(prn (m/validate [:map [:x int?] [:y int?]] {:x 1 :y 2}))     ;; true
(prn (m/validate [:map [:x int?] [:y int?]] {:x 1 :y "bad"}))  ;; false
(prn (m/validate [:or int? string?] "hello"))                   ;; true
(prn (m/parse [:* [:catn [:p string?] [:v [:altn [:s string?] [:b boolean?]]]]]
              ["-x" "foo" "-v" true]))
'
```

---
*This PR was authored with [Claude Code](https://claude.com/claude-code)*

### DIFF
--- a/src/malli/impl/util.cljc
+++ b/src/malli/impl/util.cljc
@@ -72,8 +72,7 @@
   #?(:clj  (-pred-composer or 16)
      :cljs (fn [preds] (fn [x] (boolean (some #(% x) preds))))))
 
-#?(:clj
-   (defmacro predicate-schemas* [var-syms]
-     `(-> {}
-          ~@(for [vsym var-syms]
-              `(malli.core/-register-var '~vsym ~vsym)))))
+(defmacro predicate-schemas* [var-syms]
+  `(-> {}
+       ~@(for [vsym var-syms]
+           `(malli.core/-register-var '~vsym ~vsym))))
--- a/src/malli/sci.cljc
+++ b/src/malli/sci.cljc
@@ -1,10 +1,16 @@
 (ns malli.sci
-  (:require [borkdude.dynaload :as dynaload]))
+  (:require #?@(:org.babashka/nbb []
+                :default [[borkdude.dynaload :as dynaload]])))
 
 (defn evaluator [options fail!]
-  #?(:bb      (fn []
-                (fn [form]
-                  (load-string (str "(ns user (:require [malli.core :as m]))\n" form))))
+  #?(:org.babashka/nbb
+     (fn []
+       (fn [form]
+         (load-string (str "(ns user (:require [malli.core :as m]))\n" form))))
+     :bb
+     (fn []
+       (fn [form]
+         (load-string (str "(ns user (:require [malli.core :as m]))\n" form))))
      :default (let [eval-string* (dynaload/dynaload 'sci.core/eval-string* {:default nil})
                     init (dynaload/dynaload 'sci.core/init {:default nil})
                     fork (dynaload/dynaload 'sci.core/fork {:default nil})]


opqdonut:
Thanks for this! In order to merge this we need to decide if nbb becomes an officially supported platform in malli. This means additional friction for changes since they now need to work on one more platform.

There's also some infrastructure work needed like setting up CI jobs etc., but I can handle that easily.

borkdude:
> Thanks for this! In order to merge this we need to decide if nbb becomes an officially supported platform in malli.

Note that nbb used to work already :)

opqdonut:
Yes, but we had no tests for it, thus the breakage!

robert-johansson:
Thanks for the review @opqdonut, and for the pointer @borkdude!

I've reworked the PR based on @opqdonut's suggestion to make the `predicate-schemas*` macro available to self-hosted ClojureScript rather than inlining the expansion.

### What changed in this update

**`predicate-schemas*` macro** (the big simplification): Instead of the ~40-line inline expansion under `:org.babashka/nbb`, I simply removed the `#?(:clj ...)` guard from the `defmacro` in `impl/util.cljc`. The macro body is portable (just `for` + quasiquoting, no JVM interop), so it works in self-hosted ClojureScript too. This follows the same pattern as `malli.core/assert` which is also an unguarded `defmacro` in a `.cljc` file. The existing `(:require-macros malli.impl.util)` in `core.cljc` handles loading it as a macro. This means **zero changes to `core.cljc` for the macro issue** — much cleaner.

**`IPrintWithWriter`** and **`sci.cljc`** changes are the same as before — reader conditionals to handle SCI not exposing the print protocol, and conditional dynaload/evaluator for nbb.

The branch is now rebased on current upstream master (0.20.1). All existing tests pass.

opqdonut:
Are you running the test suite with nbb? It seems to break for me:

```
% npx nbb --classpath test test/bb_test_runner.clj
----- Error --------------------------------------
Message:  Unable to resolve symbol: cljs.core/regexp?
Location: /home/joel/work/metosin/malli/.nbb/.cache/3114d59c1db3bac33b63442b64552ec4e4ae048a/nbb-deps/fipp/util.cljc:13:1
Phase:    analysis

----- Context ------------------------------------
9:            false)
10:    :clj (def char? clojure.core/char?))
11:
12: ;;TODO: CLJ-1720 and CLJS-1242
13: #?(:clj (defn regexp? [x]
    ^--- Unable to resolve symbol: cljs.core/regexp?
14:           (instance? java.util.regex.Pattern x))
15:    :cljs (def regexp? cljs.core/regexp?))
16:
17: (defn edn?
18:   "Is the root of x an edn type?"

----- Stack trace --------------------------------
fipp.util - /home/joel/work/metosin/malli/.nbb/.cache/3114d59c1db3bac33b63442b64552ec4e4ae048a/nbb-deps/fipp/util.cljc:13:1

file:///home/joel/.npm/_npx/f77abc9b469a86df/node_modules/nbb/lib/nbb_main.js:13
$APP.kD),$APP.FY);return k.v?k.v():k.call(null)}):null).then(function(k){$APP.h($APP.h(d)?k!=null:d)&&$APP.Zy.l($APP.z([k],0));return k}).catch(function(k){$APP.jha(k,b);$APP.h($APP.$y.g(b))&&console.error(k);throw Error($APP.pm(k));})})):console.error("Usage: "+$APP.J.g("nbb")+" \x3cscript\x3e or nbb -e \x3cexpr\x3e.")})();
                                                                                                                                                                                                                       ^

Error: Unable to resolve symbol: cljs.core/regexp?
    at file:///home/joel/.npm/_npx/f77abc9b469a86df/node_modules/nbb/lib/nbb_main.js:13:216

Node.js v20.19.2
```

```
% cat nbb.edn
{:deps {metosin/malli {:local/root "."}}}
```

opqdonut:
Trying to run just `malli.core-test` fails as well. Of course, it might be that I'm doing this completely wrong.

```
% npx nbb --classpath test -e "(require 'malli.core-test 'clojure.test) (clojure.test/run-tests 'malli.core-test)"
----- Error --------------------------------------
Message:  Could not find namespace: goog.math.Long
file:///home/joel/.npm/_npx/f77abc9b469a86df/node_modules/nbb/lib/nbb_main.js:13
$APP.kD),$APP.FY);return k.v?k.v():k.call(null)}):null).then(function(k){$APP.h($APP.h(d)?k!=null:d)&&$APP.Zy.l($APP.z([k],0));return k}).catch(function(k){$APP.jha(k,b);$APP.h($APP.$y.g(b))&&console.error(k);throw Error($APP.pm(k));})})):console.error("Usage: "+$APP.J.g("nbb")+" \x3cscript\x3e or nbb -e \x3cexpr\x3e.")})();
                                                                                                                                                                                                                       ^

Error: Could not find namespace: goog.math.Long
    at file:///home/joel/.npm/_npx/f77abc9b469a86df/node_modules/nbb/lib/nbb_main.js:13:216

Node.js v20.19.2
```

robert-johansson:
You're right — my testing was done interactively with `nbb -e` against core functionality, not the full test suite. The failures you're seeing are from transitive dependencies that aren't nbb-compatible:

- `fipp` uses `cljs.core/regexp?` which SCI doesn't expose — this comes from `malli.dev`/pretty-printing and `malli.clj-kondo`
- `goog.math.Long` (Google Closure) isn't available in nbb — likely pulled in via `malli.generator`/`test.check`

Note that some core malli namespaces also use Google Closure (`malli.transform` uses `goog.date`, `malli.registry` uses `goog-define`), so nbb compatibility may be limited to a subset of namespaces even beyond the test suite.

I'm happy to contribute whatever testing approach you'd prefer. Some options:

1. **A dedicated nbb smoke test** — a small script that exercises the nbb-compatible subset and can run in CI via `npx nbb`
2. **An nbb test runner** that selectively requires only nbb-compatible test namespaces
3. **Something else entirely** — you mentioned being able to handle CI infrastructure, so let me know what works best for the project

What level of nbb test coverage would you want before considering this for merge?

borkdude:
I'm not sure if the tests ran with 0.17 either so just some smoke testing seems best to me at this point. 
Also welcome to post nbb issues to improve nbb where possible (using isolated small repros preferably).

opqdonut:
I think that in order to accept changes for nbb compatibility, we need to decide what nbb compatibility means. Also, it needs to be easy for other contributors to know that their changes don't break nbb. Otherwise the support will get broken again very quickly in some other change.

Ideally, I'd want to have the same test suite pass for bb and nbb to avoid having to define yet another test scenario. Ideally, I would also want to minimize nbb-specific code.

One more thing: if a (potentially no-op) IPrintWriter were to be added to nbb, much of this diff on the malli side would disappear.

robert-johansson:
What do you say @borkdude? Is IPrintWriter in nbb an option? Maybe that's a way to start this?

borkdude:
@robert-johansson Can you make an isolated repro on nbb? I can see what I can do, I'd need to dig into this issue before I can say more.

robert-johansson:
@borkdude   

```
  nbb -e '
  (reify
    IPrintWithWriter
    (-pr-writer [this writer opts]
      (-write writer "#MyType{}")))
  '
```

borkdude:
Is this about reify specifically? Can you make an nbb github issue?

robert-johansson:
Updated the PR based on @borkdude's fix in babashka/nbb#407 (exposing `IPrintWithWriter`).

The diff is now **2 files, ~20 lines** — no changes to `core.cljc` at all:

1. **`impl/util.cljc`** — remove `#?(:clj)` from `predicate-schemas*` macro (portable body, same pattern as `malli.core/assert`)
2. **`sci.cljc`** — skip `dynaload` under nbb, use `load-string` (dynaload loads but can't be deref'd in nbb)

Tested with nbb dev build (nbb master with #407) on both Node.js and bun — validate, explain, parse, transform all working.

opqdonut:
Ok, well, I'm happy to merge this now. The decision to Officially Support NBB is still not made, but this is a small enough diff.

PR #1275, opened 2026-04-09
Close #1271: handle humanized values that don't match custom path

### LINKED ISSUE #1271: me/humanize throws on incorrect input
Hello there,

I have a more involved use-case, but something like this is happening when users are passing a vector on a multi that has an `:error/path` comprised of keywords.

This is the smallest reproducible example that I could come up with:
```clj
(-> [:fn {:error/path [:foo]}
     :foo]
    (ma/explain [])
    (me/humanize))
;; => Execution error (IllegalArgumentException) at malli.error/-push (malli.error.cljc:191).
;;    Key must be integer
```

I am really not sure what the solution is in this case, but perhaps `me/humanize` could return assume that the container is of an invalid type if this happens and return something like:
```clj
["invalid type"]
```

Thank you for your help and for the fantastic library!

frenchy64:
Close #1271

### DIFF
--- a/src/malli/error.cljc
+++ b/src/malli/error.cljc
@@ -187,10 +187,15 @@
 (defn -fill [x i fill] (-concat x (repeat (- i (count x)) fill)))
 
 (defn -push [x k v fill]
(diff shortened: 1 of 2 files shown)


opqdonut (approved):
thanks!

PR #1277, opened 2026-04-13
clj-kondo: map bytes? to :array

### LINKED ISSUE #1276: clj-kondo type for `bytes?` should be :array instead of :char-sequence
When running this code:

```clojure
(require '[malli.experimental :as mx])
(import [java.io ByteArrayOutputStream])

(mx/defn out-bytes :- bytes?
  []
  (.toByteArray (ByteArrayOutputStream.)))

(alength (out-bytes))
```

I get a clj-kondo warning over `(alength (out-bytes))` saying: "Expected: array, received: char sequence. clj-kondo(type-mismatch)"

So I believe `bytes?` should be transformed into the clj-kondo type :array instead of :char-sequence

Knotschi:
Emit `:array` for `bytes?` so JVM `byte[]` matches array APIs like `alength` and clj-kondo stops reporting “char sequence” vs “array” for Malli-annotated code.

Fixes #1276

### DIFF
--- a/src/malli/clj_kondo.cljc
+++ b/src/malli/clj_kondo.cljc
@@ -55,7 +55,7 @@
 (defmethod accept 'associative? [_ _ _ _] :associative)
 (defmethod accept 'sequential? [_ _ _ _] :sequential)
 (defmethod accept 'ratio? [_ _ _ _] :int) ;;??
-(defmethod accept 'bytes? [_ _ _ _] :char-sequence) ;;??
+(defmethod accept 'bytes? [_ _ _ _] :array)
 (defmethod accept 'ifn? [_ _ _ _] :ifn)
 (defmethod accept 'fn? [_ _ _ _] :fn)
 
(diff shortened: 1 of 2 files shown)


src/malli/clj_kondo.cljc
```diff
@@ -55,7 +55,7 @@
 (defmethod accept 'associative? [_ _ _ _] :associative)
 (defmethod accept 'sequential? [_ _ _ _] :sequential)
 (defmethod accept 'ratio? [_ _ _ _] :int) ;;??
-(defmethod accept 'bytes? [_ _ _ _] :char-sequence) ;;??
```
opqdonut, line 58: please remove the `;;??` now that we know what this should be

also, please add a test in clj_kondo_test.cljc. Perhaps adding one more child to Schema is the easiest.

opqdonut (changes requested):
Thanks for the contribution! I'd like a couple of fixes though.

Knotschi:
I removed the `;; ??`

When writing the test I realized that `bytes?` does not exist on cljs, so I wrapped the `accept 'bytes?` method into a `#?(:clj )`
For this reason I could also not just adding it to `Schema` in `clj-kondo-test` and instead added a seperate test for it.

All tests were green for me (`./bin/kaocha` and `./bin/node`)

src/malli/clj_kondo.cljc
Knotschi: done

opqdonut (approved):
thanks! good catch re: cljs and bytes?

PR #1280, opened 2026-04-23
Close #1278: correctly extract function arities before cljs instrument

### LINKED ISSUE #1278: Adding a `:registry` map to a `:function` schema
I tried adding a registry map to a function schema like this:
```clj
(def foo:schema
  [:function
   {:registry {:foo any?}}
   [:=> [:cat :foo any?] any?]
   [:=> [:cat :foo :foo any?] any?]])
```
As per the documentation [here](https://github.com/metosin/malli/blob/52ea58a36ff5172b38dfc526ca638afa7226a4a0/docs/function-schemas.md?plain=1#L256), but that throws a `:malli.core/invalid-schema` error.

This seems to work however:
```clj
(def foo:schema
  [:function
   [:=> {:registry {:foo any?}} [:cat :foo any?] any?]
   [:=> {:registry {:foo any?}} [:cat :foo :foo any?] any?]])
```

ambrosebs:


opqdonut:
Code LGTM. A test case would be nice.

opqdonut (approved):

PR #1284, opened 2026-05-01
Share pointer children

frenchy64:
Schemas like `[:schema {:registry {::FOO :bar}} ::FOO]` now create `:bar` once (before, twice). This is regardless of how many times `::FOO` is mentioned in the registry or body

For example it also holds for

```clojure
[:schema {:registry {::BAZ ::FOO ::FOO ::BAR ::BAR :bar}} [:tuple ::BAZ ::BAZ ::BAZ]]
```

which would previously create 3 copies of `:bar` via `-property-registry`, then 3 additional copies via `[:tuple ::BAZ ::BAZ ::BAZ]` for a total of 6 copies (now 1).

The situation is more subtle for nested registries, see the tests.

This tackles exponential duplication of schemas. Previously, `(m/schema [:schema {:registry exponential-schema}} ::creates-4194304-validators])` would create at least 4,194,304 copies of `::counting`, now only 1. I'm not sure of the exact number because of OOM errors, but it could be closer to 20 million.

```clojure
(def exponential-registry
  {::creates-1-validator ::counting
   ::creates-2-validators [:tuple ::creates-1-validator ::creates-1-validator ::creates-1-validator ::creates-1-validator]
   ::creates-16-validators [:tuple ::creates-2-validators ::creates-2-validators ::creates-2-validators ::creates-2-validators]
   ::creates-64-validators [:tuple ::creates-16-validators ::creates-16-validators ::creates-16-validators ::creates-16-validators]
   ::creates-256-validators [:tuple ::creates-64-validators ::creates-64-validators ::creates-64-validators ::creates-64-validators]
   ::creates-1024-validators [:tuple ::creates-256-validators ::creates-256-validators ::creates-256-validators ::creates-256-validators]
   ::creates-4096-validators [:tuple ::creates-1024-validators ::creates-1024-validators ::creates-1024-validators ::creates-1024-validators]
   ::creates-16384-validators [:tuple ::creates-4096-validators ::creates-4096-validators ::creates-4096-validators ::creates-4096-validators]
   ::creates-65536-validators [:tuple ::creates-16384-validators ::creates-16384-validators ::creates-16384-validators ::creates-16384-validators]
   ::creates-262144-validators [:tuple ::creates-65536-validators ::creates-65536-validators ::creates-65536-validators ::creates-65536-validators]
   ::creates-1048576-validators [:tuple ::creates-262144-validators ::creates-262144-validators ::creates-262144-validators ::creates-262144-validators]
   ::creates-4194304-validators [:tuple ::creates-1048576-validators ::creates-1048576-validators ::creates-1048576-validators ::creates-1048576-validators]})
```

Note: `(m/validator [:schema {:registry exponential-schema}} ::creates-4194304-validators])` still calls `m/-validator` 4,194,304 times, as noted in the `TODO` at the bottom of `malli.core-test` in this PR. This can be tackled separately.

Metabase works around this a similar issue [here](https://github.com/metabase/metabase/blob/64d2e589b2ccca746ff52a924ecfc04ed83b9b08/src/metabase/util/malli/registry.cljc#L126-L169) by creating their own `:ref` implementation. We don't handle `:ref` here yet.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   build-clj:
+    timeout-minutes: 5
     strategy:
       matrix:
         # Supported Java versions: LTS releases and latest
@@ -44,6 +45,7 @@ jobs:
 
   build-cljs:
     name: ClojureScript
+    timeout-minutes: 5
     strategy:
       matrix:
         mode: [none, advanced, cherry-none, cherry-advanced]
@@ -80,6 +82,7 @@ jobs:
   build-bb:
     name: Babashka
 
+    timeout-minutes: 5
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -111,6 +114,7 @@ jobs:
     # the examples are still valid.
     name: Doc Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     steps:
       - uses: actions/checkout@v6
       - name: Setup Java 25
--- a/README.md
+++ b/README.md
@@ -3250,8 +3250,10 @@ Note that a similar result happens when a local registry overlaps with a custom/
 
 This optimization currently only shares results amongst the transitive children
 of a schema, not between different top-level schemas. For the previous schema,
-that means a new shared schemas is created for `::list-of` when `m/schema` is
-called again, even though everything including `registry` is the same as the previous call.
+that means a new shared schema is created for `::list-of` when the `m/schema` is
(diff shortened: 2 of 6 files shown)


frenchy64:
After some reflection, I think the essence of this change is an integrated mechanism to detect recursive schemas. We endlessly work around this missing feature in order to detect cycles in each operation. If malli intrinsically handled recursive schemas to always be equal, then these cycle checks become simply equality checks.

This needs to be a user-level expectation, with documentation explaining the scenarios where the knot is tied at the schema level.

PR #1285, opened 2026-05-03
Add :decimal schema

### LINKED ISSUE #1272: Full support for decimal
Hello :)
Found myself having to make some stylistic compromises with decimals in my schema, and would have liked it if it was supported like a `:double` schema with min max preds and bound generation.

Alternatives:
- use `decimal?` and custom generators / and `:and` for bounds
- define my own schema

Thanks

josephinoo:
Introduce support for a `:decimal` primitive type across the library.
- Adds `-decimal-schema` (predicate: `decimal?` on CLJ, `false` on CLJS) and registers `:decimal` in the builtin type map.
- Implements a CLJ-only generator mapping double values to `BigDecimal` via `gen-fmap` + `bigdec`. 
- Adds CLJ-only transformer entries for string↔decimal and any→string conversion.
- Expands core tests with validation/encode/decode/ast expectations for `:decimal`.
- Adds a CLJ-only generator test asserting produced values are decimals within bounds.
- Uses reader conditionals to keep CLJS behavior safe throughout.

Closes #1272


src/malli/core.cljc
```diff
@@ -824,6 +824,7 @@
 (defn -int-schema [] (-simple-schema {:type :int, :pred int?, :property-pred (-min-max-pred nil)}))
 (defn -float-schema [] (-simple-schema {:type :float, :pred float?, :property-pred (-min-max-pred nil)}))
 (defn -double-schema [] (-simple-schema {:type :double, :pred double?, :property-pred (-min-max-pred nil)}))
+(defn -decimal-schema [] (-simple-schema {:type :decimal, :pred #?(:clj decimal? :cljs (constantly false)), :property-pred (-min-max-pred nil)}))
```
frenchy64: If the schema is meaningless in cljs then I propose `:decimal` not exist on that platform.

src/malli/generator.cljc
```diff
@@ -418,6 +418,7 @@
 (defmethod -schema-generator :int [schema options] (gen/large-integer* (-min-max schema options)))
 (defmethod -schema-generator :double [schema options] (double-gen schema options))
 (defmethod -schema-generator :float [schema options] (double-gen schema options))
+#?(:clj (defmethod -schema-generator :decimal [schema options] (gen-fmap bigdec (double-gen schema options))))
```
frenchy64: I don't think the current implementation works with `:gen/infinite true` and `:gen/NaN? true`.

It might make sense to never generate those values as in spec's generator:

https://github.com/clojure/spec.alpha/blob/6b698eed1787c14907082c86de3410dd51e816b4/src/main/clojure/clojure/spec/gen/alpha.clj#L158-L159

test/malli/generator_test.cljc
```diff
@@ -345,6 +345,11 @@
       [:set {:min 1, :gen/min 10, :max 100, :gen/max 200} int?]
       [:string {:min 1, :gen/min 10, :max 100, :gen/max 200}])))
 
+#?(:clj
+   (deftest decimal-generator-test
+     (is (every? decimal? (mg/sample [:decimal {:min 0.0M, :max 1.0M}] {:size 100})))
+     (is (every? #(<= 0.0M % 1.0M) (mg/sample [:decimal {:min 0.0M, :max 1.0M}] {:size 100})))))
```
frenchy64: Please add some tests for `:gen/min` and `:gen/max` (see existing for other schemas), and some tests providing `:min`/`:max` as longs/doubles as in `malli.generator-test/double-with-long-min-test`.

Also tests that provide bounds outside the range of Double, for example `{:min (inc (bigdec Double/MAX_VALUE))}` should generate numbers greater than `(bigdec Double/MAX_VALUE)`.

frenchy64: Actually I'm not sure if anyone has released a BigDecimal generator that supports arbitrary bounds.

This is the closest I could find:

https://github.com/clojure/test.check/blob/5ba3a25b60cf66ff531db8f44d89145abfacfa5c/src/main/clojure/clojure/test/check/generators.cljc#L1389

I think at minimum we need a test to show that it throws rather than violating the expected bounds.

PR #1288, opened 2026-05-10
Add support for JSON Schema If-Then-Else via :if

### LINKED ISSUE #1287: JSON Schema If-Then-Else conditional construct support
JSON Schema supports a recursive construct to _"allow the application of a subschema based on the outcome of another schema"_. In simpler terms, one can define `if` conditions to apply `then` schema if it matches, or `else` schema if it doesn't.

Official documentation is available here: [json-schema.org / Conditional schema validation: If-Then-Else](https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse)

The need for custom schema type `:if` comes from JSON Schema generation, as otherwise it would be quite hard to know what to pick for the matching JSON fields as values and what the intent is. As such this is what is being suggested - a new core schema which behaves like `(if conditional then else)` code block and can produce JSON Schema If-Then-Else conditionals.

Purely from a syntax point of view, what can be supported as a structure is
```
[:if [:map [:x [:= true]]]  ; condition
     [:map [:y int?]]       ; then
     [:map [:y string?]]]   ; else
```
The `else` branch may be omitted. JSON Schema itself would allow for omitting `then`, but expressing that without custom non-Malli-like data structure would be quite hard, as it's easier to handle this as a vector with 2 or 3 elements.

Since Malli's evaluation doesn't short circuit (as it shouldn't!), the equivalence between suggested schema and Malli's native core schema needs to take this into account. Effectively the suggested schema above would be equivalent to
```
[:or [:and [:map [:x [:= true]]]
           [:map [:y int?]]]
     [:and [:not [:map [:x [:= true]]]]
           [:map [:y string?]]]]
```
or as a bit of pseudocode
```
[:or [:and condition then]
     [:and [:not condition] else]]
```

---
This issue is based on discussion at [Clojurians Slack](https://clojurians.slack.com/archives/CLDK6MFMK/p1777987825383029).

esuomi:
Adds support for JSON Schema's conditional If-Then-Else structure by implementing a new schema type :if to enable detection and handling of the logic when transforming given data to JSON Schema.

This new :if schema type also has equivalence using Malli's existing core schemas (:and/:or/:not) which is internally used for transform and generator support.

Fixes #1287

### DIFF
--- a/src/malli/core.cljc
+++ b/src/malli/core.cljc
@@ -3023,6 +3023,102 @@
                                   :re-transformer (fn [_ children] (apply re/alt-transformer children))
                                   :re-min-max (fn [_ children] (reduce -re-alt-min-max {:max 0} (-vmap last children)))})})
 
+(defn -if-conditional-schema
+  "Malli schema for JSON Schema's If-Then-Else conditional support.
+
+   Usage:
+   ```
+   [::jse/if [:map [:x [:= true]]]  ; IF
+             [:map [:y int?]]       ; THEN
+             [:map [:y string?]]]   ; ELSE
+  ```
+  Validates the same as:
+  ```
+  [:or [:and [:map [:x [:= true]]]         ; IF
+             [:map [:y int?]]]             ; THEN
+       [:and [:not [:map [:x [:= true]]]]  ; NOT IF
+             [:map [:y string?]]]]         ; ELSE
+  ```
+
+  The `ELSE` branch can be omitted for `when`-like behavior."
+  []
+  ^{:type ::into-schema}
+  (reify IntoSchema
+    (-type [_] :if)
+    (-type-properties [_])
+    (-properties-schema [_ _])
+    (-children-schema [_ _])
+    (-into-schema [parent {:keys [tags] :as properties} children options]
+      (-check-children! ::if properties children 2 3)
+      (let [condition' (schema (nth children 0) options)
+            then' (schema (nth children 1) options)
+            else' (when (= 3 (count children))
+                    (schema (nth children 2) options))
+            condition-v (validator condition')
+            equivalent-schema' (schema (if else'
+                                         [:or
+                                          [:and condition' then']
+                                          [:and [:not condition'] else']]
+                                         [:and condition' then'])
+                                       options)]
+        ^{:type ::schema}
+        (reify
+          Schema
+          (-validator [this]
+            (let [then-v (validator then')
+                  else-v (when else' (validator else'))]
+              (fn [value]
+                (if (condition-v value)
+                  (then-v value)
+                  (if else-v (else-v value) true)))))
+          (-explainer [this path]
+            (let [then-explainer (-explainer then' (conj path 'then))
+                  else-explainer (when else' (-explainer else' (conj path 'else)))]
+              (fn explain [x in acc]
+                (if (condition-v x)
+                  (let [before (count acc)
+                        acc' (then-explainer x in acc)]
+                    (if (> (count acc') before)
+                      (into (subvec (vec acc') 0 before)
+                            (map #(assoc % :branch 'then))
+                            (subvec (vec acc') before))
+                      acc'))
+                  (if else-explainer
+                    (let [before (count acc)
+                          acc' (else-explainer x in acc)]
+                      (if (> (count acc') before)
+                        (into (subvec (vec acc') 0 before)
+                              (map #(assoc % :branch 'else))
+                              (subvec (vec acc') before))
+                        acc'))
+                    acc)))))
+          (-parser [_]
+            (let [then-p (-parser then')
+                  else-p (when else' (-parser else'))]
+              (fn [x]
+                (if (condition-v x)
+                  (then-p x)
+                  (if else-p (else-p x) x)))))
+          (-unparser [_]
+            (let [then-u (-unparser then')
+                  else-u (when else' (-unparser else'))]
+              (fn [x]
+                (if (condition-v x)
+                  (then-u x)
+                  (if else-u (else-u x) x)))))
+          (-transformer [this transformer method options]
+            (-transformer equivalent-schema' transformer method options))
+          (-walk [this walker path options]
+            (-walk-leaf this walker path options))
+          (-properties [this] properties)
+          (-options [this] options)
+          (-children [this] children)
+          (-parent [this] parent)
+          (-form [this] form))))))
+
+(defn conditional-schemas []
+  {:if (-if-conditional-schema)})
+
 (defn base-schemas []
   {:and (-and-schema)
    :andn (-andn-schema)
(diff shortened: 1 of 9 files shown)


src/malli/core.cljc
```diff
@@ -3023,6 +3023,102 @@
...
+          Schema
+          (-validator [this]
+            (let [then-v (validator then')
+                  else-v (when else' (validator else'))]
+              (fn [value]
+                (if (condition-v value)
+                  (then-v value)
+                  (if else-v (else-v value) true)))))
+          (-explainer [this path]
+            (let [then-explainer (-explainer then' (conj path 'then))
```
frenchy64, line 3075: The `-explainer` seems to do extra work. It's enough to simply extend the path here instead of then associng `:branch` later. And I suggest using keywords as the path elements and implementing LensSchema.

src/malli/core.cljc
```diff
@@ -3023,6 +3023,102 @@
...
+                  (if else-u (else-u x) x)))))
+          (-transformer [this transformer method options]
+            (-transformer equivalent-schema' transformer method options))
+          (-walk [this walker path options]
+            (-walk-leaf this walker path options))
+          (-properties [this] properties)
+          (-options [this] options)
+          (-children [this] children)
+          (-parent [this] parent)
+          (-form [this] form))))))
```
frenchy64, line 3117: Here's the correct implementation here for ParserInfo.

```clojure
ParserInfo
(-parser-info [_ opts]
  {:simple-parser (and (:simple-parser (-parser-info then' opts))
                       (if else' (:simple-parser (-parser-info else' opts))  true))})

```

There are some other useful methods missed that are implemented in the surrounding schemas.

EDIT: actually this depends on which implementation of `-parser` we use so let's defer this discussion until that is resolved.

src/malli/core.cljc
```diff
@@ -3023,6 +3023,102 @@
...
+                  else-p (when else' (-parser else'))]
+              (fn [x]
+                (if (condition-v x)
+                  (then-p x)
+                  (if else-p (else-p x) x)))))
+          (-unparser [_]
+            (let [then-u (-unparser then')
+                  else-u (when else' (-unparser else'))]
+              (fn [x]
+                (if (condition-v x)
```
frenchy64, line 3106: I don't think this is right. `x` here is potentially parsed so it's meaningless to test the condition validator. For example, consider `[:if :int [:orn [:int :int]]]`. It's now a tag at this point, so will always fail the `:int` validator.

I think there's a few options to choose here:

1. we return a tagged result from `-parser` similar to `:orn`
2. we ensure `:if` never transforms the parsed value by using `-simple-parser`
3. we try each unparser in turn similar to `:or`

src/malli/core.cljc
```diff
@@ -3023,6 +3023,102 @@
...
+                  (then-u x)
+                  (if else-u (else-u x) x)))))
+          (-transformer [this transformer method options]
+            (-transformer equivalent-schema' transformer method options))
+          (-walk [this walker path options]
+            (-walk-leaf this walker path options))
```
frenchy64, line 3112: `:if` is not a leaf.

src/malli/core.cljc
```diff
@@ -3023,6 +3023,102 @@
...
+  ```
+  Validates the same as:
+  ```
+  [:or [:and [:map [:x [:= true]]]         ; IF
+             [:map [:y int?]]]             ; THEN
+       [:and [:not [:map [:x [:= true]]]]  ; NOT IF
+             [:map [:y string?]]]]         ; ELSE
+  ```
+
+  The `ELSE` branch can be omitted for `when`-like behavior."
```
frenchy64, line 3043: It's more like logical implication as written. This is the opposite of what I expected. For example `[:or [:if S T] [:if S' T']]` would never hit the second schema.

Additionally, the equivalent schema to `[:if S T]` is actually `[:or [:and S T] :any]`, so there's some internal inconsistency here.

This might need more time to bake, I propose we support 2 args later and start with 3. I understand JSON Schema maps missing branches to true but this doesn't map nicely to Clojure's semantics.

test/malli/transform_test.cljc
```diff
@@ -1229,3 +1229,15 @@
...
+(deftest if-conditional-transformations-test
+  (let [schema [:if
+                [:map [:animal [:= :cat]]]
+                [:map [:says [:= :meow]]]
+                [:map [:says [:= :bark]]]]
+        encoded {:animal "cat"
+                 :says   "meow"}
+        decoded {:animal :cat
+                 :says   :meow}]
+    (testing "roundtip works"
```
opqdonut, line 1242: [typo] roundtRip

test/malli/transform_test.cljc
```diff
@@ -1229,3 +1229,15 @@
...
+      (is (= encoded (m/encode schema decoded (mt/string-transformer)))))))
```
opqdonut, line 1243: I was expecting this to also test that `(= decoded (m/decode ...))`

test/malli/json_schema_test.cljc
```diff
@@ -394,3 +394,97 @@
...
+                               [:map ["card-number" :string]]
+                               [:map ["iban" :string]]]]
+      (is (= {:if   {:properties {"type" {:const "credit"}}
+                     :required   ["type"]}
+              :then {:properties {"card-number" {:type "string"}}
+                     :required
+                     ["card-number"]}
+              :else {:properties {"iban" {:type "string"}}
+                     :required   ["iban"]}}
+             (json-schema/transform if-then-else-schema)))))
```
opqdonut, line 411: tests should use `transform-and-check`, which uses AJV to validate the schema

test/malli/core_test.cljc
```diff
@@ -3697,3 +3697,32 @@
     (is (= @count-into-schemas 3)) ;; was 6
     (is (m/coerce ConsCell [1 [2 [3 [4 [1 [2 [3 [4 nil]]]]]]]]))
     (is (= @count-into-schemas 3)))) ;; was 10
+
+(deftest if-conditional-schema-test
```
opqdonut, line 3701: good test, though it doesn't exercise the 2-arity form at all!

test/malli/json_schema_test.cljc
```diff
@@ -394,3 +394,97 @@
...
+  (testing "can override type omission"
+    (let [type-emitting-schema [:if {:omit-condition-type false}
+                                [:map ["planet" [:= "Earth"]]]
+                                [:map ["habitable" [:= true]]
+                                 ["population" integer?]]
+                                [:map ["habitable" [:= false]]]]
+          result (json-schema/transform type-emitting-schema)]
+      (is (= "object" (get-in result [:if :type])))
+      (is (= "object" (get-in result [:then :type])))
+      (is (= "object" (get-in result [:else :type]))))))
```
opqdonut, line 490: I'd like this test to show what happens when something other than `:map` is used inside `:if`

opqdonut (changes requested):
Ambrose reviewed the impl so I just had a look at the tests & docs.

PR #1289, opened 2026-05-12
Consistent seq behaviour in :validate

irigarae:
With the current behaviour, returning an empty list of errors gives `false` in `m/validate` but `nil` in `m/explain`. This PR brings consistency.

And separate (off) topic, but the test also shows that `:type` is returned as `nil` in the explain `:errors`, even if not provided. Not sure if this is consistent with other parts of the codebase where it's often not present, or if it was intentional to make it mandatory in this schema.


opqdonut (approved):
thanks for catching this! sorry for the delay 😅

PR #1295, opened 2026-06-28
[perf] Optimize m.instrument/clj-collect!

alexander-yakushev:
This is minor, but `filter-key` is so inefficient that it got my attention. On a project like Metabase that has a few thousand namespaces, calling `(clj-collect! {:ns (all-ns)})` generates a lot of unnecessary garbage and slightly slows down the initialization.

```
Before:
Time per call: 534.50 ms   Alloc per call: 318,043,024b

After:
Time per call: 44.75 ms   Alloc per call: 13,792,652b
```

### DIFF
--- a/src/malli/instrument.clj
+++ b/src/malli/instrument.clj
@@ -52,7 +52,18 @@
 (defn clj-collect!
   ([] (clj-collect! {:ns *ns*}))
   ([{:keys [ns]}]
-   (not-empty (reduce (fn [acc v] (let [v (-collect! v)] (cond-> acc v (conj v)))) #{} (vals (mapcat ns-publics (-sequential ns)))))))
+   (not-empty
+    (reduce (fn [acc nspace] ns-publics
+              (let [nspace (the-ns nspace)]
+                (reduce-kv (fn [acc _ ^clojure.lang.Var v]
+                             (if-let [v (and (instance? clojure.lang.Var v)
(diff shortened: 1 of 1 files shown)


src/malli/instrument.clj
```diff
@@ -52,7 +52,18 @@
 (defn clj-collect!
   ([] (clj-collect! {:ns *ns*}))
   ([{:keys [ns]}]
-   (not-empty (reduce (fn [acc v] (let [v (-collect! v)] (cond-> acc v (conj v)))) #{} (vals (mapcat ns-publics (-sequential ns)))))))
+   (not-empty
+    (reduce (fn [acc nspace] ns-publics
```
frenchy64:
```suggestion
    (reduce (fn [acc nspace]
```

alexander-yakushev: Thanks! Stray garbage got in 🤦.

frenchy64:
Is there any hope of factoring out a helper to improve readability while preserving perf? I'm thinking `-ns-publics` that returns an eduction of public vars, that we then reduce or transducer over.

alexander-yakushev:
I've just tried it, but some of the efficiency is gone this way because IKVReduce doesn't play along with eductions and such, and the transducing in general. You would need to iterate over MapEntries in a map, and that is an overheady thing on its own.

PR #1297, opened 2026-07-09
Close #1296: don't depend on sets throwing on 2 args

### LINKED ISSUE #1296: Clojure 1.13 possible compatibility issue
Hey, our Clojure regression suite saw this failure in the malli tests with Clojure 1.13.0-alpha3:

```
FAIL in malli.core-test/function-schema-test (core_test.cljc:2551)
:=> using generative testing
expected: (false? (m/validate schema2 #{}))
  actual: (not (false? true))

FAIL in malli.core-test/function-schema-test (core_test.cljc:2603)
:-> using generative testing
expected: (false? (m/validate schema2 #{}))
  actual: (not (false? true))
250 tests, 18772 assertions, 2 failures.
```

I did not dig into it, but I would strongly suspect some relationship to https://clojure.atlassian.net/browse/CLJ-2958 (which makes sets implement ILookup). Please let me know if you look at it more and learn anything that would be useful to us, thanks!

frenchy64:
Close #1296 

Now that sets accept two arguments in Clojure 1.13.0-alpha3 these tests will fail.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         # Supported Java versions: LTS releases and latest
-        jdk: [8, 11, 17, 21, 25]
-        clojure: [11, 12]
+        jdk: [8, 11, 17, 21, 25, 26]
+        clojure: [11, 12, 13]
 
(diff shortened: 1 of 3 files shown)


opqdonut (approved):
thanks!

Issue #1311, opened 2026-09-03
Feedback on :validate


NoahTheDuke:
# Problem

`:validate` only allows for passing in custom errors, not errors resulting from other `m/explain` calls. This makes it hard/impossible to use in conjunction with `m/explain` calls (conditional validation, dynamic dispatch, etc).

# Example

Given this `:validate` schema, `m/explain` will return explain data with the wrong `:schema`:

```clojure
(def Example
  (m/schema [:validate (fn [value]
                         (when-let [ex (m/explain :int value)]
                           (:errors ex)))]
            {:registry (merge (m/default-schemas) (mev/schemas))}))

(m/explain :int :a)
;=> {:schema :int, :value :a, :errors ({:path [], :in [], :schema :int, :value :a})}

(m/explain Example :a)
;=> {:schema #object[validate$reify...],
;    :value :a,
;    :errors ({:path [], :in [], :schema #object[validate$reify...], :value :a, :type nil})}
```

# Source of problem

It seems that the `-explainer` function in the validate schema overwrites every given error's schema with the `:validate` schema:

https://github.com/metosin/malli/blob/e0240dc1fd1cb2a63e0a04528eb4e82b68ce8c8c/src/malli/experimental/validate.cljc#L33

This is most evident when looking at the output from `me/humanize`:

```clojure
(me/humanize (m/explain :int :a))
;=> ["should be an integer"]

(me/humanize (m/explain Example :a))
;=> ["unknown error"]
```

# Potential solution

One way to solve this is to use an existing `:schema` on an error entry if it exists, or the `:validate` schema otherwise. That will flow "normal" errors through and allow for custom errors to use `:type` as currently documented.

To demonstrate, I changed line 33 to `(into acc (map (fn [{:keys [schema in type value]}] (miu/-error path (into in0 in) (or schema this) value type))) errors)`, and now the "correct" output is returned:

```clojure
(me/humanize (m/explain Example :a))
;=> ["should be an integer"]
```

opqdonut:
closed by #1312

PR #1312, opened 2026-09-03
Reuse schema in :validate errors if exists

NoahTheDuke:
Here's a small PR to fix the potential bug I found in #1311.


opqdonut (approved):
Excellent stuff, thanks!

PR #1313, opened 2026-09-12
Fix humanize with -resolve-root-error on :or of closed :maps

### LINKED ISSUE #1308: malli.error/humanize with malli.error/-resolve-root-error throws with `:or` schema of `:closed` `:map`s
```clj
(require '[malli.core :as m]
         '[malli.error :as me])

(def schema
  [:or
   [:map {:closed true} [:a :string]]
   [:map {:closed true} [:b :string]]])

(me/humanize (m/explain schema {:a 123})
             {:resolve me/-resolve-root-error})
;!! clojure.lang.ExceptionInfo: :malli.core/invalid-schema {:type :malli.core/invalid-schema, :message :malli.core/invalid-schema, :data {:schema nil, :form nil}}
```

The cause might be to be that `malli.core/explain` points at the `:a` key in the 2nd `:or` branch, which does not have an `:a` key, but `-resolve-root-error` tries to find it anyway, yielding `nil`?
```clj
(:errors (m/explain schema {:a 123}))
;=>
;({:path [0 :a], :in [:a], :schema :string, :value 123}
; {:path [1 :b], :in [:b], :schema [:map {:closed true} [:b :string]], :value nil, :type :malli.core/missing-key}
; {:path [1 :a], :in [:a], :schema [:map {:closed true} [:b :string]], :value 123, :type :malli.core/extra-key})
```

arimu1:
## Summary

- Guard `-resolve-root-error` so it skips `get-in` paths that do not exist on the root schema (nil lens), instead of calling `error-message` on a nil schema.
- Adds a regression test for the `:or` of closed `:map`s case from #1308.

Fixes #1308

## Test plan

- [x] `./bin/kaocha --focus malli.error-test/error-definion-lookup-test`


test/malli/error_test.cljc
```diff
@@ -643,6 +643,16 @@
             (m/explain [:ref #'VarSchema] {:foo "2"})
             {:resolve me/-resolve-root-error})))))
 
+  (testing "refs #1308"
```
opqdonut, line 646: perhaps something like ":or #1308" would be better. this is not about `:ref` schemas (unlike the previous block)

test/malli/error_test.cljc
```diff
@@ -643,6 +643,16 @@
...
+    (let [schema [:or
+                  [:map {:closed true} [:a :string]]
+                  [:map {:closed true} [:b :string]]]]
```
opqdonut, line 649: since this is testing `me/-resolve-root-error`, I think the test case should have an `:error/message` on the entry level or the map level to make sure the error resolving actually happens

opqdonut (changes requested):
Thanks, good stuff. I'd just like some changes to the test!

arimu1:
Addressed the review in 285babcf:

1. Renamed the block to `:or #1308` (distinct from the `:ref` case above).
2. Added entry-level `:error/message` on `:a` so `-resolve-root-error` actually resolves to `"entry-failure"` for that branch while still covering the closed-map / `:or` path that previously threw.

`clojure -M:test -m kaocha.runner --focus 'malli.error-test/error-definion-lookup-test'` — 1 test, 13 assertions, 0 failures.

opqdonut (approved):
thanks a lot!

PR #1315, opened 2026-09-21
improve clj-kondo generated types

### LINKED ISSUE #1314: Generated clj-kondo types could be improved
Several things could be improved in the generated `clj-kondo` configuration:

1. clj-kondo can handle sets of types, so a schema like `[:or :string :int]` can be fully specified using clj-kondo types (currently it will be transformed to `:any`). Less importantly, heterogenenous enums (`[:enum "a" 1]`) also get converted to `:any` and could be improved.
2. It's quite common to use a schema `[:and generic-type specific-invariant]`, currently those will be transformed to `:any`, but we could use the first type.

RolT:
This PR improves handling for several clj-kondo types, notably `:and`, `:or` and `enum` schemas. I've added some unit tests, but the best way to verify that things work it to actually use clj-kondo and check warnings on a clojure file.

Let me know if you'd prefer split PRs, split commits, etc.

Fixes #1314


opqdonut:
I'd like to start by merging the simple and obviously-correct parts of this: `:or`, `:enum`, `:=`. Then we can discuss what to do with `:and` and `:multi`. See #1182 for related work on `:and`.

RolT:
Hello, I updated my branch to remove the changes to `:and`, `:andn` & `:multi`. Should I split the issue too ?

I didn't understand your point about the linked branch. For clj-kondo types, the best we can do for `:and` type is to pick the most specific one. There's a type hierarchy (type lattice ?) defined in clj-kondo that could be reused, and we should probably use the leftmost one when 2 types don't have a hierarchical relation. But maybe that's just a bit much to implement and maintain, in that case we could simply take the first child that is not `:any`.

For `:multi` I decided to ignore the dispatch fn and just treat it like a union of all cases.

opqdonut:
Thanks, this is good.

Sorry, I didn't explain myself properly regarding #1182 . It introduces the concepts of _simple_ and _transforming_ parsers, in order to be able to handle schemas like `[:and [:map ...] [:fn ...]]` properly, without relying on the ordering of the children. I was wondering if the same sort of logic could help pick out the right schema to pass on to clj-kondo?

I agree re: treating multi like an `:and`.

You don't need to split the issue, you can open another PR against the same issue.

opqdonut (approved):

